PR #13, opened 2025-03-20
Jh/checkmacros

Jutho:


src/algorithms.jl
```diff
@@ -143,3 +143,26 @@ macro functiondef(f)
                    Core.@__doc__ $f, $f!
                end)
 end
+
```
lkdvos, line 146:
```suggestion

"""
    @check_scalar(x, y, [op], [eltype])

Check if `eltype(x) == op(eltype(y))` and throw an error if not.
By default `op = identity` and `eltype = eltype'.
"""
```

src/algorithms.jl
```diff
@@ -143,3 +143,26 @@ macro functiondef(f)
...
+    if op == :identity
+        error_message *= " != " * string(eltype) * "(" * string(y) * ")"
+    else
+        error_message *= " != " * string(op) * "(" * string(eltype) * "(" * string(y) * "))"
+    end
+    return esc(quote
+                   $eltype($x) == $op($eltype($y)) || throw(ArgumentError($error_message))
+               end)
+end
+macro check_size(x, sz, size=:size)
```
lkdvos, line 159:
```suggestion

"""
    @check_size(x, sz, [size])

Check if `size(x) == sz` and throw an error if not.
By default, `size = size`.
"""
macro check_size(x, sz, size=:size)
```

lkdvos (commented):
Looks good to me.

Do you know if there would be a difference if we try to achieve the same thing with functions instead of macros? I have no strong opinions against this, it's just slightly more hidden what we are actually testing for now.

Jutho:
The macro is macro only in order to capture the symbol or expression of the variable, and have it printed in the error message.

lkdvos (approved):

PR #15, opened 2025-04-09
Truncation composition

mtfishman:
This adds support for composing truncation strategies, which is then used to enable mixing truncating with `rtol`, `atol`, and `maxrank`.


lkdvos (commented):
Looks good to me!

Something I was wondering: do we want to extend this to an arbitrary number of truncation algorithms? Ie, something like

```julia
struct TruncationComposition{T<:Tuple{Vararg{TruncationStrategy}}}
    components::T
end
TruncationComposition(component::TruncationStrategy, components::TruncationStrategy) = TruncationComposition((component, components...))
```

It's not strictly necessary, since you could do `(trunc1 & trunc2) & trunc3` as well, and I don't think the performance would matter all that much, just a design question I guess.

As a sidenote, it seems like the JuliaFormatter v2 changes are still not entirely without change. I'm fine with switching to the new version, but let's just make that a separate commit and rebase this on that?

mtfishman:
Ok clearly I don't know how to use git, I don't know why the diff is showing all of the changes from previous PRs. But, I updated `TruncationComposition` based on your suggestion.

mtfishman:
Maybe I'll just start a new PR from scratch...

mtfishman:
Superseded by #18.

PR #18, opened 2025-04-09
Truncation composition

mtfishman:
Same as #15 with a cleaner git history.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MatrixAlgebraKit"
 uuid = "6c742aac-3347-4629-af66-fc926824e5e4"
 authors = ["Jutho <jutho.haegeman@ugent.be> and contributors"]
-version = "0.1.1"
+version = "0.1.2"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/test/svd.jl
+++ b/test/svd.jl
@@ -133,17 +133,15 @@ end
         Vᴴ = qr_compact(randn(rng, T, m, m))[1]
         A = U * S * Vᴴ
 
-        for (rtol, maxrank) in ((0.2, 1), (0.2, 3))
-            for trunc in ((; rtol, maxrank),
-                          truncrank(maxrank) & TruncationKeepAbove(0, rtol))
-                U1, S1, V1ᴴ = svd_trunc(A; alg, trunc=(; rtol=0.2, maxrank=1))
-                @test length(S1.diag) == 1
-                @test S1.diag ≈ S.diag[1:1] rtol = sqrt(eps(real(T)))
+        for trunc_fun in ((rtol, maxrank) -> (; rtol, maxrank),
+                          (rtol, maxrank) -> truncrank(maxrank) & TruncationKeepAbove(0, rtol))
+            U1, S1, V1ᴴ = svd_trunc(A; alg, trunc=trunc_fun(0.2, 1))
+            @test length(S1.diag) == 1
+            @test S1.diag ≈ S.diag[1:1] rtol = sqrt(eps(real(T)))
 
-                U2, S2, V2ᴴ = svd_trunc(A; alg, trunc=(; rtol=0.2, maxrank=3))
-                @test length(S2.diag) == 2
-                @test S2.diag ≈ S.diag[1:2] rtol = sqrt(eps(real(T)))
-            end
+            U2, S2, V2ᴴ = svd_trunc(A; alg, trunc=trunc_fun(0.2, 3))
+            @test length(S2.diag) == 2
+            @test S2.diag ≈ S.diag[1:2] rtol = sqrt(eps(real(T)))
         end
     end
 end
--- a/test/svd.jl
+++ b/test/svd.jl
@@ -115,3 +115,30 @@ end
         end
     end
 end
+
+@testset "svd_trunc! mix maxrank and tol for T = $T" for T in
+                                                         (Float32, Float64, ComplexF32,
+                                                          ComplexF64)
+    rng = StableRNG(123)
+    if LinearAlgebra.LAPACK.version() < v"3.12.0"
+        algs = (LAPACK_DivideAndConquer(), LAPACK_QRIteration(), LAPACK_Bisection())
+    else
+        algs = (LAPACK_DivideAndConquer(), LAPACK_QRIteration(), LAPACK_Bisection(),
+                LAPACK_Jacobi())
+    end
+    m = 4
+    @testset "algorithm $alg" for alg in algs
+        U = qr_compact(randn(rng, T, m, m))[1]
+        S = Diagonal([0.9, 0.3, 0.1, 0.01])
+        Vᴴ = qr_compact(randn(rng, T, m, m))[1]
+        A = U * S * Vᴴ
+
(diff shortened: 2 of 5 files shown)


test/svd.jl
```diff
@@ -115,3 +115,30 @@ end
...
+                LAPACK_Jacobi())
+    end
+    m = 4
+    @testset "algorithm $alg" for alg in algs
+        U = qr_compact(randn(rng, T, m, m))[1]
+        S = Diagonal([0.9, 0.3, 0.1, 0.01])
+        Vᴴ = qr_compact(randn(rng, T, m, m))[1]
+        A = U * S * Vᴴ
+
+        U1, S1, V1ᴴ = svd_trunc(A; alg, trunc=(; rtol=0.2, maxrank=1))
```
lkdvos, line 136: Could you also include a test for the `&` syntax here or below?

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "MatrixAlgebraKit"
 uuid = "6c742aac-3347-4629-af66-fc926824e5e4"
 authors = ["Jutho <jutho.haegeman@ugent.be> and contributors"]
-version = "0.1.1"
+version = "0.1.2"
```
lkdvos, line 4: I think this is correct, but just keep in mind that we typically only bump the version when we actually register, so the assumption that this PR doesn't introduce breaking changes therefore this update is not breaking doesn't always hold.

In principle we've added the pullback support since v0.1.1, which is a rather substantial amount of code, which might warrant increasing the version.
Definitely curious to hear your opinion on this, the v0.x semver rules are always a bit tricky to me.

mtfishman: Got it, thanks for the heads up. Would you prefer that I don't bump the version for now? This kind of thing is why I like to bias towards making releases after every PR, so there isn't a question about when to make a release and features become available right away.

I usually just follow the rule that I bump `0.x.y` to `0.x.y+1` if it isn't breaking (so that includes adding new features), and then `0.x.y` to `0.x+1.0` if it is breaking. If a package doesn't have many dependents I don't think it really matters but once there are a decent number of dependents, minimizing the number of breaking releases is obviously friendlier.

lkdvos: I'm definitely okay with bumping it here, just wanted to give you the heads up because I know the workflow is a bit different :)

mtfishman: Yes, definitely good to get on the same page about that!

test/svd.jl
mtfishman: I've added tests explicitly using and directly testing some of the truncation objects, I realize some of that may be internal and may get changed so let me know what you think.

test/svd.jl
```diff
@@ -115,3 +115,35 @@ end
...
+        A = U * S * Vᴴ
+
+        for (rtol, maxrank) in ((0.2, 1), (0.2, 3))
+            for trunc in ((; rtol, maxrank),
+                          truncrank(maxrank) & TruncationKeepAbove(0, rtol))
+                U1, S1, V1ᴴ = svd_trunc(A; alg, trunc=(; rtol=0.2, maxrank=1))
+                @test length(S1.diag) == 1
+                @test S1.diag ≈ S.diag[1:1] rtol = sqrt(eps(real(T)))
+
+                U2, S2, V2ᴴ = svd_trunc(A; alg, trunc=(; rtol=0.2, maxrank=3))
```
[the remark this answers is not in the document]
mtfishman: Whoops, good catch. Also that logic was totally off. I should have written some tests for my tests.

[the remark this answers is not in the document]
mtfishman: Should be fixed now.

lkdvos (approved):

Jutho (approved):

Jutho:
Very clean PR and nice tests. Only comment/question is that `TruncationComposition` is a generic name that does not necessarily communicate the `&` (`intersect`) nature of the composition. Would there ever be a need for an `|` (`union`) form of truncation?

mtfishman:
> Very clean PR and nice tests. Only comment/question is that `TruncationComposition` is a generic name that does not necessarily communicate the `&` (`intersect`) nature of the composition. Would there ever be a need for an `|` (`union`) form of truncation?

Good question. I actually chose that name because at first I had `TruncationComposition` store a function that would be used to combine the spectra, where `instersect` was just a special case, but then I couldn't come up with a better name for the special case of intersection.

A use case I could think of for `union` would be filtering both large and small singular values/eigenvalues (it sounds a bit contrived but it comes up when diagonalizing correlation matrices if you want to filter the most occupied/unoccupied orbitals).

Any suggestions for other names? Maybe `TruncationIntersection`? Otherwise I could change the design to:
```julia
struct TruncationComposition{F,T<:Tuple{Vararg{TruncationStrategy}}} <:
       TruncationStrategy
    f::F
    components::T
end
```
with a definition:
```julia
function Base.:&(trunc1::TruncationStrategy, trunc2::TruncationStrategy)
    return TruncationComposition(intersect, (trunc1, trunc2))
end

function findtruncated(values::AbstractVector, strategy::TruncationComposition)
    inds = map(Base.Fix1(findtruncated, values), strategy.components)
    return strategy.f(inds...)
end
```

lkdvos:
TensorKit had `MultipleTruncation`, but I'm not sure if that's any better. I think I like `TruncationIntersection`, seems to indicate well what it is doing, and from the user side you probably don't have to interact too much with that name either, because `&` is more easily accessible?

The proposed general `TruncationComposition` also looks fine to me, it might be "overengineered" in the sense that I can only come up with `union` or `intersect`, but probably that's also just my lack of imagination, so I don't see why not

mtfishman:
> TensorKit had `MultipleTruncation`, but I'm not sure if that's any better. I think I like `TruncationIntersection`, seems to indicate well what it is doing, and from the user side you probably don't have to interact too much with that name either, because `&` is more easily accessible?

Yes, I would picture these types would be internal (maybe we would use them internally in ITensor to implement some fancier truncation schemes, though even in those cases I imagine `&` will suffice).

> The proposed general `TruncationComposition` also looks fine to me, it might be "overengineered" in the sense that I can only come up with `union` or `intersect`, but probably that's also just my lack of imagination, so I don't see why not

I guess I could imagine other set operations (`setdiff`, `symdiff`) could be useful.

For now I'll go with `TruncationIntersection`, and keep in mind the more general `TruncationComposition` if we start needing other compositions and don't want to define types for each one. `TruncationIntersection` could later be defined as `TruncationComposition{typeof(intersect)}`.

Jutho:
Looks good to me. I also liked the `TruncationComposition` proposal with the `f` field. You then only loose that you cannot join more than 2 components as you might have intersections and unions. So in that case I would just have 2 component fields explicitly, instead of a tuple field, and also nest `TruncationComposition` objects (as in reality there will not be that many of them I believe).

But the current `TruncationIntersection` is good. Shall I merge?

mtfishman:
> Looks good to me. I also liked the `TruncationComposition` proposal with the `f` field. You then only loose that you cannot join more than 2 components as you might have intersections and unions. So in that case I would just have 2 component fields explicitly, instead of a tuple field, and also nest `TruncationComposition` objects (as in reality there will not be that many of them I believe).

Yes, that would be something we'd need to work out. I was thinking we could still have a tuple field, but if the `f`s are equal we could splat them together, and otherwise it would become nested.

> But the current `TruncationIntersection` is good. Shall I merge?

Sure, I'd be happy to implement the more general `TruncationComposition` in a future PR when use cases arise.

PR #19, opened 2025-04-09
More general truncation and algorithm selection in orth/null

mtfishman:
This is a work in progress for allowing more general truncation and algorithm selection in orth/null. So far I've only worked on `left_orth!` so I can get feedback on the design. Also some of the tests will depend on #18 getting merged in order to work with the new design.

@lkdvos and @Jutho, I'm curious to hear your thoughts, particularly related to the design for selecting the algorithms of different factorizations.

The basic idea is that you can select an algorithm or a set of keyword arguments for the default algorithm of each factorization separately, for example:
```julia
using MatrixAlgebraKit

A = randn(4, 4)

# QR will be selected, in which case the `blocksize` will be used and `alg_svd` will be ignored.
V, C = left_orth(A; alg_qr=(; blocksize=2), alg_svd=LAPACK_DivideAndConquer())

# Truncated SVD will be selected, in which case the truncation and specified SVD algorithm will be used and the `blocksize` will be ignore.
V, C = left_orth(A; trunc=(; maxrank=2), alg_qr=(; blocksize=2), alg_svd=LAPACK_DivideAndConquer())
```


Jutho:
I definitely like this approach. The only thing I wonder about is if the `alg_...` variables become type unstable using this approach, but that is probably a negligible effect anyway.

mtfishman:
> I definitely like this approach. The only thing I wonder about is if the `alg_...` variables become type unstable using this approach, but that is probably a negligible effect anyway.

I refactored the logic a bit, hopefully it is now friendlier for humans and compilers.

mtfishman:
I considered having separate keyword arguments `alg_qr` and `kwargs_qr` but that seemed excessive.

Jutho:
I agree that would be excessive and like the current approach. The test breaking seems to be because this is based on an older version where `rtol` truncation was not yet supported.

mtfishman:
> I agree that would be excessive and like the current approach. The test breaking seems to be because this is based on an older version where `rtol` truncation was not yet supported.

Yes, I just merged the changes to #18 so I'll be able to fix that now. I'll go through and update tests, add new tests, update the other orth and null functions, etc.

mtfishman:
Ok, this PR is ready for a review, I've updated `left_orth`, `right_orth`, `left_null`, and `right_null`, along with the docs and tests.

The most awkward part of this new design is deciding how to interpret `trunc` in `left_null` and `right_null`.

The latest design is that if you pass `trunc=(; atol, rtol, maxnullity)`, these decide which small singular vectors to keep by building a corresponding truncation strategy. `maxnullity` specifies the maximum size of the null space to output, so is like the complement of `maxrank`.

However, if you pass a `TruncationStrategy` to `trunc`, it is taken literally, so it basically acts the same way `orth` would act for that `TruncationStrategy` (i.e. it would keep the same singular vectors as if you passed that `trunc` to `svd_full`). I see now that's why the previous design may have been more restrictive about the inputs for the truncation.

The choices I see are:
1. The current design is fine, and it is up to users to pass a good truncation strategy that selects some part of the null space.
2. The truncation strategy could be run on the singular values and then the complement of the ones that are filtered would be used. That seems a bit too tricky/magical.
3. We only allow a `NamedTuple` to be passed as `trunc`, but I'm not a huge fan of that since there are reasonable alternative truncation strategies such as picking the null space based on summing the square of singular values and that would be difficult to implement without allowing a generally truncation strategy input.

Curious to hear your thoughts on that, I'm definitely open to suggestions.

src/interface/orthnull.jl
```diff
@@ -19,43 +19,47 @@ end
...
-that should be used to factor `A`, whereas `atol` and `rtol` can be used to control the
+that should be used to factor `A`, whereas `trunc` can be used to control the
 precision in determining the rank of `A` via its singular values.
 
+`trunc` can either be a truncation strategy object or a NamedTuple with fields
+`atol`, `rtol`, and `maxrank`.
+
 This is a high-level wrapper and will use one of the decompositions
-`qr!`, `svd!`, and `left_polar!` to compute the orthogonal basis `V`, as controlled
+`qr_compact!`, `svd_compact!`/`svd_trunc!`, and `left_polar!` to compute the orthogonal basis `V`, as controlled
```
lkdvos, line 35:
```suggestion
[`qr_compact!`](@ref), [`svd_compact!`](@ref)/[`svd_trunc!`](@ref), and [`left_polar!`](@ref) to compute the orthogonal basis `V`, as controlled
```
(plus maybe reformatting)

lkdvos (commented):
Overall definitely a nice set of improvements, I like the interface with the `alg_qr`, `alg_svd`, ... different options.

I left some minor comments, and we may consider using the `alg=algstruct` keywords or not.

Otherwise, given that this is definitely a breaking change, do we want to plan what we want to include in v0.2? It seems like at least the `copy_input` with `alg` argument might also be bundled in that release, since that would be breaking. I think the string macros wouldn't be breaking so that could go anywhere inbetween.
I would definitely say that we can leave the CUDA stuff to v0.3

src/implementations/truncation.jl
```diff
@@ -69,11 +69,11 @@ TruncationKeepBelow(atol::Real, rtol::Real) = TruncationKeepBelow(promote(atol,
 
 # TODO: better names for these functions of the above types
 """
-    truncrank(howmany::Int, by=abs, rev=true)
+    truncrank(howmany::Int; by=abs, rev=true)
 
 Truncation strategy to keep the first `howmany` values when sorted according to `by` or the last `howmany` if `rev` is true.
 """
-truncrank(howmany::Int, by=abs, rev=true) = TruncationKeepSorted(howmany, by, rev)
+truncrank(howmany::Int; by=abs, rev=true) = TruncationKeepSorted(howmany, by, rev)
```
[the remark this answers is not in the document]
mtfishman: Good point, I forgot I made this change. I can put that in a separate PR.

mtfishman:
> Otherwise, given that this is definitely a breaking change, do we want to plan what we want to include in v0.2? It seems like at least the `copy_input` with `alg` argument might also be bundled in that release, since that would be breaking. I think the string macros wouldn't be breaking so that could go anywhere inbetween. I would definitely say that we can leave the CUDA stuff to v0.3

That plan sounds good to me. Regarding the string macro, it would be breaking if we decide to delete the aliases and just use string macros to refer to algorithm types. I don't have a very strong opinion on that, just bringing it up as a point of discussion.

> It also looks like the branch is out of date with the main, so maybe you could merge/rebase?

I merged, I wasn't able to rebase.

src/implementations/truncation.jl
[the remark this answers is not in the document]
mtfishman: Implementing in #22.

src/implementations/orthnull.jl
```diff
@@ -159,59 +156,70 @@ end
...
-        alg = get(kwargs, :alg, select_algorithm(svd_full!, A))
-        U, _, _ = svd_full!(A, alg)
+        alg_qr′ = algorithm_or_select_algorithm(qr_null!, A, alg_qr)
+        return qr_null!(A, N, alg_qr′)
+    elseif kind == :svd && isnothing(trunc)
+        alg_svd′ = algorithm_or_select_algorithm(svd_full!, A, alg_svd)
+        U, _, _ = svd_full!(A, alg_svd′)
+        (m, n) = size(A)
+        return copy!(N, view(U, 1:m, (n + 1):m))
+    elseif kind == :svd && isnothing(trunc)
```
Jutho, line 184: Do I see correctly that this case is duplicated?
Lines 178-183 = 184 - 188

mtfishman: It looks like it, not sure how that happened... I'll fix it.

src/implementations/orthnull.jl
```diff
@@ -79,78 +79,75 @@ function initialize_output(::typeof(right_null!), A::AbstractMatrix)
     return Nᴴ
 end
 
+function algorithm_or_select_algorithm(f, A::AbstractMatrix, alg::AbstractAlgorithm)
+    return alg
+end
+function algorithm_or_select_algorithm(f, A::AbstractMatrix, kwargs::NamedTuple)
+    return select_algorithm(f, A; kwargs...)
+end
```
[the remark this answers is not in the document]
mtfishman: Thinking about it more, I think it is nice to be able to just write `alg_qr=LAPACK...` directly, so I'd prefer to keep it as is and of course users can use all of the options available in `select_algorithm` by writing `alg_qr=(; alg=...)` if they want to.

src/interface/orthnull.jl
```diff
@@ -19,43 +19,47 @@ end
...
+    This requires `isnothing(trunc)` and `left_orth!(A, [VC])` is equivalent to
     `left_polar!(A, [VC], alg)` with a default value `alg = select_algorithm(left_polar!, A)`
 
-*   `kind == :svd`: `V` and `C` are computed using the singular value decomposition `svd_trunc!`,
-    where `V` will contain the left singular vectors corresponding to the singular values that
-    are larger than `max(atol, rtol * σ₁)`, where `σ₁` is the largest singular value of `A`.
-    `C` is computed as the product of the singular values and the right singular vectors,
-    i.e. with `U, S, Vᴴ = svd_trunc!(A)`, we have `V = U` and `C = S * Vᴴ`.
+*   `kind == :svd`: `V` and `C` are computed using the singular value decomposition `svd_compact!`
+    if no truncation is specified through the `trunc` keyword argument or `svd_trunc!`
```
Jutho, line 49: How about:
```
`V` and `C` are computed using the singular value decomposition `svd_trunc!` when a truncation strategy is specified using the `trunc` keyword argument, and using `svd_compact!` otherwise.

File: src/interface/orthnull.jl [hunk 6]
```diff
@@ -74,43 +78,47 @@ function left_orth(A::AbstractMatrix; kwargs...)
...
+    `right_polar!(A, [CVᴴ], alg)` with a default value `alg = select_algorithm(right_polar!, A)`
 
-*   `kind == :svd`: `C` and `Vᴴ` are computed using the singular value decomposition `svd_trunc!`,
-    where `V = adjoint(Vᴴ)` will contain the right singular vectors corresponding to the singular
-    values that are larger than `max(atol, rtol * σ₁)`, where `σ₁` is the largest singular value of `A`.
-    `C` is computed as the product of the singular values and the right singular vectors,
-    i.e. with `U, S, Vᴴ = svd_trunc!(A)`, we have `C = rmul!(U, S)` and `Vᴴ = Vᴴ`.
+*   `kind == :svd`: `C` and `Vᴴ` are computed using the singular value decomposition `svd_compact!`
+    if no truncation is specified through the `trunc` keyword argument or `svd_trunc!`
+    if truncation is specified through the `trunc` keyword argument.
```
Jutho, line 109:
```markdown
`C` and `Vᴴ` are computed using the singular value decomposition `svd_trunc!` when a truncation strategy is specified using the `trunc` keyword argument, and using `svd_compact!` otherwise.
```

src/interface/orthnull.jl
```diff
@@ -74,43 +78,47 @@ function left_orth(A::AbstractMatrix; kwargs...)
...
-When `kind` is not provided, the default value is `:lqpos` when `iszero(atol) && iszero(rtol)`
+When `kind` is not provided, the default value is `:lq` when `isnothing(trunc)`
 and `:svd` otherwise. Finally, finer control is obtained by providing an explicit algorithm
-using the `alg` keyword argument, which should be compatible with the chosen or default value
-of `kind`.
+for backend factorizations through the `alg_lq`, `alg_polar`, and `alg_svd` keyword arguments,
+which will only be used if the corresponding factorization is called based on the other inputs.
+If `alg_lq`, `alg_polar`, or `alg_svd` are NamedTuples, a default algorithm is chosen
+with `select_algorithm` and the NamedTuple is passed as keyword arguments to that algorithm.
+`alg_lq` defaults to `(; positive=true)` so that by default a positive QR decomposition will
```
Jutho, line 120:
```suggestion
`alg_lq` defaults to `(; positive=true)` so that by default a positive LQ decomposition will
```

src/interface/orthnull.jl
```diff
@@ -179,33 +189,32 @@ function left_null(A::AbstractMatrix; kwargs...)
...
 
 *   `kind == :svd`: `N` is computed using the singular value decomposition and will contain 
     the left singular vectors corresponding to the singular values that
     are smaller than `max(atol, rtol * σ₁)`, where `σ₁` is the largest singular value of `A`.
 
-When `kind` is not provided, the default value is `:lqpos` when `iszero(atol) && iszero(rtol)`
+When `kind` is not provided, the default value is `:lq` when `isnothing(trunc)`
 and `:svd` otherwise. Finally, finer control is obtained by providing an explicit algorithm
 using the `alg` keyword argument, which should be compatible with the chosen or default value
 of `kind`.
```
Jutho, line 220:
```suggestion
using the `alg_lq` and `alg_svd` keyword arguments, which will only be used by the corresponding
factorization backend. If `alg_lq` or `alg_svd` are NamedTuples, a default algorithm is chosen
with `select_algorithm` and the NamedTuple is passed as keyword arguments to that algorithm.
`alg_lq` defaults to `(; positive=true)` so that by default a positive LQ decomposition will
be used.
```

mtfishman: Whoops, thanks for catching that.

mtfishman:
Thanks for all of the comments and corrections, I think I addressed all of them.

src/implementations/orthnull.jl
```diff
@@ -79,78 +79,75 @@ function initialize_output(::typeof(right_null!), A::AbstractMatrix)
...
+
```
lkdvos, line 88: Perhaps this definition makes slightly more sense to migrate to the `src/algorithms.jl` file?

lkdvos (commented):
Found a typo and small suggestion for where to put a function, but otherwise this looks good to me!

src/implementations/orthnull.jl
mtfishman: Yes that's a good point.

mtfishman: I moved it and renamed `algorithm_or_select_algorithm` to `_select_algorithm`, which matches the name I used for similar functionality in #23, which I'm happy to change (though maybe the name and design can be discussed in #23 since the name is less important in this PR).

lkdvos (approved):

PR #20, opened 2025-04-17
WIP CUDA support

Jutho:
Very much work in progress.

A first draft for CUDA support, for now only for QR and untested. My local machine with GPU has currently some issues with installing CUDARuntime.jl, which is a major hinderance for further developing and testing this. I hope this gets resolved soon. 

Also, most of the code in yacusolver.jl is either Copilot generated or merely copy pasted and should definitely be ignored at this point.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,9 +8,11 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 
 [weakdeps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
 
 [extensions]
 MatrixAlgebraKitChainRulesCoreExt = "ChainRulesCore"
+MatrixAlgebraKitCUDAExt = "CUDA"
 
 [compat]
 Aqua = "0.6, 0.7, 0.8"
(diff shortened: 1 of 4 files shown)


Jutho:
Ok I got QR and LQ working. 

I find the algorithm argument to `initialize_output` an annoyance often, but I now adopted the strategy to define the generic `initialize_output` methods with an `::AbstractAlgorithm` final argument which catches all cases. If a specific algorithm does want a modified `initialize_output` method, it can still do so by defining a more specific method.

Has anyone experience with setting up GPU/CUDA test infrastructure? I've created test files with are working locally, but it would of course be good to also have CI for this.

kshyatt:
Managed to rebase this locally, will work on getting the tests fixed now

src/implementations/lq.jl
```diff
@@ -181,7 +181,7 @@ function lq_via_qr!(A::AbstractMatrix, L::AbstractMatrix, Q::AbstractMatrix,
                     qr_alg::AbstractAlgorithm)
     m, n = size(A)
     minmn = min(m, n)
-    At = adjoint!(similar(A'), A)
+    At = adjoint!(similar(A'), A)::AbstractMatrix
```
Jutho, line 184: Aha, this is where the type assert was added. Is this only JET that complains, or was there really a type instability or type inference issue? I cannot see how asserting an abstract type would help with anything.

PR #21, opened 2025-04-17
Fix docstring `TruncationIntersection`

lkdvos:
Was slightly too late to get this in: 
@mtfishman 

Small change to docstring and add missing constructor.


mtfishman:
Oops, thanks @lkdvos.

Jutho:
Well spotted. Should we tag a new version after merging this?

lkdvos:
Yes, let's tag a new version. I think the version was already increased, so we can simply tag this commit

PR #22, opened 2025-04-24
Change some `truncrank` args to kwargs

mtfishman:
Split off from #19.


mtfishman:
The motivation is to make the interface closer to `sort`, and also because I think it would be easy to forget which argument is which if they are kept as positional arguments.

lkdvos (approved):

Jutho (approved):

PR #23, opened 2025-04-24
Refactor algorithm selection logic

mtfishman:
This refactors the algorithm selection logic (`select_algorithm`) in terms of two layers. In this new design, there is a generic function `select_algorithm` for any factorization function and matrix type that processes the `alg` keyword argument, and either uses the algorithm that is specified or if none is specified forwards to a lower level generic function `default_algorithm`, which then chooses a default algorithm depending on the factorization function and matrix type.

The motivation for this was that I started making a PR to make `select_algorithm` stricter about how it handles keyword arguments, but it was cumbersome to do since the `select_algorithm` logic for processing keyword arguments was repeated for each factorization function. In addition, I added logic for allowing an algorithm type (in addition to a Symbol and algorithm object) to be passed as `alg`, which was also easier to implement in the new design.

@lkdvos and I were discussing the naming of these two layers of functions (currently `select_algorithm` and `default_algorithm`), he proposed `select_algorithm` -> `_select_algorithm` and `default_algorithm` -> `select_algorithm`. I'm definitely open to suggestions, curious what you think @Jutho.

To-do:
- [x] Update the docstrings and add docs for `default_algorithm`.
- [x] Dedicated tests for `select_algorithm`/`default_algorithm`.
- [x] Investigate type stability.
- [x] Decide on the names.

### DIFF
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -31,7 +31,7 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ,
 export truncrank, trunctol, truncabove, TruncationKeepSorted, TruncationKeepFiltered
 
 VERSION >= v"1.11.0-DEV.469" &&
-    eval(Meta.parse("public default_truncation, select_algorithm"))
+    eval(Meta.parse("public default_algorithm, select_algorithm"))
 
 include("common/defaults.jl")
 include("common/initialization.jl")
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -30,6 +30,8 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ,
        LAPACK_DivideAndConquer, LAPACK_Jacobi
 export truncrank, trunctol, truncabove, TruncationKeepSorted, TruncationKeepFiltered
 
+public default_truncation, select_algorithm
+
 include("common/defaults.jl")
 include("common/initialization.jl")
 include("common/pullbacks.jl")
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -30,7 +30,8 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ,
        LAPACK_DivideAndConquer, LAPACK_Jacobi
 export truncrank, trunctol, truncabove, TruncationKeepSorted, TruncationKeepFiltered
 
-public default_truncation, select_algorithm
+VERSION >= v"1.11.0-DEV.469" &&
+    eval(Meta.parse("public default_truncation, select_algorithm"))
 
 include("common/defaults.jl")
 include("common/initialization.jl")
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -31,7 +31,7 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ,
 export truncrank, trunctol, truncabove, TruncationKeepSorted, TruncationKeepFiltered
 
 VERSION >= v"1.11.0-DEV.469" &&
-    eval(Meta.parse("public default_algorithm, select_algorithm"))
+    eval(Expr(:public, :default_algorithm, :select_algorithm))
 
 include("common/defaults.jl")
 include("common/initialization.jl")
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -61,6 +61,27 @@ implementing the function `f` on inputs of type `A`.
 """
 function select_algorithm end
 
+function select_algorithm(f, A; alg=nothing, kwargs...)
+    return _select_algorithm(f, A, alg; kwargs...)
+end
+
+function _select_algorithm(f, A, alg::Nothing; kwargs...)
+    return default_algorithm(f, A; kwargs...)
+end
+function _select_algorithm(f, A, alg::AbstractAlgorithm; kwargs...)
+    isempty(kwargs) || throw(ArgumentError("Additional keyword arguments are not allowed when an algorithm is specified."))
+    return alg
+end
+function _select_algorithm(f, A, alg::Symbol; kwargs...)
+    return _select_algorithm(f, A, Algorithm{alg}; kwargs...)
+end
+function _select_algorithm(f, A, alg::Type; kwargs...)
+    return _select_algorithm(f, A, alg(; kwargs...))
+end
+function _select_algorithm(f, A, alg; kwargs...)
+    return throw(ArgumentError("Unknown alg $alg"))
+end
+
 @doc """
     copy_input(f, A)
 
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -64,7 +64,7 @@ In general, if an algorithm is specified explicitly through the `alg` keyword ar
 that algorithm will be used instead of selecting it automatically. However, that
 behavior may be modified for factorization functions and/or matrix types.
 
-In general, if the algorithm is not specified, a default algorithm specified by
+When the `alg` keyword argument is not provided, a default algorithm specified by
 [`default_algorithm`](@ref) will be used.
 """
 function select_algorithm end
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -56,8 +56,16 @@ end
 @doc """
     select_algorithm(f, A; kwargs...)
 
-Given some keyword arguments and an input `A`, decide on an algrithm to use for
+Given some keyword arguments and an input `A`, decide on an algorithm to use for
 implementing the function `f` on inputs of type `A`.
+
+In general, if an algorithm is specified explicitly through the `alg` keyword argument
+(either as an algorithm type, an algorithm name as a Symbol, or as an algorithm object),
+that algorithm will be used instead of selecting it automatically. However, that
+behavior may be modified for factorization functions and/or matrix types.
+
+In general, if the algorithm is not specified, a default algorithm specified by
+[`default_algorithm`](@ref) will be used.
 """
 function select_algorithm end
 
@@ -81,6 +89,15 @@ end
 function _select_algorithm(f, A, alg; kwargs...)
     return throw(ArgumentError("Unknown alg $alg"))
 end
+
+@doc """
+    default_algorithm(f, A; kwargs...)
+
+Select the default algorithm for a given factorization function `f` and input `A`.
+In general, this is called by [`select_algorithm`](@ref) if no algorithm is specified
+explicitly.
+"""
+function default_algorithm end
 
 @doc """
     copy_input(f, A)
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -59,7 +59,7 @@ end
 Decide on an algorithm to use for implementing the function `f` on inputs of type `A`.
 
 If `alg` is `nothing` (the default value), an algorithm will be selected automatically
-with [`MatrixAlgebra.default_algorithm`](@ref) and the keyword arguments will be passed
+with [`MatrixAlgebraKit.default_algorithm`](@ref) and the keyword arguments will be passed
 to the algorithm constructor.
 
 If `alg` is a `NamedTuple`, an algorithm will be selected automatically
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -72,17 +72,20 @@ can't be passed to `MatrixAlgebraKit.select_algorithm`.
 """
 function select_algorithm end
 
-Base.@constprop :aggressive function select_algorithm(f::F, A, alg::Alg=nothing; kwargs...) where {F,Alg}
+Base.@constprop :aggressive function select_algorithm(f::F, A, alg::Alg=nothing;
+                                                      kwargs...) where {F,Alg}
     return _select_algorithm(f, A, alg; kwargs...)
 end
 
 function _select_algorithm(f::F, A, alg::Nothing; kwargs...) where {F}
     return default_algorithm(f, A; kwargs...)
 end
-Base.@constprop :aggressive function _select_algorithm(f::F, A, alg::Symbol; kwargs...) where {F}
+Base.@constprop :aggressive function _select_algorithm(f::F, A, alg::Symbol;
+                                                       kwargs...) where {F}
     return Algorithm{alg}(; kwargs...)
 end
-Base.@constprop :aggressive function _select_algorithm(f::F, A, ::Type{Alg}; kwargs...) where {F,Alg}
+Base.@constprop :aggressive function _select_algorithm(f::F, A, ::Type{Alg};
+                                                       kwargs...) where {F,Alg}
     return Alg(; kwargs...)
 end
 function _select_algorithm(f::F, A, alg::NamedTuple; kwargs...) where {F}
@@ -174,7 +177,8 @@ macro functiondef(f)
 
     return esc(quote
                    # out of place to inplace
-                   Base.@constprop :aggressive $f(A; kwargs...) = $f!(copy_input($f, A); kwargs...)
+                   Base.@constprop :aggressive $f(A; kwargs...) = $f!(copy_input($f, A);
+                                                                      kwargs...)
                    $f(A, alg::AbstractAlgorithm) = $f!(copy_input($f, A), alg)
 
                    # fill in arguments
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -77,7 +77,8 @@ function _select_algorithm(f, A, alg::Nothing; kwargs...)
     return default_algorithm(f, A; kwargs...)
 end
 function _select_algorithm(f, A, alg::AbstractAlgorithm; kwargs...)
-    isempty(kwargs) || throw(ArgumentError("Additional keyword arguments are not allowed when an algorithm is specified."))
+    isempty(kwargs) ||
+        throw(ArgumentError("Additional keyword arguments are not allowed when an algorithm is specified."))
     return alg
 end
 function _select_algorithm(f, A, alg::Symbol; kwargs...)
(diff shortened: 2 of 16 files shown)


mtfishman:
What this PR calls `_select_algorithm` was called `algorithm_or_select_algorithm` in #19, that is at least a more descriptive name for what it is doing.

src/algorithms.jl
```diff
@@ -56,17 +56,54 @@ end
...
+    return default_algorithm(f, A; kwargs...)
+end
+function _select_algorithm(f, A, alg::AbstractAlgorithm; kwargs...)
+    isempty(kwargs) ||
+        throw(ArgumentError("Additional keyword arguments are not allowed when an algorithm is specified."))
     return alg
 end
-function _select_algorithm(f, A::AbstractMatrix, alg::NamedTuple)
+function _select_algorithm(f, A, alg::Symbol; kwargs...)
+    return _select_algorithm(f, A, Algorithm{alg}; kwargs...)
```
Jutho, line 85: This needs to be the singleton instance
```julia
    return _select_algorithm(f, A, Algorithm{alg}(); kwargs...)
```
or maybe
```julia
    return _select_algorithm(f, A, Algorithm{alg}(; kwargs...))
```

Jutho: Never mind, this was before I've read that there now also is a `_select_algorithm` that catches types.

mtfishman: I think what you wrote is clearer anyway. Whenever I have functions like this I go back and forth about whether they should all call one inner function or if they should call each other, but calling one inner function is probably easier to reason about and read.

src/implementations/truncation.jl
```diff
@@ -32,6 +32,19 @@ Trivial truncation strategy that keeps all values, mostly for testing purposes.
 """
 struct NoTruncation <: TruncationStrategy end
 
+function to_truncationstrategy(trunc::TruncationStrategy)
```
lkdvos, line 35: It might be reasonable to use the abstract type for this, and simply call this function `TruncationStrategy`.
The main reason not to is if we are expecting to also support truncation types that are not a subtype, but I don't think that is very likely.
Alternatively, I guess this function would be the equivalent of `select_algorithm`, so perhaps `select_truncation` is fitting?

lkdvos (commented):
Looks like a reasonable change to me, should definitely offer quite a bit of flexibility.

If we decide on this, can you also look at exporting `default_algorithm`, since that would be the main point to overload, and check if the docs need updating for this new approach?

src/implementations/truncation.jl
mtfishman: `select_truncation` sounds good to me (I'd also be ok with `TruncationStrategy` but `select_truncation` feels a bit more conservative for now, for example it could remain internal for the time being).

mtfishman:
> If we decide on this, can you also look at exporting `default_algorithm`, since that would be the main point to overload, and check if the docs need updating for this new approach?

What do you think of making it public but not exporting it for now, indicating it is more of a syntax for overloading rather than using externally?

mtfishman:
Currently, if you pass an algorithm type or Symbol as the `alg` keyword argument, `select_algorithm` and therefore the factorizations are not type stable. I thought constant propagation would handle that but I guess not. I tried playing around with `@inline` but it didn't seem to help.

lkdvos:
I don't have any strong feelings about public or exported in this case. Given that our use-cases are mostly to use this as a low-level library to build on, I don't think it matters too much. As you noticed `public` doesn't work in 1.10, I think the recommended way of resolving that is using [Compat.jl](https://github.com/JuliaLang/Compat.jl) to handle this instead of doing that manually.

I'm a bit more worried about the type stability though.
Was this type-stable before?
There's also `Base.@constprop :aggressive` as an option, or explicitly `Val`-ing some arguments in the internal implementations.

mtfishman:
> I don't have any strong feelings about public or exported in this case. Given that our use-cases are mostly to use this as a low-level library to build on, I don't think it matters too much. As you noticed `public` doesn't work in 1.10, I think the recommended way of resolving that is using [Compat.jl](https://github.com/JuliaLang/Compat.jl) to handle this instead of doing that manually.

There are long debates about that: https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041, the approach I use here is the one recommended in the Julia docs: https://docs.julialang.org/en/v1.12-dev/manual/modules/#Export-lists. Some people are weary of Compat.jl because it causes a lot of invalidations and uses type piracy, so if there is a simple way to avoid it it seems better to do so, but I'm fine either way.

> I'm a bit more worried about the type stability though. Was this type-stable before? There's also `Base.@constprop :aggressive` as an option, or explicitly `Val`-ing some arguments in the internal implementations.

I doubt it was type stable before but I can double check. I also tried `Base.@constprop :aggressive` but that didn't seem to help, though I can explore that a bit more, and I can also look into if some use of `Val` can help.

mtfishman:
I think basically what we are doing is analogous to this minimal example:
```julia
julia> f(; x) = Val{x}()

julia> @code_warntype f(; x=:x)
MethodInstance for Core.kwcall(::@NamedTuple{x::Symbol}, ::typeof(f))
  from kwcall(::NamedTuple, ::typeof(f)) @ Main REPL[56]:1
Arguments
  _::Core.Const(Core.kwcall)
  @_2::@NamedTuple{x::Symbol}
  @_3::Core.Const(Main.f)
Locals
  @_4::Symbol
  x::Symbol
Body::Val
1 ─       Core.NewvarNode(:(@_4))
│   %2  = Core.isdefined(@_2, :x)::Core.Const(true)
└──       goto #3 if not %2
2 ─       (@_4 = Core.getfield(@_2, :x))
└──       goto #4
3 ─       Core.Const(:(Core.UndefKeywordError(:x)))
└──       Core.Const(:(@_4 = Core.throw(%6)))
4 ┄ %8  = @_4::Symbol
│         (x = %8)
│   %10 = Base.keys(@_2)::Core.Const((:x,))
│   %11 = (:x,)::Core.Const((:x,))
│   %12 = Base.diff_names(%10, %11)::Core.Const(())
│   %13 = Base.isempty(%12)::Core.Const(true)
└──       goto #6 if not %13
5 ─       goto #7
6 ─       Core.Const(:(Base.kwerr(@_2, @_3)))
7 ┄ %17 = Main.:(var"#f#1")::Core.Const(Main.var"#f#1")
│   %18 = x::Symbol
│   %19 = (%17)(%18, @_3)::Val
└──       return %19

julia> @code_warntype (() -> f(; x=:x))()
MethodInstance for (::var"#5#6")()
  from (::var"#5#6")() @ Main REPL[61]:1
Arguments
  #self#::Core.Const(var"#5#6"())
Body::Val{:x}
1 ─ %1 = (:x,)::Core.Const((:x,))
│   %2 = Core.apply_type(Core.NamedTuple, %1)::Core.Const(NamedTuple{(:x,)})
│   %3 = (:x,)::Core.Const((:x,))
│   %4 = (%2)(%3)::Core.Const((x = :x,))
│   %5 = Core.kwcall(%4, Main.f)::Core.Const(Val{:x}())
└──      return %5

julia> using TestExtras

julia> @constinferred f(; x=:x)
Val{:x}()

julia> @constinferred f(; x=Symbol("x"))
Test Failed at REPL[75]:1
  Expression: @constinferred f(; x = Symbol("x"))
   Evaluated: Val{:x} != Val

ERROR: There was an error during testing
```
so it seems like at least in principle it should be inferrable.

src/MatrixAlgebraKit.jl
```diff
@@ -30,6 +30,9 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ,
        LAPACK_DivideAndConquer, LAPACK_Jacobi
 export truncrank, trunctol, truncabove, TruncationKeepSorted, TruncationKeepFiltered
 
+VERSION >= v"1.11.0-DEV.469" &&
```
Jutho, line 33: Is it worth specifying the specific DEV version that introduced the `public` keyword, or can we just use `v"1.11"`.

src/algorithms.jl
```diff
@@ -54,19 +54,56 @@ function _show_alg(io::IO, alg::Algorithm)
...
-Given some keyword arguments and an input `A`, decide on an algrithm to use for
+Given some keyword arguments and an input `A`, decide on an algorithm to use for
 implementing the function `f` on inputs of type `A`.
+
+In general, if an algorithm is specified explicitly through the `alg` keyword argument
+(either as an algorithm type, an algorithm name as a Symbol, or as an algorithm object),
+that algorithm will be used instead of selecting it automatically. However, that
+behavior may be modified for factorization functions and/or matrix types.
+
+In general, if the algorithm is not specified, a default algorithm specified by
```
Jutho, line 67: I wouldn't start the second paragraph also with "In general"

Jutho, line 67:
```suggestion
When the `alg` keyword argument is not provided, a default algorithm specified by
```

src/algorithms.jl
```diff
@@ -54,19 +54,56 @@ function _show_alg(io::IO, alg::Algorithm)
...
+[`default_algorithm`](@ref) will be used.
 """
 function select_algorithm end
 
-function _select_algorithm(f, A::AbstractMatrix, alg::AbstractAlgorithm)
+function select_algorithm(f, A; alg=nothing, kwargs...)
```
Jutho, line 72: I am not sure using multiple dispatch via `_select_algorithm` to deal with the different cases of `alg` keywords is the right approach, as it might put more strain on the compiler. I don't think there is need to be able to extend this functionality, so something like
```julia
function select_algorithm(f, A; alg=nothing, kwargs...)
	if isnothing(alg)
		return default_algorithm(f, A; kwargs...)
	elseif alg isa AbstractAlgorithm
		isempty(kwargs) ||
			throw(ArgumentError("Additional keyword arguments are not allowed when an algorithm is specified."))
		return alg
	elseif
	...
	end
end
```
should work. Or does the `_select_algorithm` lowering actually help with the lowering?

test/eig.jl
```diff
@@ -8,21 +8,35 @@ using MatrixAlgebraKit: diagview
 @testset "eig_full! for T = $T" for T in (Float32, Float64, ComplexF32, ComplexF64)
     rng = StableRNG(123)
     m = 54
-    for alg in (LAPACK_Simple(), LAPACK_Expert())
+    for alg in
+        (LAPACK_Simple(), LAPACK_Expert(), LAPACK_Simple, LAPACK_Expert, :LAPACK_Simple,
+         :LAPACK_Expert)
```
Jutho, line 13: Is the idea that if we only modify this list for the `eig_full` tests, this will test the selection/default functionality and should thus also work for the other factorisations? I am fine with this, and definitely think we should not run all tests 3 times over as this would be a large amount of unnecessary compute. Just checking if this is the plan?

src/algorithms.jl
mtfishman: I'm totally fine with that, using dispatch was a bit arbitrary. I generally like using dispatch rather than if-statements purely as a style thing, but if you and the compiler prefer if-statements I can switch to that.

src/MatrixAlgebraKit.jl
mtfishman: I just copied that from the Julia docs: https://docs.julialang.org/en/v1.12-dev/manual/modules/#Export-lists.

I also know Compat.jl uses this style of being more specific about the version. I suppose if you just use `v"1.11"` someone could use the package with an earlier dev version of v1.11 before `public` was introduced, obviously that is quite a corner case and it is reasonable to tell users to not do that/that isn't supported.

mtfishman: I guess "worth" is subjective, it is a few extra characters and very slightly safer, and something that we write once and never think about again (until we drop support for v1.11 in which case we delete the entire logic anyway).

src/algorithms.jl
Jutho: If the `_select_algorithm` lowering is better for type stability, I am certainly fine with that, and there is no need to waste time trying to trick the compiler to make the `if elseif else` construction inferable as well.

test/eig.jl
Jutho: Standalone `select_algorithm` and `default_algorithm` tests are definitely a good idea. Regarding the names: I like those names 😄 !

lkdvos:
Regarding the type stability, I just remembered/realized that this change has an additional side-effect that might affect this: before we didn't have two layers (`select_algorithm` and `default_algorithm`), and as a result the implementations were written directly in terms of `select_algorithm(::typeof(f), ...)`, while now there are a bunch of definitions for `select_algorithm(f, ...)`. Given that Julia doesn't specialize on functions by default, I'm wondering if this might be affecting the outcomes. In other words, I'm wondering if it's not better to write `select_algorithm(f::F, ...) where {F}` if we want to have these two layers of functions in there.

mtfishman:
Ok, I've solved most type instability issues and added more tests. I had to rely both on forcing method specialization and `Base.@constprop :aggressive` as @lkdvos suggested.

Oddly, the one case that still isn't type stable is specifying `alg` as a type, i.e. `svd_compact!(randn(2, 2); alg=LAPACK_QRIteration)`. Any suggestions with that would be welcome...

test/eig.jl
```diff
@@ -16,16 +16,33 @@ using MatrixAlgebraKit: diagview
...
+    A = randn(rng, T, m, m)
+    Tc = complex(T)
+    D, V = @constinferred eig_full(A; alg=:LAPACK_Simple)
+    @test eltype(D) == eltype(V) == Tc
+    @test A * V ≈ V * D
+
+    A = randn(rng, T, m, m)
+    Tc = complex(T)
+    ## Inference is broken for this case for some reason.
+    ## D, V = @constinferred eig_full(A; alg=LAPACK_Simple)
```
mtfishman, line 42: Note that this is still broken.

mtfishman: This is fixed now, @lkdvos explained to me that I was using `@constinferred` wrong and should have been interpolating the `alg` input.

mtfishman:
This is ready for review. All type stability issues are fixed, except for one issue in Julia 1.10, I would vote for letting that slide (as opposed to making the code more complicated to fix it) but please let me know if you want me to look into that.

In the latest, I was able to avoid all uses of `Base.@constprop :aggressive`. @lkdvos pointed out to me that I was using `@constinferred` wrong and should have been interpolating the `alg` input in the tests.

test/svd.jl
```diff
@@ -11,17 +11,23 @@ using MatrixAlgebraKit: TruncationKeepAbove, diagview
...
-                    LAPACK_Jacobi())
+                    LAPACK_Jacobi(), LAPACK_DivideAndConquer, :LAPACK_DivideAndConquer)
         end
         @testset "algorithm $alg" for alg in algs
             n > m && alg isa LAPACK_Jacobi && continue # not supported
             minmn = min(m, n)
             A = randn(rng, T, m, n)
 
-            U, S, Vᴴ = svd_compact(A; alg)
+            if VERSION < v"1.12"
```
mtfishman, line 25:
```suggestion
            if VERSION < v"1.11"
```

test/svd.jl
```diff
@@ -11,17 +11,23 @@ using MatrixAlgebraKit: TruncationKeepAbove, diagview
...
             minmn = min(m, n)
             A = randn(rng, T, m, n)
 
-            U, S, Vᴴ = svd_compact(A; alg)
+            if VERSION < v"1.11"
+                # This is type unstable on older versions of Julia.
+                U, S, Vᴴ = svd_compact(A; alg)
+            else
+                U, S, Vᴴ = @constinferred svd_compact(A; alg=($alg))
+            end
```
mtfishman, line 30: This is the case I mentioned that isn't type stable in older versions of Julia. I'd vote for letting this slide and telling users to use a newer versions of Julia if they care about type stability, let me know what you think.

lkdvos: I'm assuming that this is also just some strange interplay with the non-heterogeneous tuple of algs, so this is more of a test artifact than a real problem. Definitely okay with letting this slide.

lkdvos (approved):
I only looked over this briefly on my phone so it would be good if @Jutho can also give his blessing, but this looks like a nice implementation now, I'm very happy the constprop annotations weren't necessary in the end.

src/MatrixAlgebraKit.jl
```diff
@@ -30,6 +30,9 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ,
...
+    eval(Meta.parse("public default_algorithm, select_algorithm"))
```
Jutho, line 34: Alternative that does not require `parse` is simply
`eval(Expr(:public, :default_algorithm, :select_algorithm))`
but I am fine with `Meta.parse` (the overhead is probably negligible).

mtfishman, line 34:
```suggestion
    eval(Expr(:public, :default_algorithm, :select_algorithm))
```

mtfishman: Makes sense, I changed it to that.

Jutho (approved):
Very nice PR, only pending comment is a suggestion in an attempt to further improve the clarity of the `select_algorithm` doc string.

src/algorithms.jl
```diff
@@ -54,20 +54,60 @@ function _show_alg(io::IO, alg::Algorithm)
...
+with [`MatrixAlgebraKit.default_algorithm`](@ref) and the keyword arguments will be passed
+to the algorithm constructor.
+
+If `alg` is a `NamedTuple`, an algorithm will be selected automatically
+with [`default_algorithm`](@ref) and `alg` will be passed to the algorithm
+as keyword arguments. In that case, keyword arguments can't be passed
+to `MatrixAlgebraKit.select_algorithm`
+
+If `alg` is an `AbstractAlgorithm`, it will be returned as-is. In that case, keyword arguments
+can't be passed to `MatrixAlgebraKit.select_algorithm`.
```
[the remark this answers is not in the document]
mtfishman: That's definitely clearer, thanks for the suggestion!

Jutho (approved):
This looks ready to be merged! Feel free to merge yourself @mtfishman , otherwise without reaction I will merge this tonight.

mtfishman:
Thanks again for the careful review.

Should we register v0.2 after #25 gets merged?

Jutho:
Yes that sounds good to me. I went through #25 and do not have any comments.

PR #24, opened 2025-05-07
Jenkins setup - GPU testing framework

lkdvos:
This is an attempt to add automated tests for the (upcoming) GPU functionality.


ci/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+    DEBIAN_FRONTEND=noninteractive \
+    apt-get install --yes --no-install-recommends \
+                    # basic stuff
+                    curl ca-certificates \
+                    libcutensor-dev \
+                    && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+ARG JULIA=1.6
```
Jutho, line 14: Does this mean Julia v1.6. Is that what we want?

lkdvos: Still working on getting this ironed out, I think this is only the default value and this is actually replaced in the Jenkinsfile. This is mostly just copied from the ITensors.jl repository and doesn't currently work yet.

kshyatt:
Is this now obviated by having the JuliaGPU setup working?

PR #25, opened 2025-05-13
Make orthnull more customizable

mtfishman:
As suggested by @lkdvos. Requires #23.

### DIFF
--- a/src/implementations/orthnull.jl
+++ b/src/implementations/orthnull.jl
@@ -1,9 +1,9 @@
 # Inputs
 # ------
-copy_input(::typeof(left_orth), A::AbstractMatrix) = copy_input(qr_compact, A) # do we ever need anything else
-copy_input(::typeof(right_orth), A::AbstractMatrix) = copy_input(lq_compact, A) # do we ever need anything else
-copy_input(::typeof(left_null), A::AbstractMatrix) = copy_input(qr_null, A) # do we ever need anything else
-copy_input(::typeof(right_null), A::AbstractMatrix) = copy_input(lq_null, A) # do we ever need anything else
+copy_input(::typeof(left_orth), A) = copy_input(qr_compact, A) # do we ever need anything else
+copy_input(::typeof(right_orth), A) = copy_input(lq_compact, A) # do we ever need anything else
+copy_input(::typeof(left_null), A) = copy_input(qr_null, A) # do we ever need anything else
+copy_input(::typeof(right_null), A) = copy_input(lq_null, A) # do we ever need anything else
 
 function check_input(::typeof(left_orth!), A::AbstractMatrix, VC)
     m, n = size(A)
@@ -81,7 +81,7 @@ end
 
 # Implementation of orth functions
 # --------------------------------
-function left_orth!(A::AbstractMatrix, VC; trunc=nothing,
+function left_orth!(A, VC; trunc=nothing,
                     kind=isnothing(trunc) ? :qr : :svd, alg_qr=(; positive=true),
                     alg_polar=(;), alg_svd=(;))
     check_input(left_orth!, A, VC)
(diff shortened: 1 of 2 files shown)


mtfishman:
@lkdvos @Jutho this is ready for review.

The motivation here is both improved code organization and generality (it removes some `AbstractMatrix` type constraints) and to provide lower level functions like `left_orth_qr!` that are an extra layer of customizability, for example they can be overloaded for GPU arrays, block sparse arrays, etc. where the current design may not be general enough.

The main considerations I can think of are:
1. Making those functions public. I'm leaning towards holding off on that to see how they get used in practice.
2. Should the `select_algorithm` logic be inside the lower level functions `left_orth_qr!`, or outside? If they are inside, it provides a bit more customizability, while if they are outside it means the implementations of functions like `left_orth_qr!` can be simpler. Again, I'm leaning towards keeping the implementation as-is for now to see how this is used in practice.

src/implementations/orthnull.jl
```diff
@@ -81,71 +81,87 @@ end
...
+    return qr_compact!(A, VC, alg′)
+end
+function left_orth_polar!(A, VC, alg)
+    alg′ = select_algorithm(left_polar!, A, alg)
+    return left_polar!(A, VC, alg′)
+end
+function left_orth_svd!(A, VC, alg, trunc::Nothing=nothing)
+    alg′ = select_algorithm(svd_compact!, A, alg)
+    V, C = VC
+    S = Diagonal(initialize_output(svd_vals!, A, alg′))
```
Jutho, line 112: Unrelated to this PR, but this part of the implementation is again very much `(Abstract)Matrix` specific, whereas `left_orth_qr!` and `left_orth_polar!` are quite generic and probably work for other types as well.

Jutho: Maybe a more general implementation can be added that is less economic with its memory usage:
```julia
function left_orth_svd!(A, VC, alg, trunc::Nothing=nothing)
    alg′ = select_algorithm(svd_compact!, A, alg)
    U, S, Vᴴ = svd_compact!(A, alg′)
    V, C = VC
    return copy!(V, U), mul!(C, S, Vᴴ)
end
```

Jutho (dismissed):
I definitely approve of splitting the different branches into separate functions that can be further specialized upon, and in omitting the `AbstractMatrix` restriction where possible. But as mentioned in the comments, it would still be good to add this restriction in the methods that use specific matrix constructions such as `Diagonal`.

src/implementations/orthnull.jl
mtfishman: Good point, I wasn't being careful about the type restriction in this case. I think that's a good suggestion to have a more generic version.

Another thing I can think of is replacing `Diagonal` with `diagonal`, and then `diagonal` could be an interface function that types can overload to construct a diagonal matrix-like object. But I like your suggestion better.

Jutho: Well, internally there is already the `diagview` to have a uniform interface to extracting a view of the diagonal of both `Matrix` and `Diagonal`, but that could also be specialized for other types. So a type-agnostic construction of a `diagonal` matrix could also be useful. With `diagonal` as name, it might get confusing though, as `LinearAlgebra.diag` is exactly the opposite, namely extracting the diagonal of a matrix and returning it as a vector. But I wouldn't have another naming suggestion.

I am also not sure if we generally want that the output type of `svd_vals` can be used to construct the `S` output of `svd_compact`. In TensorKit tensors, for example, a `DiagonalTensorMap` would still store all the singular values in a single list (`Vector`), but with internal structure such that it is known which parts of this vector are associated with which sectors/quantum numbers. `svd_vals` would than rather return that information as a `Dict` where for every sector (being the keys into the dict) there is a separate vector with only the singular values for that sector.

mtfishman: Yeah, it definitely would require some work to make the design based on `diagview`/`diagonal` work for exotic types like that.

I've updated the code to add a non-AbstractMatrix codepath based on your suggestion above. I also added a test that defines a simple non-AbstractMatrix to test that codepath.

Jutho (approved):
I think this is ready? I have no further comments.

PR #28, opened 2025-05-16
Allow customizing `truncate!` based on the algorithm

### LINKED ISSUE #27: Add the algorithm as an argument to `truncate!` for more customization
Add the algorithm as an argument to `truncate!` for more customization. Different algorithm backends might output the spectrum in different ways and therefore require different truncation implementations.

See also the discussion in #26.

mtfishman:
Closes #27.

The idea here is that this allows customizing the truncation based on the algorithm if needed, in addition to the truncation strategy. The reasoning is that certain algorithm backends might output the spectrum in a different format/ordering, which would require a different implementation of the truncation.

I'm open to suggestions on other ways of implementing this. Right now it is based on passing the `TruncatedAlgorithm` object into `truncate!`, which includes both the algorithm and the truncation strategy, and then by default it unwraps just the truncation strategy. An alternative could be to pass the truncation strategy and algorithm as separate arguments, though in general they might not be separate, for example in a sparse/low rank factorization like a randomized SVD or Krylov SVD.

### DIFF
--- a/src/implementations/eig.jl
+++ b/src/implementations/eig.jl
@@ -84,5 +84,5 @@ end
 
 function eig_trunc!(A::AbstractMatrix, DV, alg::TruncatedAlgorithm)
     D, V = eig_full!(A, DV, alg.alg)
-    return truncate!(eig_trunc!, (D, V), alg.trunc)
+    return truncate!(eig_trunc!, (D, V), alg)
 end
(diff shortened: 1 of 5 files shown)


mtfishman:
I talked to @lkdvos about this and we decide against this for now, we can bring it back if there is a compelling use case.

PR #30, opened 2025-05-27
Refactor algorithm selection in type domain

### LINKED ISSUE #29: Allow specializing `default_algorithm`/`select_algorithm` on the matrix type
As a followup to #23, it would be helpful to be able to specialize `default_algorithm`/`select_algorithm` on the matrix type in addition to specializing it on the instance.

An application of this is block sparse matrices, where we might want to determine the default algorithm based on the block type (as opposed to actually grabbing an instance of a block, which is finicky and doesn't work in the limit where there aren't any blocks).

lkdvos:
This alters the algorithm selection and default algorithms to be implemented in the type domain instead.
While doing this, I also realized that the new `default_algorithm(f, ...)` and `default_f_alg` are tautologous, so I removed the latter.
Given that these were not public, and the implementations in the "value domain" still dispatch to the type domain, I think this is non-breaking, so should be v0.2.1.

Fixes #29.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MatrixAlgebraKit"
 uuid = "6c742aac-3347-4629-af66-fc926824e5e4"
 authors = ["Jutho <jutho.haegeman@ugent.be> and contributors"]
-version = "0.2.0"
+version = "0.2.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -36,4 +36,5 @@ TestExtras = "5ed8adda-3752-4e41-b88a-e8b09835ee3a"
 Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f"
 
 [targets]
-test = ["Aqua", "JET", "SafeTestsets", "Test", "TestExtras","ChainRulesCore", "ChainRulesTestUtils", "StableRNGs", "Zygote"]
+test = ["Aqua", "JET", "SafeTestsets", "Test", "TestExtras", "ChainRulesCore",
+        "ChainRulesTestUtils", "StableRNGs", "Zygote"]
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -73,8 +73,7 @@ automatically with [`MatrixAlgebraKit.default_algorithm`](@ref) and
 the keyword arguments in `kwargs` will be passed to the algorithm constructor.
 Finally, the same behavior is obtained when the keyword arguments are
 passed as the third positional argument in the form of a `NamedTuple`. 
-"""
-function select_algorithm end
+""" select_algorithm
 
 function select_algorithm(f::F, A, alg::Alg=nothing; kwargs...) where {F,Alg}
     return _select_algorithm(f, A, alg; kwargs...)
@@ -109,8 +108,7 @@ end
 Select the default algorithm for a given factorization function `f` and input `A`.
 In general, this is called by [`select_algorithm`](@ref) if no algorithm is specified
 explicitly.
-"""
-function default_algorithm end
+""" default_algorithm
 
 @doc """
     copy_input(f, A)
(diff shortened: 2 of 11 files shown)


mtfishman:
Looks good to me, this will be helpful.

lkdvos:
The main point where this appeared was in trying to select a default algorithm for either TensorMap or BlockSparseArrays: I wanted to reuse the `default_algorithm` that is defined for the individual blocks, but these might not be present (an annoying edge-case is for example an empty block-sparse array), or not so easy to obtain.

I somehow assumed this solution was the easiest, since both TensorMap and BlockSparseArrays can quite easily tell you their blocktype, but an alternative solution is to simply define a `DefaultAlgorithm` dummy type to effectively delay choosing the algorithm type, similar to how TensorOperations handles this. I'm definitely open to either (or both?).
In principle currently the code should still work in the value domain, so if a type really wants to do runtime checks in order to alter the default algorithm selection that could still be possible.

I'll reintroduce the other `default_...` functions, I see what you mean.

mtfishman:
> > This alters the algorithm selection and default algorithms to be implemented in the type domain instead.
> 
> While I am not opposed to this change, I was wondering if this helps with anything? Semantically, I guess it makes sense that the default algorithm only depends on the type of `A` and not on the values, but that was what you anyway do with multiple dispatch if you define new methods for a given combination of types, in this case mainly the type of `A`.

The use case we have in mind is a block diagonal matrix, where you might want to extract the type of the block to determine the default algorithm for the blocks, i.e.:
```julia
default_algorithm(f::typeof(svd_compact!), A::BlockDiagonalMatrix) = BlockDiagonalAlgorithm(default_algorithm(f, blocktype(A))
```
It's a bit easier to reason about working in the type domain, rather than grabbing instances of blocks.

> Furthermore, there might actually be cases where the default algorithm selection does want to depend on properties of `A` that are not encoded in the type, e.g. `isposdef` or `ishermitian` (I don't like the `Hermitian` wrapper from LinearAlgebra.jl very much, and we don't have it for `TensorMap` objects.)

That's true, but from what I can tell this PR still allows customizing based on instances if you want to.

> > While doing this, I also realized that the new `default_algorithm(f, ...)` and `default_f_alg` are tautologous, so I removed the latter.
> 
> Not really, as the `default_f_algorithm` was another level of where to specify default algorithms in a way that affected all methods related to the decomposition in `f`, so e.g. all of `qr_full(!)`, `qr_compact(!)` and `qr_null(!)`. While it was perhaps not yet declared as a public and documented method, I did find it a useful entry point, for example in the (WIP) CUDA extension. It is an easy enough change to the new style of this PR, but it looked natural that one would often want to specify a new default for all methods related to the same mathematical decomposition.

I agree with this, I think it is a bit easier to reason about code like:
```julia
default_svd_algorithm(A) = LAPACK_DivideAndConquer()
default_algorithm(::typeof(svd_compact!), A) = default_svd_algorithm(A)
default_algorithm(::typeof(svd_full!), A) = default_svd_algorithm(A)
```
rather than:
```julia
default_algorithm(::typeof(svd_compact!), A) = LAPACK_DivideAndConquer()
default_algorithm(::typeof(svd_full!), A) = default_algorithm(svd_compact!, A)
```
since you can more easily remember that the actual definition is `default_svd_algorithm`, rather than having a convention that `svd_compact!` is the "core" definition, which could be documented but is a bit arbitrary.

mtfishman:
> I somehow assumed this solution was the easiest, since both TensorMap and BlockSparseArrays can quite easily tell you their blocktype, but an alternative solution is to simply define a `DefaultAlgorithm` dummy type to effectively delay choosing the algorithm type, similar to how TensorOperations handles this. I'm definitely open to either (or both?). In principle currently the code should still work in the value domain, so if a type really wants to do runtime checks in order to alter the default algorithm selection that could still be possible.

A `DefaultAlgorithm` dummy type may be best anyway, for example that could handle the case where there are different types of blocks. In that case, the block type wouldn't be that helpful anyway, since it would be some abstract supertype of the block types.

lkdvos:
I've reinstated the desired functions and opened an issue to remind myself to add the default algorithm, which I'll implement in a separate PR.

Jutho (approved):

Jutho:
Yes looks good to me, feel free to merge.

PR #38, opened 2025-06-12
Add `isisometry` function and integrate into tests

lkdvos:
Introduce the `isisometry` function to determine if a linear map is an isometry, and update existing tests to utilize this new function.

### DIFF
--- a/src/common/matrixproperties.jl
+++ b/src/common/matrixproperties.jl
@@ -1,10 +1,10 @@
 """
-    isisometry(A; kind=:left, isapprox_kwargs...) -> Bool
+    isisometry(A; side=:left, isapprox_kwargs...) -> Bool
 
 Test whether a linear map is an isometry, where the type of isometry is controlled by `kind`:
 
-- `kind = :left` : `A' * A ≈ I`. 
-- `kind = :right` : `A * A` ≈ I`.
+- `side = :left` : `A' * A ≈ I`. 
+- `side = :right` : `A * A` ≈ I`.
 
 The `isapprox_kwargs` are passed on to `isapprox` to control the tolerances.
 
@@ -12,11 +12,11 @@ New specializations should overload [`is_left_isometry`](@ref) and [`is_right_is
 
 See also [`isunitary`](@ref).
 """
-function isisometry(A; kind::Symbol=:left, isapprox_kwargs...)
-    kind === :left && return is_left_isometry(A; isapprox_kwargs...)
-    kind === :right && return is_right_isometry(A; isapprox_kwargs...)
+function isisometry(A; side::Symbol=:left, isapprox_kwargs...)
+    side === :left && return is_left_isometry(A; isapprox_kwargs...)
+    side === :right && return is_right_isometry(A; isapprox_kwargs...)
 
-    throw(ArgumentError(lazy"Invalid isometry kind: $kind"))
+    throw(ArgumentError(lazy"Invalid isometry side: $side"))
 end
 
 """
--- a/src/common/matrixproperties.jl
+++ b/src/common/matrixproperties.jl
@@ -1,0 +1,7 @@
+"""
+    isisometry(A; kwargs...) -> Bool
+
+Test whether a linear map is an isometry, ie `A' * A ≈ I`. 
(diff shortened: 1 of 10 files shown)


src/common/matrixproperties.jl
```diff
@@ -0,0 +1,7 @@
+"""
+    isisometry(A; kwargs...) -> Bool
+
+Test whether a linear map is an isometry, ie `A' * A ≈ I`. 
+The `kwargs` are passed on to `isapprox` to control the tolerances.
+"""
+isisometry(A::AbstractMatrix; kwargs...) = isapprox(A' * A, LinearAlgebra.I; kwargs...)
```
mtfishman, line 7: What do you think of introducing `is_left_isometry` and `is_right_isometry` and then define `isisometry` as a shorthand? I realize you can just call `isisometry(A')` as you do in the tests but it could allow for some more customization for exotic types and also matches with the generally naming schemes in the library (i.e. `left_orth`/`right_orth`, `left_null`/`right_null`, etc.).

mtfishman: In TensorAlgebra.jl I defined some wrappers like `orth(A; side=:left) -> left_orth(A)` and `orth(A; side=:right) -> right_orth(A)` so that would match that convention, it could be nice to establish a convention like that in MatrixAlgebraKit.jl that we could follow elsewhere. I couldn't think of anything better in terms of an interface for that code pattern, and for whatever it's worth it matches the interface of [SciPy's polar function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.polar.html).

I think it makes sense to have `is_left_isometry` and `is_right_isometry` as implementations and lower level interfaces to overload, and then `isisometry(A; side=:left)` and `isisometry(A; side=:right)` could be higher level layers on top that are meant to be called by users. I think a Symbol makes sense, I think a Bool would be too terse (what would the keyword be?) and an enum would require exposing that as a new object in the namespace which feels a bit heavy.

Jutho:
I like how this shows that this was clearly very badly needed for the tests.

src/common/matrixproperties.jl
lkdvos: I'll look into implementing this tomorrow. With respect to the AbstractArray implementation, I think my idea was that the function should apply to generic linear maps, as most of our functions do, while the method in terms of LinearAlgebra.I seems reasonable to restrict to AbstractMatrix. I'll also change this, so the docstring applies to the generic case while the implementation is separate.

lkdvos: I actually prefer `isisometry` and `isunitary`, but that might be because I'm used to `LinearAlgebra.ishermitian`, `LinearAlgebra.isposdef` etc. On the other hand, I do agree that `isleftisometry` is too long, so that would better be `is_left_isometry`...
The actual horror of programming, coming up with names 😱

mtfishman: I'm definitely fine with `isisometry` and `isunitary` (I thought `isunitary` would be harder to read but I think it is fine), I think the inconsistency with `is_left_isometry` is also fine since that is meant to be mostly internal/for developers anyway.

mtfishman: @lkdvos I see you went with the name `kind` for the keyword argument, which matches `left_orth(A; kind=:svd)`. I slightly prefer `side` since it is more specific, and additionally it leaves open the possibility of having an interface like `orth(A; kind=:svd, side=:right)` (i.e. `kind` can be used for selecting a backend and `side` can used for selecting a direction). What do you think?

src/common/matrixproperties.jl
```diff
@@ -0,0 +1,59 @@
+"""
+    isisometry(A; side=:left, isapprox_kwargs...) -> Bool
+
+Test whether a linear map is an isometry, where the type of isometry is controlled by `kind`:
+
+- `side = :left` : `A' * A ≈ I`. 
+- `side = :right` : `A * A` ≈ I`.
```
Jutho, line 7: That seems to be the wrong symbol for adjoint:
```suggestion
- `side = :right` : `A * A' ≈ I`.
```

Jutho (approved):

PR #39, opened 2025-08-07
Add generalized eigenvalue decomposition, fix some bugs

kshyatt:
Tried this as a way to learn about the structure of the package, so please feel free to aggressively critique it :).

I've added supported for [generalized eigendecomposition](https://en.wikipedia.org/wiki/Eigendecomposition_of_a_matrix#Generalized_eigenvalue_problem) over two inputs, `A` and `B`. All this arose out of a desire to add coverage to the `ggev` methods. I fixed some errors in the `ccall` wrapper and tried to implement the interface in a semi-reasonable way. Annoyingly, LAPACK doesn't provide the support for `LAPACK_Expert` for this...


src/MatrixAlgebraKit.jl
```diff
@@ -19,6 +19,9 @@ export eigh_full, eigh_vals, eigh_trunc
 export eigh_full!, eigh_vals!, eigh_trunc!
 export eig_full, eig_vals, eig_trunc
 export eig_full!, eig_vals!, eig_trunc!
+export gen_eig, gen_eig!
+export gen_eig_full, gen_eig_vals
+export gen_eig_full!, gen_eig_vals!
```
lkdvos, line 24: Just because I don't know this: is there a distinction between hermitian/nonhermitian versions, just like the `eig`-`eigh` ?

src/algorithms.jl
```diff
@@ -167,13 +171,51 @@ the final `f!(A, out, alg::Algorithm)`.
...
+        f_n_args = val
+    end
+
+    f = args[end]
     f isa Symbol || throw(ArgumentError("Unsupported usage of `@functiondef`"))
     f! = Symbol(f, :!)
 
-    ex = quote
+    # TODO is the right way?
+    @gensym A B
```
lkdvos, line 217: I'm not sure if this is necessary, I think the scoping rules always make sure that the `A` and `B` refer to the function arguments. (Also I don't think you are using these right now, since it would have to be interpolated in the quoted expression)

src/algorithms.jl
```diff
@@ -215,7 +257,53 @@ macro functiondef(f)
...
+        end
+
+        # copy documentation to both functions
+        Core.@__doc__ $f, $f!
+    end
+    if f_n_args == 1
+        return esc(ex_single_arg)
+    elseif f_n_args == 2
+        return esc(ex_double_arg)
+    end
```
lkdvos, line 306: It might be cleaner to put this if-clause before the actual expressions are constructed, and only construct a single one?

src/yalapack.jl
```diff
@@ -1733,6 +1738,37 @@ function _reorder_realeigendecomposition!(W, WR, WI, work, VR, jobvr)
     end
 end
 
+function _reorder_realeigendecomposition!(W, WR, WI, Wbeta, work, VR, jobvr)
```
lkdvos, line 1741: Possibly can use inbounds annotations in this function, not sure if this matters?

lkdvos (commented):
Overall looks really good to me, and definitely in the same style as the rest so I think you nailed it!

I like the generalization of the macro to multiple arguments. Given that we only have a single keyword supported and this is mostly an internal thing anyways, I would even be fine with dropping the specification altogether: `@functiondef 2 f` seems also good to me. This ultimately doesn't really matter though so I have no strong opinions here.

As an alternative, an idea that might also work is to simply bundle the arguments together: `gen_eig!((A, B), ...)` might actually require less work for altering the rest of the functionality, but I'm not sure if it is that much more convenient, nor easy to remember/read etc. Just wanted to share the idea, definitely feel free to let me know what you think about it because I'm not convinced it is better.

For the distinction between `gen_eig` and `gen_eig_full`, I assume this is because for `eig` and `eig_full` we have a similar function laying around somewhere. I think this is an idea that made it in at some point and was never properly worked out, I'm not sure if we really need both (I would personally prefer if we could get rid of one of them even).
For `eig`, we only export the `eig_full` variants, so I would say to do the same here, and at some point just remove `eig` etc altogether?

src/yalapack.jl
kshyatt: It might for large arrays

src/algorithms.jl
kshyatt: Ah yes whoops 😅

src/MatrixAlgebraKit.jl
kshyatt: No, there's no specialized LAPACK support for *generalized* eigenproblems for symmetric or Hermitian matrices, only the general (stupid naming system...) case is supported.

src/algorithms.jl
lkdvos: definitely agreed!

kshyatt: Done! Used a `Val` so we can dispatch nicely and would work with switching to dispatching from an `NTuple` in the future.

src/interface/gen_eig.jl
```diff
@@ -0,0 +1,78 @@
+# Gen Eig API
+# -------
+function gen_eig!(A::AbstractMatrix, B::AbstractMatrix, args...; kwargs...)
+    return gen_eig_full!(A, B, args...; kwargs...)
+end
+function gen_eig(A::AbstractMatrix, B::AbstractMatrix, args...; kwargs...)
+    return gen_eig_full(A, B, args...; kwargs...)
+end
```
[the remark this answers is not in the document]
lkdvos: Yes, I think that's a good idea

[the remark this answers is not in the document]
kshyatt: Already done!

test/gen_eig.jl
```diff
@@ -0,0 +1,62 @@
...
+
+        Ac = similar(A)
+        Bc = similar(B)
+        W2, V2 = @constinferred gen_eig_full!(copy!(Ac, A), copy!(Bc, B), (W, V), alg′)
+        @test W2 === W
+        @test V2 === V
+        @test A * V ≈ B * V * Diagonal(W)
+
+        Ac = similar(A)
+        Bc = similar(B)
```
Jutho, line 33: I guess the previous `Ac` and `Bc` could have been recycled.

src/yalapack.jl
```diff
@@ -1695,7 +1696,11 @@ for (gees, geesx, geev, geevx, ggev, elty, celty, relty) in
 
             # Cleanup the output in the real case
             if eltype(A) <: Real
-                _reorder_realeigendecomposition!(W, WR, WI, work, VR, jobvr)
+                _reorder_realeigendecomposition!(W, WR, WI, Wbeta, work, VR, jobvr)
```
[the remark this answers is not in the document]
lkdvos: I guess the extra maintenance overhead is also pretty negligable, so I'll just leave this as-is for now and resolve this conversation.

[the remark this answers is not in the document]
Jutho: But isn't it literally the same function line by line, aside for the extra division? I am happy to file a PR for this, but first wanted to hear if I had overlooked something or other comments?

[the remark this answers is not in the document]
kshyatt: Yes it is the same function, which was set up that way for convenience at the time. We can certainly do another PR to unify them.

PR #41, opened 2025-08-12
[WIP] Attempting to wrap randomized SVD

kshyatt:
Attempted to wrap the randomized SVD method which relies on oversampling and the power method. Definitely open to some interface improvements here! Also, the ugly look of the tests suggests perhaps we should extend `diagview` a little...

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
@@ -247,6 +247,66 @@ for (bname, fname, elty, relty) in
     end
 end
 
+# Wrapper for randomized SVD
+function Xgesvdr!(A::StridedCuMatrix{T},
+                   S::StridedCuVector=similar(A, real(T), min(size(A)...)),
+                   U::StridedCuMatrix{T}=similar(A, T, size(A, 1), min(size(A)...)),
+                   Vᴴ::StridedCuMatrix{T}=similar(A, T, min(size(A)...), size(A, 2));
+                   k::Int=length(S),
+                   p::Int=min(size(A)...)-k-1,
+                   niters::Int=1) where {T<:BlasFloat}
+    chkstride1(A, U, S, Vᴴ)
+    m, n = size(A)
+    minmn = min(m, n)
+    jobu = length(U) == 0 ? 'N' : 'S'
+    jobv = length(Vᴴ) == 0 ? 'N' : 'S'
+    k = min(size(S)...)
+    R = eltype(S)
+    k < minmn || throw(DimensionMismatch("length of S ($k) must be less than the smaller dimension of A ($minmn)"))
+    k + p < minmn || throw(DimensionMismatch("length of S ($k) plus oversampling ($p) must be less than the smaller dimension of A ($minmn)"))
+    R == real(T) ||
+        throw(ArgumentError("S does not have the matching real `eltype` of A"))
+
+    Ṽ = similar(Vᴴ, (n, n))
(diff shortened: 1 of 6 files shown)


Jutho:
I haven't looked at this in detail yet. Do you think `CUSOLVER_Randomized` should be a valid argument for `svd_full!` given the warning note in the doc string? Wouldn't this make more sense as a `svd_trunc!` algorithm ?

kshyatt:
Could well do better there since for best performance/accuracy `k` should be much much smaller than `min(m,n)`. I can switch it in the morning.

kshyatt:
The reason I didn't do this at first I think is that `Xgesvdr` expects `U`, `S`, and `V` to be *full* sized, which is rather annoying...

kshyatt:
If we can modify `check_input` then that seems fine to me! I could do that as part of this PR, even?

ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
```diff
@@ -242,11 +242,70 @@ for (bname, fname, elty, relty) in
...
         end
     end
 end
 
+# Wrapper for randomized SVD
+function Xgesvdr!(A::StridedCuMatrix{T},
+                   S::StridedCuVector=similar(A, real(T), min(size(A)...)),
+                   U::StridedCuMatrix{T}=similar(A, T, size(A, 1), min(size(A)...)),
+                   Vᴴ::StridedCuMatrix{T}=similar(A, T, min(size(A)...), size(A, 2));
+                   k::Int=length(S),
```
Jutho, line 255: Would this work? The Xgesvdr docs say that `S` needs to have length `min(size(A)...)`, but also that `k` needs to be (strictly?) less than `min(size(A)...)`. I find the whole CUSOLVER Xgesvdr interface and documentation horrible. In the general explanation part, it mentions $$U$$ as an $$m \times k $$ matrix, and similar for $$\Sigma$$ and $$V$$, and then in the arguments it requires `S`, `U` and `V` to be full size and doesn't mention `k` again:
Screenshot 2025-09-03 at 12 24 52

Jutho: This makes me even wonder if `U` really needs to be full size, or only of size $$m \times k$$. Or $$m \times (k+p)$$. I don't understand these interfaces; if `U` is an output variable, why does it need to have a different size than the targetted output. Can't they just use `workspace` for whatever else is needed?

src/interface/decompositions.jl
```diff
@@ -151,6 +151,19 @@ a general matrix using the Jacobi algorithm.
 """
 @algdef CUSOLVER_Jacobi
 
+"""
+    CUSOLVER_Randomized(; p, niters)
```
Jutho, line 155: This misses `k` as a keyword argument, as well as an explanation of `p` and `niters`.

ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
kshyatt: It's *very* confusing. I think the idea is that `U` should have size `m x m` **but** only the first `k` columns are actually "valid results"

Jutho: I also suspect it as such, but clearly this part of the documentation has just been copied from other SVD methods without adapting, which makes me wonder whether the suggestion that `U` needs to have size `m x m` for this to work is correct or also just erroneously copied 😄 .

PR #47, opened 2025-08-20
Initial attempt to wrap CUSOLVER.Xgeev

kshyatt:
Had to write some kernels for the reordering which could certainly be improved but do seem to work for now.

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
@@ -2,10 +2,10 @@ module YACUSOLVER
 
 using LinearAlgebra
 using LinearAlgebra: BlasInt, BlasFloat, checksquare, chkstride1, require_one_based_indexing
-using LinearAlgebra.LAPACK: chkargsok, chklapackerror, chktrans, chkside, chkdiag, chkuplo
+using LinearAlgebra.LAPACK: chkargsok, chklapackerror, chktrans, chkside, chkdiag, chkuplo#, chkfinite
 
 using CUDA
-using CUDA: @allowscalar
+using CUDA: @allowscalar, i32
 using CUDA.CUSOLVER
 
 # QR methods are implemented with full access to allocated arrays, so we do not need to redo this:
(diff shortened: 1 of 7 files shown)


lkdvos (commented):
Looks great! At some point we might want to look into actually formalizing the reordering of eigenvalues in some way and add that to the features, but this definitely looks good!

I left some small comments, but otherwise I'm happy to merge this.

ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
```diff
@@ -306,6 +306,73 @@ function Xgesvdr!(A::StridedCuMatrix{T},
...
+            n = checksquare(A)
+            # TODO GPU appropriate version
+            #chkfinite(A) # balancing routines don't support NaNs and Infs
+            n == length(D) || throw(DimensionMismatch("length mismatch between A and D"))
+            if length(V) == 0
+                jobvr = 'N'
+            elseif length(V) == n*n
+                jobvr = 'V'
+            else
+                throw(DimensionMismatch("size of VR must match size of A"))
```
Jutho, line 324: I guess we just call it `V` instead of `VR` in our function signatures and doc strings.

ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
```diff
@@ -306,6 +306,73 @@ function Xgesvdr!(A::StridedCuMatrix{T},
...
+            lda  = max(1, stride(A, 2))
+            ldvl = max(1, stride(VL, 2))
+            params = CUSOLVER.CuSolverParameters()
+            dh = CUSOLVER.dense_handle()
+            
+            if $elty <: Real
+                D2 = reinterpret($elty, D)
+                # reuse memory, we will have to reorder afterwards to bring real and imaginary
+                # components in the order as required for the Complex type
+                VR = reinterpret($elty, V)
```
Jutho, line 344: It seems like `cusolverDnXgeev` supports inputting complex `W` and `VR` arguments, and then immediately returns the results as we want them, making the whole `reinterpret` and `_reorder_realeigendecomposition` (forced upon us by LAPACK) unneccessary: https://docs.nvidia.com/cuda/cusolver/index.html?highlight=cusolverDnXgeev#cusolverdnxgeev

kshyatt: I don't think this is true for the eigen**vectors** from testing I did. For `VL`:

> Array of dimension ldvl * n. If jobvl = CUSOLVER_EIG_MODE_VECTOR, the left eigenvectors u(j) are stored one after another in the columns of VL, in the same order as their eigenvalues. If datatypeVL is complex or the j-th eigenvalue is real, then u(j) = VL(:,j), the j-th column of VL. If dataTypeVL is real and the j-th and (j+1)-st eigenvalues form a complex conjugate pair, then u(j) = VL(:,j) + i*VL(:,j+1) and u(j+1) = VL(:,j) - i*VL(:,j+1). If jobvl = CUSOLVER_EIG_MODE_NOVECTOR, VL is not referenced.

kshyatt: The datatype of `VR` **must** be same as the datatype of `A` though per the table further down

Jutho: That is completely ridiculous. Who comes up with writing the documentation in such a confusing way. Or why would you allow complex `W` but then not still use the outdated lapack scheme for `V`. It feels like CUSOLVER managed to make the worst aspects of the LAPACK interface even worse.

Jutho: Ok, given what we have, it might still be possible to simplify the implementation a bit by using the complex eigenvalue vector interface. If possible, I would prefer having to maintain CUDA kernels in MatrixAlgebraKit (even though this likely never needs to be touched again). For example, I wouldn't actually mind assigning a temporary real `VR` that is distinct from the `V` in the arguments, and then simply copying over, but maybe it's possible without. But also interested in @lkdvos opinion. I'll might give it a try and then we can still decide what to do with this.

Jutho: AMD currently doesn't seem to have general eigenvalue decomposition, but when they do, who knows what interface they will adopt.

kshyatt: Technically this is all CUSOLVER, btw, not cuTENSOR

kshyatt: Also FWIW historically AMD tends to hew very close to the line CUDA adopts for easily compatibility, since their hip* libraries allow users to "drop in" AMD to replace CUSOLVER/CUBLAS/CUWHATEVER

PR #49, opened 2025-08-26
Updates for TensorKit compatibility

kshyatt:
The `ReshapedArray` overrides are needed to dispatch to the correct GPU algorithms. Needed to modify the type signature for the default algorithms to avoid ambiguities. Also it's nice to give some more info about dimension mismatches.

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -13,8 +13,6 @@ using CUDA: i32
 using LinearAlgebra
 using LinearAlgebra: BlasFloat
 
-using CUDA: i32
-
 include("yacusolver.jl")
 
 function MatrixAlgebraKit.default_qr_algorithm(::Type{T}; kwargs...) where {TT <: BlasFloat, T <: StridedCuMatrix{TT}}
--- a/src/implementations/eig.jl
+++ b/src/implementations/eig.jl
@@ -137,7 +137,7 @@ function eig_full!(A::AbstractMatrix, DV, alg::GPU_EigAlgorithm)
     D, V = DV
     if alg isa GPU_Simple
         isempty(alg.kwargs) ||
-            throw(ArgumentError("GPU_Simple (geev) does not accept any keyword arguments"))
+            @warn "GPU_Simple (geev) does not accept any keyword arguments"
         _gpu_geev!(A, D.diag, V)
     end
     # TODO: make this controllable using a `gaugefix` keyword argument
@@ -149,9 +149,8 @@ function eig_vals!(A::AbstractMatrix, D, alg::GPU_EigAlgorithm)
     check_input(eig_vals!, A, D, alg)
     V = similar(A, complex(eltype(A)), (size(A, 1), 0))
     if alg isa GPU_Simple
-        # TODO filter out nothing kwargs
-        #isempty(alg.kwargs) ||
-        #    throw(ArgumentError("GPU_Simple (geev) does not accept any keyword arguments"))
+        isempty(alg.kwargs) ||
+            @warn "GPU_Simple (geev) does not accept any keyword arguments"
         _gpu_geev!(A, D, V)
     end
     return D
--- a/test/cuda/projections.jl
+++ b/test/cuda/projections.jl
@@ -12,7 +12,7 @@ const BLASFloats = (Float32, Float64, ComplexF32, ComplexF64)
     m = 54
     noisefactor = eps(real(T))^(3 / 4)
     for alg in (NativeBlocked(blocksize = 16), NativeBlocked(blocksize = 32), NativeBlocked(blocksize = 64))
-        A  = CuArray(randn(rng, T, m, m))
+        A = CuArray(randn(rng, T, m, m))
         Ah = (A + A') / 2
         Aa = (A - A') / 2
         Ac = copy(A)
@@ -69,7 +69,7 @@ end
             # test that W is closer to A then any other isometry
             for k in 1:10
                 δA = CuArray(randn(rng, T, m, n))
-                W  = project_isometric(A, alg)
+                W = project_isometric(A, alg)
                 W2 = project_isometric(A + δA / 100, alg)
                 @test norm(A - W2) > norm(A - W)
             end


src/implementations/lq.jl
```diff
@@ -181,7 +181,7 @@ function lq_via_qr!(A::AbstractMatrix, L::AbstractMatrix, Q::AbstractMatrix,
                     qr_alg::AbstractAlgorithm)
     m, n = size(A)
     minmn = min(m, n)
-    At = adjoint!(similar(A'), A)::AbstractMatrix
+    At = min(m, n) > 0 ? adjoint!(similar(A'), A)::AbstractMatrix : similar(A')
```
Jutho, line 184: Is this `::AbstractMatrix` type assert useful?

kshyatt:
Changed the format of some of the `orthnull` tests -- `CuArray` and `UniformScaling` and `isapprox` *really* don't play nicely together. Also added a path to use `CUSOLVER_QRIteration` even in the case where `m < n` via transpose. Happy for feedback on the implementation - should this be a separate algorithm?

src/implementations/svd.jl
```diff
@@ -288,7 +288,7 @@ function MatrixAlgebraKit.svd_compact!(A::AbstractMatrix, USVᴴ, alg::GPU_SVDAl
     U, S, Vᴴ = USVᴴ
     if alg isa GPU_QRIteration
         isempty(alg.kwargs) ||
-            throw(ArgumentError("GPU_QRIteration does not accept any keyword arguments"))
+            @warn "GPU_QRIteration does not accept any keyword arguments"
```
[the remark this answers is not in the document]
Jutho: I would need to check, but some of the LAPACK algorithm names are used for different factorisations, and might support keywords for one and not for the other, or might support a different set of keywords. We can also remove these explicit checks, just pass on `alg.kwargs` to the underlying LAPACK/CUSOLVER/rocSOLVER routine, and then have these complain if the set of keyword arguments is invalid.

src/common/matrixproperties.jl
```diff
@@ -42,6 +42,7 @@ See also [`isisometry`](@ref) and [`is_right_isometry`](@ref).
 """ is_left_isometry
 
 function is_left_isometry(A::AbstractMatrix; isapprox_kwargs...)
+    iszero(min(size(A)...)) && return true
```
Jutho, line 45: I would say: `iszero(size(A, 2)) && return true` for `is_left_isometry` and `iszero(size(A,1))` for `is_right_isometry below`.

src/implementations/svd.jl
```diff
@@ -245,17 +245,43 @@ _gpu_gesvd!(A::AbstractMatrix, S::AbstractVector, U::AbstractMatrix, Vᴴ::Abstr
...
 _gpu_gesvdj!(A::AbstractMatrix, S::AbstractVector, U::AbstractMatrix, Vᴴ::AbstractMatrix; kwargs...) = throw(MethodError(_gpu_gesvdj!, (A, S, U, Vᴴ)))
+function _gpu_gesvd_maybe_transpose!(A::AbstractMatrix, S::AbstractVector, U::AbstractMatrix, Vᴴ::AbstractMatrix)
+    m, n = size(A)
+    m ≥ n && return _gpu_gesvd!(A, S, U, Vᴴ)
+    # both CUSOLVER and ROCSOLVER require m ≥ n for gesvd (QR_Iteration)
+    # if this condition is not met, do the SVD via adjoint
+    minmn = min(m, n)
+    At = min(m, n) > 0 ? adjoint!(similar(A'), A)::AbstractMatrix : similar(A')
+    Ut = similar(U')
+    Vᴴt = similar(Vᴴ')
```
Jutho, line 256: Can we call these variables `Aᴴ`, `Uᴴ` and simple `V` ?

src/implementations/svd.jl
```diff
@@ -245,17 +245,43 @@ _gpu_gesvd!(A::AbstractMatrix, S::AbstractVector, U::AbstractMatrix, Vᴴ::Abstr
...
+    if size(U) == (m, m)
+        _gpu_gesvd!(At, view(S, 1:minmn, 1), Vᴴt, Ut)
+    else
+        _gpu_gesvd!(At, S, Vᴴt, Ut)
+    end
+    length(U) > 0 ? adjoint!(U, Ut) : one!(U)
```
Jutho, line 262: Why is the `one!(U)` necessary if `U` is empty, i.e. why not

```julia
length(U) > 0 && adjoint!(U, Ut)
```

src/implementations/svd.jl
```diff
@@ -245,17 +245,43 @@ _gpu_gesvd!(A::AbstractMatrix, S::AbstractVector, U::AbstractMatrix, Vᴴ::Abstr
...
+    length(Vᴴ) > 0 ? adjoint!(Vᴴ, Vᴴt) : one!(Vᴴ)
+    conj!(S)
```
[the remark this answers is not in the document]
kshyatt: Yes... the risks of making updates at 9pm at night

src/implementations/svd.jl
```diff
@@ -245,17 +245,43 @@ _gpu_gesvd!(A::AbstractMatrix, S::AbstractVector, U::AbstractMatrix, Vᴴ::Abstr
...
+    m, n = size(A)
+    m ≥ n && return _gpu_gesvd!(A, S, U, Vᴴ)
+    # both CUSOLVER and ROCSOLVER require m ≥ n for gesvd (QR_Iteration)
+    # if this condition is not met, do the SVD via adjoint
+    minmn = min(m, n)
+    At = min(m, n) > 0 ? adjoint!(similar(A'), A)::AbstractMatrix : similar(A')
+    Ut = similar(U')
+    Vᴴt = similar(Vᴴ')
+    if size(U) == (m, m)
+        _gpu_gesvd!(At, view(S, 1:minmn, 1), Vᴴt, Ut)
```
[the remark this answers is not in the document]
kshyatt: Here I'm following what the CPU bindings have done, but I suppose we could be reusing an S over and over between differently sized arrays?

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -8,30 +8,49 @@ using MatrixAlgebraKit: LQViaTransposedQR, TruncationByValue, AbstractAlgorithm
...
 import MatrixAlgebraKit: _gpu_geqrf!, _gpu_ungqr!, _gpu_unmqr!, _gpu_gesvd!, _gpu_Xgesvdp!, _gpu_Xgesvdr!, _gpu_gesvdj!, _gpu_geev!
 import MatrixAlgebraKit: _gpu_heevj!, _gpu_heevd!
-using CUDA
+using CUDA, CUDA.CUBLAS
 using CUDA: i32
 using LinearAlgebra
 using LinearAlgebra: BlasFloat
 
+using CUDA: i32
+
```
lkdvos:
```suggestion
```

I think this is already in the import list above

PR #50, opened 2025-09-02
Support for `Diagonal`

lkdvos:
Here I've added specialized implementations for working with `Diagonal`s, through a dedicated backend.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "MatrixAlgebraKit"
 uuid = "6c742aac-3347-4629-af66-fc926824e5e4"
 authors = ["Jutho <jutho.haegeman@ugent.be> and contributors"]
-version = "0.3.0"
+version = "0.3.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"


src/implementations/eigh.jl
```diff
@@ -85,6 +115,25 @@ function eigh_trunc!(A::AbstractMatrix, DV, alg::TruncatedAlgorithm)
     return truncate!(eigh_trunc!, (D, V), alg.trunc)
 end
 
+# Diagonal logic
+# --------------
+function eigh_full!(A::Diagonal, DV, alg::DiagonalAlgorithm)
+    check_input(eigh_full!, A, DV, alg)
+    D, V = DV
+    D === A || (diagview(D) .= real.(diagview(A)))
```
Jutho, line 123: Do we simply discard possible imaginary values on the diagonal? I guess we also don't check for hermiticity in the general matrix case. Do we want to add such a check in `check_input`?

lkdvos: I just double-checked and noticed that the LAPACK implementations are actually checking for symmetric/hermitian, so I'll adapt this here as well.

src/implementations/svd.jl
```diff
@@ -66,6 +94,18 @@ function initialize_output(::typeof(svd_trunc!), A::AbstractMatrix, alg::Truncat
     return initialize_output(svd_compact!, A, alg.alg)
 end
 
+function initialize_output(::typeof(svd_full!), A::Diagonal, ::DiagonalAlgorithm)
+    TA = eltype(A)
+    TUV = Base.promote_op(sign_safe, TA)
+    return similar(A, TUV, size(A)), similar(A, real(TA)), similar(A, TUV, size(A))
```
Jutho, line 100: If you want `Diagonal` `U` and `Vᴴ` there shouldn't be any `size` arguments in `similar`.

lkdvos: See other comment, I specifically don't want `Diagonal`s here.

src/implementations/svd.jl
```diff
@@ -203,7 +242,39 @@ function svd_trunc!(A::AbstractMatrix, USVᴴ, alg::TruncatedAlgorithm)
...
+    Ad = diagview(A)
+    U, S, Vᴴ = USVᴴ
+    Sd = diagview(S)
+    Sd .= abs.(Ad)
+    p = sortperm(Sd; rev=true)
+    permute!(Sd, p)
+    T = eltype(Vᴴ)
+    zero!(U)
+    zero!(Vᴴ)
+    @inbounds for (i, pi) in enumerate(p)
```
Jutho, line 258: The gauge fixing for the svd in the general matrix case is such that the largest element on every column of `U` is positive, and the sign/phase is absorved in `Vᴴ`. So, `U` should be a pure permutation matrix, and `Vᴴ` should incorporate the sign of the diagonal element.

Jutho: Once again, an implementation without scalar indexing would be great, but in this case I would also need to think about how to do that.

src/implementations/svd.jl
```diff
@@ -42,6 +44,32 @@ function check_input(::typeof(svd_vals!), A::AbstractMatrix, S, ::AbstractAlgori
     return nothing
 end
 
+function check_input(::typeof(svd_full!), A::AbstractMatrix, USVᴴ, ::DiagonalAlgorithm)
+    m, n = size(A)
+    @assert m == n && isdiag(A)
+    U, S, Vᴴ = USVᴴ
+    @assert U isa AbstractMatrix && S isa Diagonal && Vᴴ isa AbstractMatrix
```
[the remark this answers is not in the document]
mtfishman: `LinearAlgebra.eigen` supports custom sorting of eigenvalues: https://docs.julialang.org/en/v1/stdlib/LinearAlgebra/#LinearAlgebra.eigen, I would vote for supporting the same thing in MatrixAlgebraKit.jl in some form, though of course that would mean the matrix of eigenvectors would in general be a permutation matrix.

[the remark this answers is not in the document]
mtfishman: Maybe there could be a different algorithm backend such as `SortedEigvalAlgorithm` that post-processes the default `eigh_full!` and sorts in a specified way, and by default the eigenalue ordering of `eigh_full!` could be left arbitrary and therefore in the Diagonal case the eigenvector matrix could be Diagonal.

src/implementations/svd.jl
```diff
@@ -203,7 +242,40 @@ function svd_trunc!(A::AbstractMatrix, USVᴴ, alg::TruncatedAlgorithm)
...
+function svd_full!(A::AbstractMatrix, USVᴴ, alg::DiagonalAlgorithm)
+    check_input(svd_full!, A, USVᴴ, alg)
+    Ad = diagview(A)
+    U, S, Vᴴ = USVᴴ
+    p = sortperm(Ad; by=abs, rev=true)
+    zero!(U)
+    zero!(Vᴴ)
+    T = eltype(U)
+    U[CartesianIndex.(enumerate(p))] .= Ref(one(T))
+    Vᴴ[CartesianIndex.(reverse.(enumerate(p)))] .= sign_safe.(view(Ad, p))
```
Jutho, line 256: I don't know if we want to support `U` and `Vᴴ` that don't have efficient linear indexing. With that restriction in mind, one solution that would have slightly fewer allocations could be:
```julia
p = sortperm(Ad; by=abs, rev=true)
ip = p .+ (0:n-1) .* n
p .= (1:n) .+ (p .- 1) .* n
U[p] .= one(T)
Vᴴ[ip] .= sign_safe.(view(Ad, p))
```

Not sure if that is fully correct; I haven't checked/tested this, but I guess the idea is clear.

lkdvos: I'm slightly confused about this, doesn't my current implementation have no allocations since the entire broadcasting expression should fuse, while your suggested implementation has one additional vector allocated? I agree that the linear indexing might be more efficient, but it's hard to say without testing

Jutho (approved):

src/implementations/svd.jl
```diff
@@ -203,7 +242,46 @@ function svd_trunc!(A::AbstractMatrix, USVᴴ, alg::TruncatedAlgorithm)
...
+    Sd = diagview(S)
+    if Ad === Sd
+        @. Sd = abs(Ad)
+        permute!(Sd, p)
+    else
+        Sd .= abs.(view(Ad, p))
+    end
+
+    p .+= (0:(n - 1)) .* n
+    U[p] .= Ref(one(eltype(U)))
```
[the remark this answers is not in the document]
lkdvos: I think I was thinking when implementing this that since the `Diagonal` implementation really requires very little from the scalars, it might be that someone at some point uses this for some exotic type that actually has some unwanted broadcasting behavior, so better to be explicit about it. Realistically though, this totally doesn't matter 🙃

PR #55, opened 2025-09-19
Refactor truncation module + add `truncerror`

lkdvos:
This PR slightly refactors the truncation implementation, by separating out the interface from the implementation, more similar to the other parts of the library.

Additionally, I ported over `truncerror` (`TruncationError`), based on [TensorKit.jl](https://github.com/QuantumKitHub/TensorKit.jl/blob/ec78a85e72b8e673af6a02ef7fca066d17a0476e/src/tensors/truncation.jl#L10) and [TensorAlgebra.jl](https://github.com/ITensor/TensorAlgebra.jl/blob/581d9a6647eb99e549bbb5c5e4a113443fad57e2/src/MatrixAlgebra.jl#L146)

I tentatively renamed it from `truncerr` to `truncerror`, I'm not sure how useful it is to shorten that name further. Happy to change that back if we don't like this.


src/implementations/truncation.jl
```diff
@@ -243,19 +70,49 @@ function findtruncated(values::AbstractVector, strategy::TruncationIntersection)
...
+    rank = length(values)
+    for i in reverse(I)
+        @inbounds truncerrᵖ += abs(values[i])^strategy.p
+        if truncerrᵖ ≥ ϵᵖ
+            rank = i
+            break
+        end
+    end
+    return I[Base.OneTo(rank)]
+end
```
mtfishman, line 95: Could this be written as:
```julia
function findtruncated(values::AbstractVector, strategy::TruncationError)
  I = sortperm(values; by=abs, rev=true)
  values_sorted = @view values[I]
  I_sorted = findtruncated_sorted(values_sorted, strategy)
  return I[I_sorted]
end
```
Just to avoid code duplication (I may have gotten the logic wrong but hopefully you get the idea).

mtfishman (approved):

src/algorithms.jl
```diff
@@ -131,6 +131,74 @@ If this is not possible, for example when the output size is not known a priori
...
+standard case for SVD) see [`MatrixAlgebraKit.findtruncated_sorted`](@ref).
+""" findtruncated
+
+@doc """
+    MatrixAlgebraKit.findtruncated_sorted(values::AbstractVector, strategy::TruncationStrategy)
+
+Like [`MatrixAlgebraKit.findtruncated`](@ref) but assumes that the values are sorted in reverse order.
+They are assumed to be sorted in a way that is consistent with the truncation strategy,
+which generally means they are sorted by absolute value but some truncation strategies allow
+customizing that. However, note that this assumption is not checked, so passing values that are not sorted
```
Jutho, line 180: I think this is phrased a bit strangely, namely they are sorted in reverse order, and only then you start talking about that they may be sorted in a way that is consistent with the truncation strategy. Do we actually have any nontrivial example of that at the moment?

Jutho (approved):

PR #56, opened 2025-09-21
`TruncationStrategy` types and constructors: consistency in names and implementations

lkdvos:
This is a (breaking) change that attempts to make the truncation structs and their derived (lowercase) constructors a bit more uniform, while also removing some of them.

We previously had the following structs and constructors:
```julia
notrunc()::NoTruncation
truncrank(maxrank; by, rev)::TruncationKeepSorted
TruncationKeepBelow # direct constructor
TruncationKeepAbove # direct constructor
trunctol(val; by)::TruncationKeepFiltered
truncabove(val; by)::TruncationKeepFiltered
truncerror(err)::TruncationError
```
where both `trunctol` and `truncabove` would actually be equivalent to the `TruncationKeepAbove` and `TruncationKeepBelow` with an absolute tolerance, while not making use of the sorted implementations because they were implemented using filters.
Here, I merged `TruncationKeepBelow` and `TruncationKeepAbove` into `TruncationByValue` by adding an additional boolean flag to select between the two, while trying to be more consistent in the other names as well.

The updated scheme consists of:

```julia
notrunc()::NoTruncation
truncrank(maxrank; by, rev)::TruncationByOrder
truncfilter(f)::TruncationByFilter
trunctol(; atol, rtol, p, by, rev)::TruncationByValue
truncerror(; atol, rtol, p)::TruncationByError
```

Additionally, I exported all the lowercase constructors, and updated the docs accordingly.

PS: before actually tagging a 0.4 release, I would additionally like to change the formatter to runic (in a separate PR), to be consistent with the other packages in the organization.

### DIFF
--- a/src/implementations/truncation.jl
+++ b/src/implementations/truncation.jl
@@ -2,7 +2,7 @@
 # ---------
 # Generic implementation: `findtruncated` followed by indexing
 function truncate!(::typeof(svd_trunc!), (U, S, Vᴴ), strategy::TruncationStrategy)
-    ind = findtruncated_sorted(diagview(S), strategy)
+    ind = findtruncated_svd(diagview(S), strategy)
     return U[:, ind], Diagonal(diagview(S)[ind]), Vᴴ[ind, :]
 end
 function truncate!(::typeof(eig_trunc!), (D, V), strategy::TruncationStrategy)
@@ -29,7 +29,7 @@ end
 # findtruncated
 # -------------
 # Generic fallback
-function findtruncated_sorted(values::AbstractVector, strategy::TruncationStrategy)
+function findtruncated_svd(values::AbstractVector, strategy::TruncationStrategy)
     return findtruncated(values, strategy)
 end
 
@@ -79,7 +79,7 @@ function findtruncated(values::AbstractVector, strategy::TruncationByError)
     I′ = _truncerr_impl(values, I; strategy.atol, strategy.rtol, strategy.p)
     return I[I′]
 end
-function findtruncated_sorted(values::AbstractVector, strategy::TruncationByError)
+function findtruncated_svd(values::AbstractVector, strategy::TruncationByError)
     I = eachindex(values)
     I′ = _truncerr_impl(values, I; strategy.atol, strategy.rtol, strategy.p)
     return I[I′]
@@ -107,8 +107,8 @@ function findtruncated(values::AbstractVector, strategy::TruncationIntersection)
     return mapreduce(Base.Fix1(findtruncated, values), _ind_intersect, strategy.components;
                      init=trues(length(values)))
 end
-function findtruncated_sorted(values::AbstractVector, strategy::TruncationIntersection)
-    return mapreduce(Base.Fix1(findtruncated_sorted, values), _ind_intersect,
+function findtruncated_svd(values::AbstractVector, strategy::TruncationIntersection)
+    return mapreduce(Base.Fix1(findtruncated_svd, values), _ind_intersect,
                      strategy.components; init=trues(length(values)))
 end
 
--- a/src/implementations/truncation.jl
+++ b/src/implementations/truncation.jl
@@ -50,10 +50,7 @@ function findtruncated_svd(values::AbstractVector, strategy::TruncationByOrder)
 end
 
 function findtruncated(values::AbstractVector, strategy::TruncationByFilter)
-    # pre-allocate bitvector to enforce the filter function returns a Bool
-    mask = similar(BitArray, eachindex(values))
-    mask .= strategy.filter.(values)
-    return mask
+    return strategy.filter.(values)::AbstractVector{Bool}
 end
 
 function findtruncated(values::AbstractVector, strategy::TruncationByValue)
--- a/src/implementations/truncation.jl
+++ b/src/implementations/truncation.jl
@@ -41,12 +41,9 @@ function findtruncated(values::AbstractVector, strategy::TruncationByOrder)
     return partialsortperm(values, 1:howmany; strategy.by, strategy.rev)
 end
 function findtruncated_svd(values::AbstractVector, strategy::TruncationByOrder)
-    if strategy.by === abs
-        howmany = min(strategy.howmany, length(values))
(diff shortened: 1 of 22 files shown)


kshyatt:
Yes a map + findfirst should work too.

src/interface/truncation.jl
```diff
@@ -42,85 +42,102 @@ Truncation strategy that does nothing, and keeps all the values.
...
 """
-trunctol(val::Real; by=abs) = TruncationKeepFiltered(≥(val) ∘ by)
+truncfilter(f) = TruncationByFilter(f)
 
 """
-    truncabove(val::Real; by=abs)
+    TruncationByValue(atol::Real, rtol::Real, p::Real, by, rev::Bool=true)
 
-Truncation strategy to discard the values that are larger than `val` according to `by`.
+Truncation strategy to keep the values that satisfy `by(val) < max(atol, rtol * norm(values, p)`
```
Jutho, line 85: Is the default to keep smaller values? Is the default value `rev = true`? I would think we want `rev=false` as default, and then for the docstring
```markdown
Truncation strategy to keep the values that satisfy
`by(val) >= max(atol, rtol * norm(values, p)`, or
to discard those values if `rev = true`.
```

src/interface/truncation.jl
```diff
@@ -42,85 +42,102 @@ Truncation strategy that does nothing, and keeps all the values.
...
+function TruncationError(atol::Real, rtol::Real, p::Real=2)
+    return TruncationError(promote(atol, rtol)..., p)
 end
-function TruncationKeepBelow(atol::Real, rtol::Real, p::Real=2, by=abs)
-    return TruncationKeepBelow(promote(atol, rtol)..., p, by)
+
+"""
+    truncerror(; atol::Real=0, rtol::Real=0, p::Real=2)
+
+Create a truncation strategy for truncating such that the error in the factorization
```
Jutho, line 128: "Create a" only appears in this docstring, whereas the others simply start with "Truncation strategy for ..."

src/implementations/truncation.jl
```diff
@@ -28,82 +28,78 @@ end
...
 findtruncated(values::AbstractVector, ::NoTruncation) = Colon()
 
-function findtruncated(values::AbstractVector, strategy::TruncationKeepSorted)
+function findtruncated(values::AbstractVector, strategy::TruncationByOrder)
     howmany = min(strategy.howmany, length(values))
-    return partialsortperm(values, 1:howmany; by=strategy.by, rev=strategy.rev)
+    return partialsortperm(values, 1:howmany; strategy.by, strategy.rev)
 end
-function findtruncated_sorted(values::AbstractVector, strategy::TruncationKeepSorted)
+function findtruncated_sorted(values::AbstractVector, strategy::TruncationByOrder)
```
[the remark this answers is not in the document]
lkdvos: Good catch, I think I actually need to return the complement if rev is false, in the sense that this is less about the sorting of the input, and more about which values are selected. I'm still not entirely sure if there is a better name for this field, very open to suggestions

src/implementations/truncation.jl
```diff
@@ -28,82 +28,97 @@ end
...
-    return 1:howmany
+    return partialsortperm(values, 1:howmany; strategy.by, strategy.rev)
+end
+function findtruncated_sorted(values::AbstractVector, strategy::TruncationByOrder)
+    @assert strategy.by === abs
+    if strategy.by === abs
+        howmany = min(strategy.howmany, length(values))
+        return strategy.rev ? (1:howmany) : ((length(values) - howmany + 1):length(values))
+    else
+        return findtruncated(values, strategy)
```
mtfishman, line 49: Isn't this branch never reached because of the assertion above?

mtfishman:
@lkdvos I lost track, is the plan to replace all uses of `findtruncated_sorted` with `findtruncated_svd`?

src/implementations/truncation.jl
```diff
@@ -28,82 +28,93 @@ end
...
+    return partialsortperm(values, 1:howmany; strategy.by, strategy.rev)
+end
+function findtruncated_svd(values::AbstractVector, strategy::TruncationByOrder)
+    if strategy.by === abs
+        howmany = min(strategy.howmany, length(values))
+        return strategy.rev ? (1:howmany) : ((length(values) - howmany + 1):length(values))
+    else
+        return findtruncated(values, strategy)
+    end
 end
```
mtfishman, line 50:
```suggestion
function findtruncated_svd(values::AbstractVector, strategy::TruncationByOrder)
    strategy.by === abs || return findtruncated(values, strategy)
    howmany = min(strategy.howmany, length(values))
    return strategy.rev ? (1:howmany) : ((length(values) - howmany + 1):length(values))
end
```
Just to match the style of `findtruncated_svd(values::AbstractVector, strategy::TruncationByValue)` (also personally I like that style better).

src/implementations/truncation.jl
```diff
@@ -28,82 +28,89 @@ end
...
-    return 1:howmany
+    return strategy.rev ? (1:howmany) : ((length(values) - howmany + 1):length(values))
 end
 
-# TODO: consider if worth using that values are sorted when filter is `<` or `>`.
-function findtruncated(values::AbstractVector, strategy::TruncationKeepFiltered)
-    ind = findall(strategy.filter, values)
-    return ind
+function findtruncated(values::AbstractVector, strategy::TruncationByFilter)
+    return strategy.filter.(values)::AbstractVector{Bool}
```
Jutho, line 50: Is it really worth returning a bool vector here, instead of just indices? Is `findall(strategy.filter, values)` so much worse? Or is that for GPU compatibility?

Jutho:
I would favour a consistent return type, at least functionally. I know that both boolean vector and list of indices can be used for indexing, but they do behave very different for other purposes. If I want to know the number of values kept, I would need to do `sum` for the boolean vector, and `length` for the list of indices, which is annoying if I don't know what to expect.

I am fine with either choice. I can see some benefits to the boolean vector return type, so if this is easier for @mtfishman that could be a good choice as well. The only consideration that comes to mind is that Julia `find*` methods return lists of indices, so there would be a bit of a mismatch.

Hence, I definitely support and approve of the current state of this PR.

PR #61, opened 2025-10-01
More consistent API for when to have `_full`, `_compact`, ...

lkdvos:
This PR replaces `qr_compact_pullback!` (`lq_compact_pullback!`) with `qr_pullback!` (`lq_pullback!`) since these methods also work for the full case.

Additionally I removed some unused and unexported API that were remnants of old code.
I also fixed the formatting in some of the docstrings.

### DIFF
--- a/src/deprecate.jl
+++ b/src/deprecate.jl
@@ -1,2 +1,0 @@
-Base.@deprecate qr_compact_pullback!(args...) qr_pullback!(args...) false
-Base.@deprecate lq_compact_pullback!(args...) lq_pullback!(args...) false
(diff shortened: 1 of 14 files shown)


src/deprecate.jl
```diff
@@ -0,0 +1,2 @@
+Base.@deprecate qr_compact_pullback!(args...) qr_pullback!(args...) false
+Base.@deprecate lq_compact_pullback!(args...) lq_pullback!(args...) false
```
Jutho, line 2: I assume deprecated methods are exported? Hence, the methods we deprecate are "more visible" than the ones that replace them. How should one deprecate non-exported public functions?

(I am fine with a breaking version after the current bunch of PRs, and not yet caring about deprecating at this stage)

lkdvos: They are, but there is a final argument to the macro `false` to disable that behavior, so these methods aren't exported in this case.

PR #64, opened 2025-10-01
[Feature] Add `ishermitian`, `isantihermitian`, `hermitianpart(!)` and `antihermitianpart(!)`

lkdvos:
This is a small feature which I noticed can be useful in various contexts, and which I have definitely wanted to have in the toolbox myself at some point. It still needs some more tests and making sure we have the correct implementations, but I think this would be a good addition to the package.

### DIFF
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -4,7 +4,7 @@ using LinearAlgebra: LinearAlgebra
 using LinearAlgebra: norm # TODO: eleminate if we use VectorInterface.jl?
 using LinearAlgebra: mul!, rmul!, lmul!, adjoint!, rdiv!, ldiv!
 using LinearAlgebra: sylvester
-using LinearAlgebra: isposdef, ishermitian, issymmetric
+using LinearAlgebra: isposdef, issymmetric
 using LinearAlgebra: Diagonal, diag, diagind, isdiag
 using LinearAlgebra: UpperTriangular, LowerTriangular
 using LinearAlgebra: BlasFloat, BlasReal, BlasComplex, BlasInt
--- a/src/common/matrixproperties.jl
+++ b/src/common/matrixproperties.jl
@@ -57,3 +57,67 @@ See also [`isisometry`](@ref) and [`is_left_isometry`](@ref).
 function is_right_isometry(A::AbstractMatrix; isapprox_kwargs...)
     return isapprox(A * A', LinearAlgebra.I; isapprox_kwargs...)
 end
+
+"""
+    ishermitian(A; isapprox_kwargs...)
+
+Test whether a linear map is Hermitian, i.e. `A = A'`.
+The `isapprox_kwargs` can be used to control the tolerances of the equality.
+"""
+function ishermitian(A; atol::Real = 0, rtol::Real = 0, kwargs...)
+    return (atol == rtol == 0) ? (A == A') : isapprox(A, A'; atol, rtol, kwargs...)
+end
+function ishermitian(A::AbstractMatrix; atol::Real = 0, rtol::Real = 0, norm = LinearAlgebra.norm, kwargs...)
+    Base.require_one_based_indexing(A)
+    m, n = size(A)
+    m == n || return false
+    if atol == rtol == 0
+        for j in 1:n
+            for i in 1:j
+                A[i, j] == adjoint(A[j, i]) || return false
+            end
+        end
+    elseif norm === LinearAlgebra.norm
+        atol = max(atol, rtol * norm(A))
+        for j in 1:n
+            for i in 1:j
+                isapprox(A[i, j], adjoint(A[j, i]); atol, abs, kwargs...)  || return false
+            end
+        end
+    else
+        return isapprox(A, A'; atol, rtol, norm, kwargs...)
+    end
+    return true
+end
+
+"""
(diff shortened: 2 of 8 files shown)


Jutho:
Ok I've added a blocked version of the `(anti)hermitianpart!` function, and made it using our `@functiondef` infrastructure. This means that the output can be a different matrix, but by default the same matrix is selected.

The cache-friendliness of the blocked strategy only seems to show up on my Mac M4 only for really large matrices (several thousands), but that might be because the Apple M processors have very big caches if I remember correctly. The performance seems roughly constant as soon as the blocksize is somewhere in the range 32 - 512, though I did not do any serious benchmarking. Also, this will be hardware dependent, and I assume older hardware would benefit from blocksizes on the smaller end of this range.

Interface-wise, I would still like to propose the following changes:

* Rename `hermitianpart!` to `project_hermitian!` and similar for the antihermitian case.
* Similarly, rename the `hermitian.jl` file to `projections.jl`.
* Also change `uppertriangular!` to `project_uppertriangular!` and similar for lower and add it there.
* Maybe also have `project_isometric!` (not necessarily in this PR), although this in principle covered by `left_orth!` and is "mathematically" somewhat different as it is a manifold projection rather than a subspace projection, and is therefore nonlinear.

Of course this will need dedicated GPU implementations (@kshyatt 😇). And maybe chainrules?

I have not yet looked at the `ishermitian` and friends, but I guess something similar can be done there, although I am not sure this is really necessary.

lkdvos (commented):
Definitely happy with `project_hermitian` etc instead, might even be nicer to not name-clash with `LinearAlgebra` too.

I think it might even pay off to add the optimization to the `ishermitian` check, since this is actually called in all the `check_input(eigh_*, ...)` methods.

Jutho:
The difference with `LinearAlgebra.ishermitian` is that our implementation accepts a tolerance. We currently have `ishermitian` and `isisometry` and `is_left_isometry` and so forth.

One proposal could be to use underscores always: `is_hermitian`, `is_antihermitian`, `is_isometry`, … (even though I am not myself a big fan of underscores). That seems quite orthogonal to Base, that has a whole bunch of `is...` methods without any underscore anywhere.

Alternatively, we could use `isapprox` in the name: `isapprox_hermitian`, `isapprox_isometry`, …
However, I think we will often be using the `atol = rtol = 0` version. 

More ideas are welcome.

kshyatt:
For the approx versions, we can probably then adapt the one in GPUArrays and revisit it if it's a performance problem, then

src/common/matrixproperties.jl
```diff
@@ -54,6 +61,86 @@ The `isapprox_kwargs` can be used to control the tolerances of the equality.
...
+    if iszero(atol) && iszero(rtol)
+        return isantihermitian_exact(A; kwargs...)
+    else
+        return 2 * norm(project_hermitian(A; kwargs...)) ≤ max(atol, rtol * norm(A))
+    end
+end
+function isantihermitian_exact(A)
+    return A == -A'
+end
+function isantihermitian_exact(A::StridedMatrix; kwargs...)
```
lkdvos, line 107: I think this is probably the right default, but we have to keep in mind that many strided arrays aren't subtyping `StridedMatrix`. (e.g. `Strided.StridedView <: StridedArray == false`), and several wrappers have a similar issue. 
Maybe we should rename the `_ishermitian_exact` functions to `strided_ishermitian_exact`, such that it is at least easy to patch that with package extensions?

lkdvos:
I'm happy with the current names. I'm also happy with the default tolerances right now, since I would like to have `is*` be exact, with the option of having a tolerance, but whenever that is not feasible because of floating points we additionally account for that. I'm assuming that this is precisely why LinearAlgebra doesn't support `isunitary` etc, but I think that as long as we are clear in our docstrings we should just take whichever is more practical.

src/common/matrixproperties.jl
Jutho: Can other types not simply overload `ishermitian_exact` and `isantihermitian_exact`? 

The `_ishermitian_exact` function is really an implementation detail (and is in fact not type restricted to strided matrices)

src/MatrixAlgebraKit.jl
```diff
@@ -3,14 +3,16 @@ module MatrixAlgebraKit
 using LinearAlgebra: LinearAlgebra
 using LinearAlgebra: norm # TODO: eleminate if we use VectorInterface.jl?
 using LinearAlgebra: mul!, rmul!, lmul!, adjoint!, rdiv!, ldiv!
-using LinearAlgebra: sylvester
-using LinearAlgebra: isposdef, ishermitian, issymmetric
+using LinearAlgebra: sylvester, lu!
+using LinearAlgebra: isposdef, issymmetric, opnorm
```
Jutho, line 7: This can go again, it resulted from things I was trying with `PolarNewton`. 
```suggestion
using LinearAlgebra: sylvester
using LinearAlgebra: isposdef
```

Jutho:
Ok, for me this is also ready I think. I know I mentioned to also add `project_uppertriangular!` in this style, but for now I would like to keep it simple like this.

src/common/matrixproperties.jl
lkdvos: What I mean is that the other types should overload `ishermitian_exact`, by making it call `strided_ishermitian_exact`, but it doesn't matter too much?

PR #67, opened 2025-10-06
[Features] `project_isometric` and `PolarNewton`

Jutho:
PolarNewton implementation. This is branched off and requires #64 .


Jutho:
Some random benchmark for `size(A) == (2000, 1000)`.

```julia-repl
julia> @benchmark left_polar!($Ac, ($W, $P)) setup=(copy!($Ac, $A))
BenchmarkTools.Trial: 24 samples with 1 evaluation per sample.
 Range (min … max):  207.428 ms … 232.445 ms  ┊ GC (min … max): 0.12% … 10.51%
 Time  (median):     208.516 ms               ┊ GC (median):    0.11%
 Time  (mean ± σ):   211.400 ms ±   8.050 ms  ┊ GC (mean ± σ):  1.53% ±  3.52%

   █                                                             
  ▃█▆█▅▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▆ ▁
  207 ms           Histogram: frequency by time          232 ms <

 Memory estimate: 53.57 MiB, allocs estimate: 130.

julia> @benchmark left_polar!($Ac, ($W, $P), PolarNewton()) setup=(copy!($Ac, $A))
BenchmarkTools.Trial: 45 samples with 1 evaluation per sample.
 Range (min … max):  106.843 ms … 133.498 ms  ┊ GC (min … max): 0.00% … 0.33%
 Time  (median):     112.492 ms               ┊ GC (median):    0.41%
 Time  (mean ± σ):   112.160 ms ±   3.850 ms  ┊ GC (mean ± σ):  0.39% ± 0.11%

              ▁ ▁█                                               
  ▃▁▃█▃▆▁▅▅█▃▁█▅██▅▃▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃ ▁
  107 ms           Histogram: frequency by time          133 ms <

 Memory estimate: 38.79 MiB, allocs estimate: 121.
```

Jutho:
Ok I managed to screw up this PR with git rebase on the current main.

Jutho:
Ok, I managed to recover. So this now includes the `PolarNewton` implementation, support for not computing the positive definite factor similar to in the QR case, a `project_isometric!` function that uses the latter functionality, and some more tests.

If the tests pass, this is ready for a review.

src/MatrixAlgebraKit.jl
```diff
@@ -3,16 +3,16 @@ module MatrixAlgebraKit
 using LinearAlgebra: LinearAlgebra
 using LinearAlgebra: norm # TODO: eleminate if we use VectorInterface.jl?
 using LinearAlgebra: mul!, rmul!, lmul!, adjoint!, rdiv!, ldiv!
-using LinearAlgebra: sylvester
-using LinearAlgebra: isposdef, issymmetric
+using LinearAlgebra: sylvester, lu!
+using LinearAlgebra: isposdef, issymmetric, opnorm
```
lkdvos, line 7:
```suggestion
using LinearAlgebra: isposdef, issymmetric
```
I don't think this is being used (anymore)?

src/interface/polar.jl
```diff
@@ -49,6 +49,15 @@ struct PolarViaSVD{SVDAlg} <: AbstractAlgorithm
     svdalg::SVDAlg
 end
 
+"""
```
lkdvos, line 52: I think the other algorithms are all defined in the `src/interface/decompositions.jl` file. How do you feel about also moving these (including the `PolarViaSVD` above) there?

src/MatrixAlgebraKit.jl
```diff
@@ -34,6 +34,7 @@ export LAPACK_HouseholderQR, LAPACK_HouseholderLQ, LAPACK_Simple, LAPACK_Expert,
     LAPACK_QRIteration, LAPACK_Bisection, LAPACK_MultipleRelativelyRobustRepresentations,
     LAPACK_DivideAndConquer, LAPACK_Jacobi
 export LQViaTransposedQR
+export PolarViaSVD, PolarNewton
```
lkdvos, line 37: Definitely agree that we should export this, but before we explicitly mark this as public, I noticed that we have:
`PolarViaSVD(svdalg)` and `LQViaTransposedQR(qr_alg)` (note the underscore in the second).
I'm fine with either, but we should probably be consistent, and this might be the correct time to do that?

src/MatrixAlgebraKit.jl
```diff
@@ -3,16 +3,16 @@ module MatrixAlgebraKit
...
 using LinearAlgebra: Diagonal, diag, diagind, isdiag
 using LinearAlgebra: UpperTriangular, LowerTriangular
 using LinearAlgebra: BlasFloat, BlasReal, BlasComplex, BlasInt
 
 export isisometry, isunitary, ishermitian, isantihermitian
 
-export project_hermitian, project_antihermitian
-export project_hermitian!, project_antihermitian!
+export project_hermitian, project_antihermitian, project_isometric
+export project_hermitian!, project_antihermitian!, project_isometric!
```
lkdvos, line 15: I'm slightly unhappy with `project_isometric` vs `isisometry`. Is it at all reasonable to have `project_isometry`, or does that not really make sense? Happy to leave as is too, just noticing this here.

lkdvos (commented):
Overall looks great to me. Left a couple small comments, but otherwise happy to merge.

src/MatrixAlgebraKit.jl
Jutho: That is a good catch. But I would prefer `isisometric`, as all other checks are also using the adjective form, e.g. `ishermitian`, ...

I know this constitutes yet another breaking change 😄 .

I guess unitary is the odd duck, as that is both an adjective and a noun.

src/MatrixAlgebraKit.jl
Jutho: Either way is fine with me.

src/MatrixAlgebraKit.jl
Jutho: No, that's another remnant of my github rebase struggles.

lkdvos:
I addressed all the issues except for the `isisometric` one, for which I opened an issue: https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/69
I would say to merge this like it is now, so non-breaking, and then add the breaking change when we decide to go to 0.6.0?

lkdvos (approved):

Jutho:
Ok, I'll merge when the tests turn green.

Issue #66, opened 2025-10-05
`left_orth` and `right_orth` interface improvements


lkdvos:
I think the current interface of `left_orth` and `right_orth` (and probably also the `null` equivalents) could use some improvements, mainly for the following kinds of issues:

- Since the output is allocated before the algorithm/decomposition type is selected, it is possible to run into issues when different decompositions require different output types. This already shows up with the `DiagonalAlgorithm` implementations, where a polar decomposition would in general not have a diagonal output type because of the current implementation, while a qr does have that.
- Preselecting an algorithm to pass on to `left_orth` additionally requires you to select a `kind`, and then also make sure to fill out the correct `alg_kind = ...` keyword, which seems rather unpractical, as that in principle requires you to keep track of `kind, alg_qr, alg_polar, alg_svd` and select values for that, even though only one of them might be used.

I'll try and come up with an improved interface to address these over the next couple days.

PR #77, opened 2025-10-12
Refactor orthogonalization and nullspace interface

lkdvos:
This PR aims to address #66:

The starting point is to have two main modes of operation, similar to what we had before but with some small changes:

1. Being able to easily vary the `kind` of factorization, in which case you provide a set of keywords for either of the possible kinds that can be chosen, and the correct ones are taken into account while the other ones are being ignored
2. Being able to specify an algorithm, where the `kind` of factorization is determined automatically.

Going from there, I made a number of changes to help achieve this goal.

- I promoted `left_orth` and friends to full `@functiondef` implementations. This means that they all follow the same procedure as the other functions, and importantly this cleanly separates out the algorithm selection procedure from the implementation.
- This ensures that the decomposition type is selected *before* the output is initialized, leaving the option to specialize what kinds of output are generated
- I added/kept the helper functions `left_orth_qr` etc, with some boilerplate to provide good default implementations, but leaving the option to specialize here.

To achieve this while keeping type stability intact as much as possible, I had to slightly rework the interface. In particular, as the `@functiondef` algorithm selection procedure depends/specializes mostly on the `alg` keyword, I had a hard time making things type stable by only using the other keyword arguments. Therefore, I adopted `alg` as the driving selection mechanism:

- `alg::Symbol` replaces the previous `kind::Symbol`, and in this case the selection procedure is based on the `alg_qr`, `alg_svd` etc keyword arguments.
- `alg::AbstractAlgorithm` uses a "trait-like" system to determine the factorization kind
- the default `alg::Nothing=nothing` selects either a QR or SVD based decomposition based on the presence of `trunc`.

This necessarily removes the ability to select an algorithm type through a symbol, as is possible for example for the `qr_compact(A; alg=:LAPACK_HouseholderQR)`. I don't think this is something I can make work in a type-stable manner (I definitely tried a bunch of things, but short of inlining everything and using generated functions, I don't see how I can convince the compiler to constant-propagate that symbol, which would then still not really be type-stable if that symbol is meant to be varied)

Some other relevant changes:

- I altered the `TruncationIntersection` implementation to use a recursive approach, which seems to improve type-stability
- I reworked some of the tests, by separating out the `LinearMap` implementation to a separate file, and removing some of the checks on the outputs of `left_orth` etc aliasing the inputs. As our interface explicitly specifies this is not behavior that should be depended on, I feel like it is warranted that our tests should not test for this?
- The default `left_null_svd` implementation now ignores the specified output, as this cannot reasonably be used since the size varies. This doesn't preclude us from adding this back in for specialized cases, and the infrastructure is still there, albeit bypassed through `N = nothing`.

I think overall I've removed some boilerplate code, and the added boilerplate should be beneficial for downstream packages, and low-maintenance on our side here.
I'm obviously curious about any comments and suggestions for possible improvements.
Some remarks that I've encountered which might need addressing:

- The truncation strategies for the null spaces are a bit funny, in the sense that we might want to consider them as being the inverted case of what they are for the truncated decompositions (by explicitly returning the complement of the `ind` computed currently). One thing we may therefore consider is adding a `findtruncated_null` on top of the `findtruncated` and `findtruncated_svd`, which has precisely this interpretation. I'm not sure if this would overall make a big change, but it might make it easier to reason about.
- There are some gymnastics required to make the two use-cases fit together in a single function, and I can't really see a use-case for wanting both of these uses at the same time. Do we want to just split this into two separate functions? `left_orth1(A; kind, trunc, alg_svd, ...)` and `left_orth2(A; alg, trunc, kwargs...)` just requires us to come up with 2 names, but is otherwise a bit more straightforward.
- Do we want to rework the `alg::Nothing` mode to select an `alg::Symbol` to match the presence of `trunc`, and then dispatch to the `alg::Symbol` mode? Currently these are slightly distinct, as `alg::Nothing` uses all `kwargs` for algorithm selection, while `alg::Symbol` uses the `alg_qr` etc.

### DIFF
--- a/test/orthnull.jl
+++ b/test/orthnull.jl
@@ -7,9 +7,6 @@ using LinearAlgebra: LinearAlgebra, I
 # testing non-AbstractArray codepaths:
 include("linearmap.jl")
 
-# testing non-AbstractArray codepaths:
-include("linearmap.jl")
-
 eltypes = (Float32, Float64, ComplexF32, ComplexF64)
 @testset "left_orth and left_null for T = $T" for T in eltypes
     rng = StableRNG(123)


src/implementations/truncation.jl
```diff
@@ -26,6 +26,17 @@ function truncate(::typeof(right_null!), (S, Vᴴ), strategy::TruncationStrategy
     return Vᴴ[ind, :], ind
 end
 
+# special case `NoTruncation` for null: should keep exact zeros due to rectangularity
+function truncate(::typeof(left_null!), (U, S), strategy::NoTruncation)
+    ind = (1:max(0, size(S, 1) - size(S, 2))) .+ length(diagview(S))
```
Jutho: `length(diagview(S)) == size(S, 2)` ?
`ind = (size(S, 2) + 1):size(S,1)` ?

PR #84, opened 2025-10-22
 WIP Enzyme and Mooncake rules

kshyatt:
This adds *CPU* based reverse rules for all the factorizations (NB below) in MatrixAlgebraKit, using Mooncake or Enzyme. I tried my best to use existing pullbacks.

Some notes:
- The Enzyme tests do not directly test the `eigh_trunc!`, `eig_trunc!`, or `svd_trunc!` rules due to https://github.com/EnzymeAD/Enzyme.jl/issues/2677 and https://github.com/EnzymeAD/Enzyme.jl/issues/2678. This is because of their mixed return of mutable and immutable objects. I'll work on creating a "bypass" signature for each that passes in a reference to epsilon that should allow us to test directly.
- The Enzyme complex number tests are extremely slow due to https://github.com/EnzymeAD/Enzyme.jl/issues/2631 so are disabled for CI
- None of the rule-testing methods for Mooncake or Enzyme work with GPU, it seems, so I've disabled those tests for now.
- I added some "pull-forwards" for some factorizations but they're rather annoying to test directly. Also a TODO I will work on while this is commented upon :)


kshyatt:
@Jutho would it make more sense to remove the forward mode code entirely from this PR and shunt it off into a new one?

kshyatt:
I do think moving the fwd mode stuff out might make this substantially easier to review, actually. We can always test the forward rules once the (working) reverse ones are in place.

lkdvos:
I think I agree, it might be nice to split this up into several separated parts to get things moving easier. I would propose the following:

- Since the GPU stuff isn't used yet, it might be nice to not have that code in the PR, just to avoid clutter
- Can we consider Enzyme and Mooncake separately? Am I correct thinking that these don't actually interact, so we could review/test/implement them on their own?
- I think there are a bunch of small improvements to the error and warning messages, I think these are all good and we can simply merge them already
- It might actually be reasonable to add the `remove_f_gaugedependence` functions into the main package, not sure how yo feel about this? We kind of need that in the downstream package tests anyways, so it might be useful to just include them directly.

The rest definitely looks great, do you think it could be reasonable to schedule a meeting to just go over the general Mooncake and/or Enzyme approach as a whole? Obviously I don't want to put this on you if you aren't up for it, and I can try and read through them myself, but if you would be okay it might be nice to get the explanation from someone who already has some experience with it :)

src/pullfwds/qr.jl
```diff
@@ -0,0 +1,69 @@
...
+
+    # fwd rule for Q1 and R11 -- for a non-rank redeficient QR, this is all we need
+    invR11  = inv(R11)
+    tmp     = Q1' * dA1 * invR11
+    Rtmp    = tmp + tmp'
+    diagview(Rtmp) ./= 2
+    ltRtmp  = view(Rtmp, MatrixAlgebraKit.lowertriangularind(Rtmp))
+    ltRtmp .= zero(eltype(Rtmp))
+    dR11   .= Rtmp * R11
+    dQ1    .= dA1 * invR11 - Q1 * dR11 * invR11
```
Jutho: `dR11 * invR11` is `Rtmp`, so that can be simplified.

src/pullfwds/qr.jl
```diff
@@ -0,0 +1,69 @@
...
+    R11 = view(R, 1:m1, 1:n1)
+    R12 = view(R, 1:m1, n1+1:n)
+
+    dA1 = view(dA, 1:m, 1:n1)
+    dA2 = view(dA, 1:m, (n1 + 1):n)
+
+    dQ, dR = dQR
+    dQ1    = view(dQ, 1:m, 1:m1)
+    dQ2    = view(dQ, 1:m, m1+1:m2+m1)
+    dQ3    = m1+m2+1 < size(dQ, 2) ? view(dQ, 1:m, m1+m2+1:size(dQ,2)) : similar(dQ, eltype(dQ), (0, 0))
```
Jutho: Do you mean `m1+m2 < size(dQ, 2)` or `m1+m2+1 <= size(dQ, 2)`?

Also, does just using `view(dQ, 1:m, m1+m2+1:size(dQ,2))` always not just work? In the case that `m1+m2 == size(dQ, 2)`, we automatically get a view with `size(dQ3, 2)=0`, no ?

PR #85, opened 2025-10-22
Mooncake reverse rules

kshyatt:


ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,225 @@
...
+        (lq_compact!, lq_pullback!, :dlq_adjoint),
+        (eig_full!, eig_pullback!, :deig_adjoint),
+        (eigh_full!, eigh_pullback!, :deigh_adjoint),
+        (left_polar!, left_polar_pullback!, :dleft_polar_adjoint),
+        (right_polar!, right_polar_pullback!, :dright_polar_adjoint),
+    )
+
+    @eval begin
+        @is_primitive Mooncake.DefaultCtx Mooncake.ReverseMode Tuple{typeof($f), Any, Tuple{<:Any, <:Any}, MatrixAlgebraKit.AbstractAlgorithm}
+        function Mooncake.rrule!!(::CoDual{typeof($f)}, A_dA::CoDual, args_dargs::CoDual, alg_dalg::CoDual{<:MatrixAlgebraKit.AbstractAlgorithm}; kwargs...)
```
lkdvos:
```suggestion
        function Mooncake.rrule!!(::CoDual{typeof($f)}, A_dA::CoDual, args_dargs::CoDual, alg_dalg::CoDual{<:MatrixAlgebraKit.AbstractAlgorithm})
```

I don't think we are allowing any additional keyword arguments on top of what is already in the `AbstractAlgorithm`, so this probably does not need to be here

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,225 @@
...
+                    dAtmp_
+                else
+                    map(A_ -> Mooncake.build_tangent(typeof(A_), real(A_), imag(A_)), dAtmp_)
+                end
+                Mooncake.increment!!(Mooncake.tangent(A_dA), dAtmp)
+                arg1 .= arg1c
+                arg2 .= arg2c
+                A .= Ac
+                darg1 .= zero(darg1)
+                darg2 .= zero(darg2)
```
lkdvos:
```suggestion
                MatrixAlgebraKit.zero!(darg1)
                MatrixAlgebraKit.zero!(darg2)
```
Is the other implementation not making copies? I might be wrong about that though.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,225 @@
...
+    function deig_vals_adjoint(::Mooncake.NoRData)
+        dA .= zero(eltype(dA))
+        A .= Ac
+        PΔV = V' \ Diagonal(dD)
+        if eltype(dA) <: Real
+            ΔAc = PΔV * V'
+            dA .+= real.(ΔAc)
+        else
+            mul!(dA, PΔV, V', 1, 0)
+        end
```
lkdvos: I think this logic can be achieved by passing an appropriate zerotangent for `V` to the `eig_pullback` function, see https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/c52614bffe120e3e63fbbd222230cdc695f476ea/src/pullbacks/eig.jl#L38

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,225 @@
...
+    # compute primal
+    D_ = Mooncake.primal(D_dD)
+    dD_ = Mooncake.tangent(D_dD)
+    A_ = Mooncake.primal(A_dA)
+    dA_ = Mooncake.tangent(A_dA)
+    A, dA = arrayify(A_, dA_)
+    D, dD = arrayify(D_, dD_)
+    DV = eigh_full(A, Mooncake.primal(alg_dalg); kwargs...)
+    function deigh_vals_adjoint(::Mooncake.NoRData)
+        mul!(dA, DV[2] * Diagonal(real(dD)), DV[2]', 1, 0)
```
lkdvos: same comment as above

src/pullbacks/eig.jl
```diff
@@ -48,7 +48,8 @@ function eig_pullback!(
         Δgauge < gauge_atol ||
             @warn "`eig` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
 
-        VᴴΔV .*= conj.(inv_safe.(transpose(D) .- D, degeneracy_atol))
+        VᴴΔV ./= conj.(transpose(D) .- D)
+        diagview(VᴴΔV) .= zero(eltype(VᴴΔV))
```
lkdvos: leftover that can be removed?

test/ad_utils.jl
```diff
@@ -0,0 +1,32 @@
...
+    gaugepart[abs.(transpose(diagview(D)) .- diagview(D)) .>= degeneracy_atol] .= 0
+    mul!(ΔV, V / (V' * V), gaugepart, -1, 1)
+    return ΔV
+end
+function remove_eighgauge_dependence!(
+        ΔV, D, V;
+        degeneracy_atol = MatrixAlgebraKit.default_pullback_gaugetol(S)
+    )
+    gaugepart = V' * ΔV
+    gaugepart = (gaugepart - gaugepart') / 2
```
lkdvos:
```suggestion
    gaugepart = project_antihermitian!(gaugepart)
```

kshyatt:
> One thing that I am still struggling with is understanding why the adjoint functions return NoRData() for all of the input arguments. Am I correct in assuming that this only works for mutable data? Should we maybe reconsider the strictness of our rules in that case?

Yes, afaik it's set up this way because all the reverse-data you need should be "packaged in" to the tangents of the `CoDual` for arrays or tuples of arrays -- you'd return explicit `RData` if you have scalar arguments

test/mooncake.jl
```diff
@@ -0,0 +1,561 @@
...
+    return (Mooncake.increment!!(x[1], y[1]), Mooncake.increment!!(x[2], y[2]), Mooncake.increment!!(x[3], y[3]))
+end
+function Mooncake.increment!!(x::Tuple{Mooncake.Tangent{@NamedTuple{diag::Vector{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}}, Matrix{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}, y::Tuple{Mooncake.Tangent{@NamedTuple{diag::Vector{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}}, Matrix{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}) where {T <: Real}
+    return (Mooncake.increment!!(x[1], y[1]), Mooncake.increment!!(x[2], y[2]))
+end
+function Mooncake.increment!!(x::Tuple{Mooncake.Tangent{@NamedTuple{diag::Vector{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}}, Matrix{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}, y::Tuple{Mooncake.Tangent{@NamedTuple{diag::Vector{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}}}, Matrix{Mooncake.Tangent{@NamedTuple{re::T, im::T}}}, Vector{T}}) where {T <: Real}
+    return (Mooncake.increment!!(x[1], y[1]), Mooncake.increment!!(x[2], y[2]))
+end
```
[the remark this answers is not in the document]
kshyatt: I need to see if this has been fixed recently but yes this is needed for some of the derivative tracking. If it doesn't work with these removed we can contribute them upstream.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
kshyatt: Or you'd have to check that each element of the tangent is zero valued which could get expensive

src/pullbacks/eig.jl
lkdvos: In that case we should probably update the tolerances instead?

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
lkdvos: That's definitely a good question. What I really meant is that we are not allowing any extra kwargs to the primal calls, which I think has as a consequence that there shouldn't be kwargs here to pass on? I agree that we will at some point have to figure out how to control these settings, but I'm open to suggestions there because I definitely haven't figured that out.

src/pullbacks/eig.jl
kshyatt: Let me see if I can find a good updated one that doesn't blow everything up

kshyatt: OK, seems to be ok... Maybe again this was fixed inadvertently

test/mooncake.jl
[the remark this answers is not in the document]
kshyatt: It has apparently! Hooray :)

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,225 @@
...
+            output = $f(A, args, Mooncake.primal(alg_dalg); kwargs...)
+            function $adj(::Mooncake.NoRData)
+                dAtmp_ = zero(dA)
+                dAtmp_ .= $pb(dAtmp_, A, (arg1, arg2), (darg1, darg2); kwargs...)
+                dAtmp = if eltype(dA) <: Real
+                    dAtmp_
+                else
+                    map(A_ -> Mooncake.build_tangent(typeof(A_), real(A_), imag(A_)), dAtmp_)
+                end
+                Mooncake.increment!!(Mooncake.tangent(A_dA), dAtmp)
```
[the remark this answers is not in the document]
kshyatt: OK, the tests are now passing using the `arrayify`-ed objects (which are `reinterpret`'d correctly). I'll mark this resolved.

test/ad_utils.jl
lkdvos: Given that I have to copy this implementation in TensorKit, and failed horrendously:
https://github.com/QuantumKitHub/TensorKit.jl/blob/bbeb8e510727fd2ce77c579ae9833f5056c85cc8/test/autodiff/ad.jl#L105-L120
I think it would be nice to have this as part of the main package, and if not, to turn this into a "test module" that can be included from within downstream packages (probably mostly by adding some comments to this file to indicate that its location and contents are considered "public" and cannot be moved/removed without breaking changes)

kshyatt:
FWIW for Enzyme, there is no requirement to restore the state of he variables on the reverse pass which is the other reason I did things this way -- to be able to compare the two "like to like".

src/pullbacks/eig.jl
```diff
@@ -46,7 +46,7 @@ function eig_pullback!(
         Δgauge < gauge_atol ||
             @warn "`eig` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
 
-        VᴴΔV .*= conj.(inv_safe.(transpose(D) .- D, degeneracy_atol))
+        VᴴΔV .*= conj.(inv_safe.(transpose(D) .- D, tol))
```
lkdvos:
```suggestion
        VᴴΔV .*= conj.(inv_safe.(transpose(D) .- D, degeneracy_atol))
```

lkdvos (commented):
Okay, let me summarize what I think for this PR specifically, also keep in mind that this is my opinion and you are allowed to just overrule that:

- [x] `kwargs...` in the `rrule!!`
I don't actually have a great solution for the fact that we don't really support keyword arguments on top of the factorizations.
I don't think the current way of just adding these as kwargs to the `rrule!!` works (although it doesn't hurt either), since the primal call can't actually handle any keyword arguments (e.g. https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/c99084bffcbb2a4c84390ae78e92ada4ab68d97f/src/implementations/eig.jl#L81).
The only precedences for this we currently have are either KrylovKit, where we ended up adding a `alg_rrule` keyword argument to all of the primal computations that is simply ignored, which could be an empty `NamedTuple` in our case.
The other is PEPSKit, where I designed a `hook_pullback` function that collaborates with the ChainRules.jl system to insert keyword arguments into a function that are passed on to the pullback: https://github.com/QuantumKitHub/PEPSKit.jl/blob/master/src/utility/hook_pullback.jl
Importantly, we don't need to solve that for this PR, since we are more or less everywhere just using the defaults anyways, so we can tackle that later. I would therefore maybe suggest to remove the keywords for now, since adding that would be non-breaking so that's easier to do in a follow up. (I'm fine leaving them as is too, it shouldn't matter since these shouldnt be able to ever be generated anyways)

- [ ] Did we decide what to do with the `@is_primitive` directions?
I remember talking about them being more or less specific, but can't remember what we decided on. What is worrying me slightly is that we are clearly assuming everywhere that the inputs and outputs are mutable, which I guess doesn't always have to be the case. In the interest of moving things forwards, I guess the easiest solution here is to just restrict these definitions more aggressively, claiming that the AD engine should then figure out how to handle other cases and start adding cases back in, rather than the other direction?

- [x] There's a little bit of leftover commented out code in the tests, which could do with either removing or perhaps some explanatory comments about why they are there/why they are commented out

- [ ] Overall it might be slightly nicer to have some additional comments to explain what is going on, as it really takes a bit of time to understand all of this. I'm not saying this has to be done neither by you nor now, just a general remark that this could actually benefit the overall progress here a bit.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,337 @@
...
+        function Mooncake.rrule!!(::CoDual{<:typeof($f!)}, A_dA::CoDual, D_dD::CoDual, alg_dalg::CoDual; kwargs...)
+            # compute primal
+            D_ = Mooncake.primal(D_dD)
+            dD_ = Mooncake.tangent(D_dD)
+            A_ = Mooncake.primal(A_dA)
+            dA_ = Mooncake.tangent(A_dA)
+            A, dA = arrayify(A_, dA_)
+            D, dD = arrayify(D_, dD_)
+            # update primal
+            DV = $full_f(A, Mooncake.primal(alg_dalg); kwargs...)
```
lkdvos: I think this one can be the in-place version of the full function?

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,337 @@
...
+                dU_, dS_, dVᴴ_ = Mooncake.tangent(USVᴴ_codual)
+                U, dU = Mooncake.arrayify(U, dU_)
+                S, dS = Mooncake.arrayify(S, dS_)
+                Vᴴ, dVᴴ = Mooncake.arrayify(Vᴴ, dVᴴ_)
+                if ($f == svd_compact)
+                    svd_pullback!(dA, A, (U, S, Vᴴ), (dU, dS, dVᴴ))
+                else # full
+                    minmn = min(size(A)...)
+                    vU = view(U, :, 1:minmn)
+                    vS = Diagonal(diagview(S)[1:minmn])
```
lkdvos:
```suggestion
                    vS = Diagonal(view(diagview(S), 1:minmn))
```
I think?

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,337 @@
...
+                    vVᴴ = view(Vᴴ, 1:minmn, :)
+                    vdU = view(dU, :, 1:minmn)
+                    vdS = Diagonal(diagview(dS)[1:minmn])
```
lkdvos:
```suggestion
                    vdS = Diagonal(view(diagview(dS, 1:minmn))
```

test/runtests.jl
lkdvos: some accidentally commented out tests here

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,337 @@
...
+                arg1, arg2 = Mooncake.primal(output_codual)
+                darg1_, darg2_ = Mooncake.tangent(output_codual)
+                arg1, darg1 = Mooncake.arrayify(arg1, darg1_)
+                arg2, darg2 = Mooncake.arrayify(arg2, darg2_)
+                $pb(dA, A, (arg1, arg2), (darg1, darg2); kwargs...)
+                MatrixAlgebraKit.zero!(darg1)
+                MatrixAlgebraKit.zero!(darg2)
+                return Mooncake.NoRData(), Mooncake.NoRData(), Mooncake.NoRData()
+            end
+            MatrixAlgebraKit.zero!(dA)
```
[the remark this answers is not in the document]
kshyatt: I think in fact you're right and on the fwd pass this is effectively a no-op and should not be here.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,352 @@
...
+end
+
+@is_primitive Mooncake.DefaultCtx Mooncake.ReverseMode Tuple{typeof(MatrixAlgebraKit.svd_vals), Any, MatrixAlgebraKit.AbstractAlgorithm}
+function Mooncake.rrule!!(::CoDual{typeof(MatrixAlgebraKit.svd_vals)}, A_dA::CoDual, alg_dalg::CoDual)
+    # compute primal
+    A = Mooncake.primal(A_dA)
+    dA_ = Mooncake.tangent(A_dA)
+    A, dA = arrayify(A, dA_)
+    S = svd_vals(A, Mooncake.primal(alg_dalg))
+    U, _, Vᴴ = svd_compact(A, Mooncake.primal(alg_dalg))
```
Jutho: Calling `svd_` twice could become a real overhead. Any reason not to simply use the following:

```suggestion
    U, S_, Vᴴ = svd_compact(A, Mooncake.primal(alg_dalg))
    S = diagview(S_)
```

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,352 @@
...
+    alg = Mooncake.primal(alg_dalg)
+    output = svd_trunc(A, alg)
+    # fdata call here is necessary to convert complicated Tangent type (e.g. of a Diagonal
+    # of ComplexF32) into the correct **forwards** data type (since we are now in the forward
+    # pass). For many types this is done automatically when the forward step returns, but
+    # not for nested structs with various fields (like Diagonal{Complex})
+    output_codual = Mooncake.CoDual(output, Mooncake.fdata(Mooncake.zero_tangent(output)))
+    function dsvd_trunc_adjoint(dy::Tuple{Mooncake.NoRData, Mooncake.NoRData, Mooncake.NoRData, T}) where {T <: Real}
+        Utrunc, Strunc, Vᴴtrunc, ϵ = Mooncake.primal(output_codual)
+        dUtrunc_, dStrunc_, dVᴴtrunc_, dϵ = Mooncake.tangent(output_codual)
```
Jutho: We currently do not support a nontrivial cotangent for the truncation error, but this could be done (and is probably not even that hard). However, how should I within Mooncake world now read of teh cotangent for `ϵ`. Is it completely in `dϵ`, or also in the fourth component of `dy`?

test/mooncake.jl
```diff
@@ -0,0 +1,547 @@
...
+    copy_pb!!(rdata)
+
+    # not every f! has a handwritten rrule!!
+    inplace_sig = isnothing(alg) ? Tuple{typeof(f!), typeof(A), typeof(args)} : Tuple{typeof(f!), typeof(A), typeof(args), typeof(alg)}
+
+    has_handwritten_rule = hasmethod(Mooncake.rrule!!, inplace_sig)
+    if has_handwritten_rule
+        inplace_out, inplace_pb!! = isnothing(alg) ? Mooncake.rrule!!(Mooncake.CoDual(f!, Mooncake.NoFData()), Mooncake.CoDual(Ac, dA_inplace), Mooncake.CoDual(args, dargs_inplace)) : Mooncake.rrule!!(Mooncake.CoDual(f!, Mooncake.NoFData()), Mooncake.CoDual(Ac, dA_inplace), Mooncake.CoDual(args, dargs_inplace), Mooncake.CoDual(alg, Mooncake.NoFData()))
+    else
+        inplace_sig = isnothing(alg) ? Tuple{typeof(f!), typeof(A), typeof(args)} : Tuple{typeof(f!), typeof(A), typeof(args), typeof(alg)}
```
Jutho: This looks the same as the `inplace_sig` defined on line 62

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -0,0 +1,352 @@
...
+                    svd_pullback!(dA, A, (U, S, Vᴴ), (dU, dS, dVᴴ))
+                else # full
+                    minmn = min(size(A)...)
+                    vU = view(U, :, 1:minmn)
+                    vS = Diagonal(view(diagview(S), 1:minmn))
+                    vVᴴ = view(Vᴴ, 1:minmn, :)
+                    vdU = view(dU, :, 1:minmn)
+                    vdS = Diagonal(view(diagview(dS), 1:minmn))
+                    vdVᴴ = view(dVᴴ, 1:minmn, :)
+                    svd_pullback!(dA, A, (vU, vS, vVᴴ), (vdU, vdS, vdVᴴ))
```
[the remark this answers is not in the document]
kshyatt: created a rough issue that links here, will fill in details when I find time

src/common/view.jl
```diff
@@ -1,6 +1,7 @@
 # diagind: provided by LinearAlgebra.jl
 diagview(D::Diagonal) = D.diag
 diagview(D::AbstractMatrix) = view(D, diagind(D))
+diagview(D::AbstractVector) = D
```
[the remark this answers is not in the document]
kshyatt: Now removed!

test/ad_utils.jl
```diff
@@ -0,0 +1,31 @@
...
+    mul!(ΔU, U, gaugepart, -1, 1)
+    return ΔU, ΔVᴴ
+end
+function remove_eiggauge_dependence!(
+        ΔV, D, V;
+        degeneracy_atol = MatrixAlgebraKit.default_pullback_gaugetol(S)
+    )
+    gaugepart = V' * ΔV
+    gaugepart[abs.(transpose(diagview(D)) .- diagview(D)) .>= degeneracy_atol] .= 0
+    mul!(ΔV, V / (V' * V), gaugepart, -1, 1)
```
Jutho: Not sure how important this is, because it is only used for tests, but `V / (V' * V)` could probably be replaced with
```julia
Q, R = qr_compact(V)
pinvV = rdiv!(Q, UpperTriangular(R)')
```
for optimal numerical accuracy/stability.

test/mooncake.jl
```diff
@@ -0,0 +1,542 @@
...
+                LAPACK_HouseholderQR(),
+                LAPACK_HouseholderQR(; positive = true),
+            )
+            @testset "qr_compact" begin
+                QR = qr_compact(A, alg)
+                Q = randn(rng, T, m, minmn)
+                R = randn(rng, T, minmn, n)
+                Mooncake.TestUtils.test_rule(rng, qr_compact, A, alg; mode = Mooncake.ReverseMode, is_primitive = false, atol = atol, rtol = rtol)
+                qr_c = (A, QR, alg) -> qr_compact!(MatrixAlgebraKit.copy_input(qr_compact, A), QR, alg)
+                Mooncake.TestUtils.test_rule(rng, qr_c, A, (Q, R), alg; mode = Mooncake.ReverseMode, is_primitive = false, atol = atol, rtol = rtol)
```
Jutho: What is the logic for this extra test, which you don't seem to have for the other functions?

kshyatt: Ah, it's a proof of concept relic from me developing the `test_pullbacks_match`

test/mooncake.jl
```diff
@@ -0,0 +1,542 @@
...
+        CVᴴ = right_orth(A)
+        Mooncake.TestUtils.test_rule(rng, left_orth, A; mode = Mooncake.ReverseMode, atol = atol, rtol = rtol, is_primitive = false)
+        test_pullbacks_match(rng, left_orth!, left_orth, A, VC, (randn(rng, T, size(VC[1])...), randn(rng, T, size(VC[2])...)))
+        Mooncake.TestUtils.test_rule(rng, right_orth, A; mode = Mooncake.ReverseMode, atol = atol, rtol = rtol, is_primitive = false)
+        test_pullbacks_match(rng, right_orth!, right_orth, A, CVᴴ, (randn(rng, T, size(CVᴴ[1])...), randn(rng, T, size(CVᴴ[2])...)))
+
+        left_orth_qr(X) = left_orth(X; alg = :qr)
+        left_orth_polar(X) = left_orth(X; alg = :polar)
+        MatrixAlgebraKit.copy_input(left_orth_qr, A) = MatrixAlgebraKit.copy_input(left_orth, A)
+        MatrixAlgebraKit.copy_input(left_orth_polar, A) = MatrixAlgebraKit.copy_input(left_orth, A)
```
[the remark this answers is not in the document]
Jutho: Also, would it make sense to similarly have
```julia
left_orth_qr(X, args) = left_orth(X, args; alg = :qr)
```
and similar for `_polar`, and use this in `test_pullbacks_match` below.

test/mooncake.jl
```diff
@@ -0,0 +1,542 @@
...
+end
+
+function copy_eigh_trunc!(A, DV, alg; kwargs...)
+    A = (A + A') / 2
+    return eigh_trunc!(A, DV, alg; kwargs...)
+end
+
+MatrixAlgebraKit.copy_input(::typeof(copy_eigh_full), A) = MatrixAlgebraKit.copy_input(eigh_full, A)
+MatrixAlgebraKit.copy_input(::typeof(copy_eigh_vals), A) = MatrixAlgebraKit.copy_input(eigh_vals, A)
+MatrixAlgebraKit.copy_input(::typeof(copy_eigh_trunc), A) = MatrixAlgebraKit.copy_input(eigh_trunc, A)
```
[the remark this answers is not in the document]
kshyatt: Sorry for the rather scrambled-brains answer 🔍

Jutho (approved):

test/mooncake.jl
```diff
@@ -0,0 +1,560 @@
...
+
+The arguments to this function are:
+  - `f!` the mutating, in-place version of the function (accepts `args` for the function result)
+  - `f` the non-mutating version of the function (does not accept `args` for the function result)
+  - `A` the input matrix to factorize
+  - `args` preallocated output for `f!` (e.g. `Q` and `R` matrices for `qr_compact!`)
+  - `Δargs` precomputed derivatives of `args` for pullbacks of `f` and `f!`, to ensure they receive the same input
+  - `alg` optional algorithm keyword argument
+  - `rdata` Mooncake reverse data to supply to the pullback, in case `f` and `f!` return scalar results (as truncating functions do)
+"""
```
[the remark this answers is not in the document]
kshyatt: Done!

lkdvos (commented):
Looks great, thanks a lot for the hard work (and the hard patience!).

I only have some cosmetic comments left, you can choose to ignore this if you like but I think it keeps everything somewhat in the same style.
These should otherwise not make any difference though:

- I think you imported `NoRData` but then qualify `Mooncake.NoRData` everywhere. Since we are already using `CoDual` everywhere and this is a Mooncake extension, it might be fine to simply use `NoRData()` everywhere to lighten the notation a bit.
- In the `@eval` for loops there is the combination of `Symbol`s and `Function` objects. I have absolutely no clue about any performance differences, or burdens on the compiler, but I think in most other parts of the ecosystem we consistently use `Symbol`s, and I think here it wouldn't make a difference either? Correct me if I'm wrong though, no need to spend a whole lot of effort here.
- For the adjoint function names, I don't think the extra `d*` at the front is totally necessary, `qr_adjoint` looks maybe a little cleaner? I checked the Mooncake rrules to see if they have any consistent conventions, and from what I can tell [here](https://github.com/chalk-lab/Mooncake.jl/blob/main/src/rrules/blas.jl) there seems to be a combination of `f_adjoint` and `f_pb!!`. I don't have any strong preference, so maybe `f_adjoint`?

Otherwise definitely a great PR that we will benefit from greatly :)

PR #86, opened 2025-10-22
 Reverse rules for Enzyme

kshyatt:


kshyatt:
OK, I think this is ready for review!

src/pullbacks/eig.jl
```diff
@@ -46,7 +46,8 @@ function eig_pullback!(
         Δgauge ≤ gauge_atol ||
             @warn "`eig` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
 
-        VᴴΔV .*= conj.(inv_safe.(transpose(D) .- D, degeneracy_atol))
+        VᴴΔV ./= conj.(transpose(D) .- D)
+        diagview(VᴴΔV) .= zero(eltype(VᴴΔV))
```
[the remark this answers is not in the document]
kshyatt: OK, with actual testing it seems it is necessary, I think `inv_safe` is still a bit finicky with `Float32`, but it's odd to me that only Enzyme triggers this.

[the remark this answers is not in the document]
lkdvos: It might really be an RNG thing as well? Would it help if we just change the default value of the `degeneracy_atol`?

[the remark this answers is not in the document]
Jutho: The `inv_safe` part should introduce no fewer (but possibly more) zeros than the approach where you only put zeros on the diagonal, i.e. if there are degeneracies between the eigenvalues, the current approach will still be introducing `NaN's or extremely large values on the off-diagonal elements within the degenerate blocks. 

I have to admit I am not sure anymore that the original rule was really all that useful if there are degeneracies, as it does seem that you just bump into the fundamental issue that the eigenvalue decomposition is non-differentiable at those points.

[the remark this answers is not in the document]
kshyatt: Then should we switch this back to `inv_safe` and try to find an input `A` with non degenerate eigenvalues? Or keep as is?

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,371 @@
...
+using Enzyme
+using Enzyme.EnzymeCore
+using Enzyme.EnzymeCore: EnzymeRules
+using LinearAlgebra
+
+@inline EnzymeRules.inactive_type(v::Type{<:MatrixAlgebraKit.AbstractAlgorithm}) = true
+
+function EnzymeRules.augmented_primal(
+        config::EnzymeRules.RevConfigWidth{1},
+        func::Const{typeof(copy_input)},
```
lkdvos: Do we even need to define rules for `copy_input`? I would expect that the AD engines know how to differentiate through that, but I might be missing the underlying reason?

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,371 @@
...
+        ::Type{RT},
+        f::Annotation,
+        A::Annotation
+    ) where {RT}
+    func.val(f.val, A.val)
```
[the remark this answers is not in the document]
kshyatt: Yeah it's effectively copying twice (see line below), I'll fix

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
kshyatt: No it turns out we really do need this rule, I think it's this Julia calling convention bug I found or something similar

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,374 @@
...
+                config::EnzymeRules.RevConfigWidth{1},
+                func::Const{typeof($f)},
+                ::Type{RT},
+                A::Annotation,
+                arg::Annotation{Tuple{TA, TB}},
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT, TA, TB}
+            cache_arg = nothing
+            # form cache if needed
+            cache_A = !(typeof(arg) <: Const) ? copy(A.val) : nothing
```
Jutho: I prefer the `isa` construction

```suggestion
            cache_A = !isa(arg, Const) ? copy(A.val) : nothing
```

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,374 @@
...
+                arg::Annotation{Tuple{TA, TB}},
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT, TA, TB}
+            cache_A, cache_arg = cache
+            argval = arg.val
+            Aval = !isnothing(cache_A) ? cache_A : A.val
+            ∂arg = isa(arg, Const) ? nothing : arg.dval
+            if !isa(A, Const) && !isa(arg, Const)
+                $pb(A.dval, Aval, argval, ∂arg)
+            end
```
Jutho: In the case where `arg` is not actually containing the outputs, will `∂arg` contain the adjoints to the output? I am very confused by how all of this works.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,374 @@
...
+                ::Type{RT},
+                cache,
+                A::Annotation,
+                D::Annotation,
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT}
+
+            cache_A, cache_D, V = cache
+            Dval = !isnothing(cache_D) ? cache_D : D.val
+            Aval = !isnothing(cache_A) ? cache_A : A.val
```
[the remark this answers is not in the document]
kshyatt: Done!

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
kshyatt: This has been fixed now, I think

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,387 @@
...
+                ::Type{RT},
+                A::Annotation,
+                arg::Annotation{Tuple{TA, TB}},
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT, TA, TB}
+            cache_arg = nothing
+            # form cache if needed
+            cache_A = isa(arg, Const) ? copy(A.val) : nothing
+            ret = func.val(A.val, arg.val, alg.val)
+            dret = (TA == Nothing && TA == Nothing) ? make_zero!.(similar.(ret)) : arg.dval
```
Jutho: Not sure if this is what you intended. Also, can we use `MatrixAlgebraKit.zero!` for `make_zero!`?

```suggestion
            dret = (TA == Nothing && TB == Nothing) ? make_zero!.(similar.(ret)) : arg.dval
```

kshyatt:
FWIW I think the Mac runners are very underpowered. I may just run the tests on linux for now if we're in CI mode. On my Mac laptop they run no problem.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,403 @@
...
+        (qr_full!, qr_pullback!),
+        (lq_full!, lq_pullback!),
+        (qr_compact!, qr_pullback!),
+        (lq_compact!, lq_pullback!),
+        (eig_full!, eig_pullback!),
+        (eigh_full!, eigh_pullback!),
+        (left_polar!, left_polar_pullback!),
+        (right_polar!, right_polar_pullback!),
+    )
+    @eval begin
```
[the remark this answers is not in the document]
kshyatt: I think 1 & 2 are now resolved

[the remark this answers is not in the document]
Jutho: These were genuine questions, so I don't know if they can be "resolved" in the usual sense, rather than by improving my understanding 😄 . 

I am also trying to understand whether these Enzyme "bindings" are completely generic, and would thus automatically deal with the `TensorMap` case as well, provided our pullback functions are overloaded. They mostly seem to be, except for maybe the occasional broadcast add `.+=`. Maybe VectorInterface.jl could help with that (in a later stage).

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,401 @@
...
+                config::EnzymeRules.RevConfigWidth{1},
+                func::Const{typeof($f)},
+                ::Type{RT},
+                A::Annotation,
+                arg::Annotation,
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT}
+            cache_A = copy(A.val)
+            ret = func.val(A.val, arg.val, alg.val)
+            dret = isa(arg, Const) ? nothing : arg.dval
```
Jutho: Why is the condition on `dret` here very different than in the previous set of methods. I would expect that the only difference comes from having a single rather than 2 output arguments, so I would expect something like
`dret = arg.val isa Nothing ? zero(ret) : arg.dval`
as analogue of line 99 above.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,401 @@
...
+                dret::Type{RT},
+                cache,
+                A::Annotation,
+                arg::Annotation,
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT}
+            cache_A, darg = cache
+            Aval = something(cache_A, A.val)
+            if !isa(A, Const)
+                $pb(A.dval, Aval, arg.val, darg)
```
Jutho: Same question about `darg` vs `dret`

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,374 @@
...
+                if $(f == svd_compact!) # compact
+                    svd_pullback!(A.dval, Aval, USVᴴval, ∂USVᴴ)
+                else # full
+                    vU = view(U, :, 1:minmn)
+                    vS = Diagonal(diagview(S)[1:minmn])
+                    vVᴴ = view(Vᴴ, 1:minmn, :)
+                    vdU = view(∂USVᴴ[1], :, 1:minmn)
+                    vdS = Diagonal(diagview(∂USVᴴ[2])[1:minmn])
+                    vdVᴴ = view(∂USVᴴ[3], 1:minmn, :)
+                    svd_pullback!(A.dval, Aval, (vU, vS, vVᴴ), (vdU, vdS, vdVᴴ))
```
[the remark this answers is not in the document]
Jutho: Can you add some `# TODO:` comment below the `else` of line 210 to remind me that this needs to be fixed in `svd_pullback!`.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,401 @@
...
+function EnzymeRules.reverse(
+        config::EnzymeRules.RevConfigWidth{1},
+        func::Const{typeof(svd_trunc_no_error!)},
+        dret::Type{RT},
+        cache,
+        A::Annotation,
+        USVᴴ::Annotation,
+        alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+    ) where {RT}
+    cache_A, cache_USVᴴ, shadow_USVᴴ, ind = cache
```
Jutho: `cache_A, cache_USVᴴ, dUSVᴴ, ind = cache`

and the same `dUSVᴴ` vs `dret` question.

Jutho (commented):
I mostly have some remaining comments about consistent variable naming, and some final questions to help me resolve my confusions.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,400 @@
...
+                func::Const{typeof($f)},
+                ::Type{RT},
+                A::Annotation,
+                USVᴴ::Annotation{Tuple{TU, TS, TVᴴ}},
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT, TU, TS, TVᴴ}
+            # form cache if needed
+            cache_A = !isa(A, Const) ? copy(A.val) : nothing
+            ret = func.val(A.val, USVᴴ.val, alg.val)
+            cache_USVᴴ = EnzymeRules.overwritten(config)[3] ? copy(ret) : nothing
```
Jutho: Doesn't this need to be
```julia
cache_USVᴴ = !isa(USVᴴ, Const) && EnzymeRules.overwritten(config)[3] ? copy(ret) : nothing
```

I am also confused about why here `cache_USVᴴ` is `copy(ret)` whereas in the methods above `cache_arg = copy(arg.val)` and this is done before even applying `ret = func(...)`?

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,400 @@
...
+                # seems to be necessary due to Enzyme's type analysis
+                dS = $(f == svd_compact!) ? Diagonal(zero(ret[2].diag)) : zero(ret[2])
+                dVᴴ = zero(ret[3])
+                (dU, dS, dVᴴ)
+            else
+                USVᴴ.dval
+            end
+            primal = EnzymeRules.needs_primal(config) ? ret : nothing
+            shadow = EnzymeRules.needs_shadow(config) ? dret : nothing
+            return EnzymeRules.AugmentedReturn(primal, shadow, (cache_A, cache_USVᴴ, ret, dret))
```
Jutho: More confusion here: why does the cache contain both `cache_USVᴴ` and `ret`, when `cache_USVᴴ` is either nothing or a copy of `ret`?

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,400 @@
...
+            func::Const{typeof($f)},
+            ::Type{RT},
+            A::Annotation,
+            DV::Annotation{Tuple{TA, TB}},
+            alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+        ) where {RT, TA, TB}
+        # form cache if needed
+        cache_A = !isa(A, Const) ? copy(A.val) : nothing
+        ret = $full_f(A.val, DV.val, alg.val.alg)
+        cache_DV = copy.(ret)
```
Jutho: Same question here about the need of copying `ret`.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,400 @@
...
+                config::EnzymeRules.RevConfigWidth{1},
+                func::Const{typeof($f)},
+                ::Type{RT},
+                A::Annotation,
+                arg::Annotation{Tuple{TA, TB}},
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT, TA, TB}
+            # form cache if needed
+            cache_A = !isa(A, Const)   && EnzymeRules.overwritten(config)[2] ? copy(A.val) : nothing
+            cache_arg = !isa(arg, Const) && EnzymeRules.overwritten(config)[3] ? copy(arg.val) : nothing
```
Jutho: Below you write `copy.` when it acts on a tuple, which I kind of like, so I would suggest:

```suggestion
            cache_arg = !isa(arg, Const) && EnzymeRules.overwritten(config)[3] ? copy.(arg.val) : nothing
```

kshyatt:
Sorry, some of these were oversights based on the values that the API accepted changing underneath this PR as it waited and waited, I'll fix.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
kshyatt: https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/86#discussion_r2710179174 also applies I think

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
kshyatt: Also, if USVh are all nothings, their annotation may be Const even though we would want to copy the output `ret` in case **`ret`** is modified further down the call graph

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,400 @@
...
+        func::Const{typeof(svd_trunc_no_error!)},
+        ::Type{RT},
+        A::Annotation,
+        USVᴴ::Annotation{Tuple{TU, TS, TVᴴ}},
+        alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+    ) where {RT, TU, TS, TVᴴ}
+    # form cache if needed
+    cache_A = !isa(A, Const) ? copy(A.val) : nothing
+    ret = svd_compact!(A.val, USVᴴ.val, alg.val.alg)
+    cache_USVᴴ = copy.(ret)
```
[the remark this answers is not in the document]
Jutho: Oh I see. Even though we don't return `ret` but `USVᴴ′`, `ret` is probably stored in `USVᴴ` and so these arguments are still accessible later in the call chain.

Jutho:
My apologies for my neurotic mind tripping over the apparent inconsistencies. I just want to make everything as homogeneous as possible (with of course special casing for the truncating methods), so that these rules keep working for future algorithms that might have a different behavior with respect to the relation between `arg.val` and `ret`. Also, it will just be impossible to understand these rules in a few months time (at least for me) if they have these subtle differences.

kshyatt:
OK, made the caching of arg and ret a lot more consistent

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,410 @@
...
+            cache_A = !isa(A, Const) ? copy(A.val) : nothing
+            ret = func.val(A.val, USVᴴ.val, alg.val)
+            # if USVᴴ.val == ret, the annotation must be Duplicated or DuplicatedNoNeed
+            # if USVᴴ isa Const, ret may still be modified further down the call graph so we should
+            # copy it to protect ourselves
+            cache_USVᴴ = (USVᴴ.val !== ret) || (!isa(USVᴴ, Const) && EnzymeRules.overwritten(config)[3]) ? copy.(ret) : nothing
+            # the USVᴴ may be nothing for eltypes handled by GenericLinearAlgebra
+            dret = if (TU == TS == TVᴴ == Nothing)
+                dU = zero(ret[1])
+                # seems to be necessary due to Enzyme's type analysis
```
Jutho:
```suggestion
                # special casing `Diagonal` seems to be necessary due to Enzyme's type analysis
```

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,410 @@
...
+                config::EnzymeRules.RevConfigWidth{1},
+                func::Const{typeof($f)},
+                ::Type{RT},
+                cache,
+                A::Annotation,
+                USVᴴ::Annotation,
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT}
+            cache_A, USVᴴval, dUSVᴴ = cache
+            Aval = something(cache_A, A.val)
```
Jutho: We might put `nothing` in the cache for `cache_USVᴴ`, so I think we need this:
```suggestion
            cache_A, cache_USVᴴ, dUSVᴴ = cache
            Aval = something(cache_A, A.val)
            USVᴴval = something(cache_USVᴴ, USVᴴ.val)
```

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,410 @@
...
+        ::Type{RT},
+        cache,
+        A::Annotation,
+        USVᴴ::Annotation,
+        alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+    ) where {RT}
+    cache_A, cache_USVᴴ, dUSVᴴ, ind = cache
+    Aval = something(cache_A, A.val)
+    if !isa(A, Const)
+        svd_pullback!(A.dval, Aval, cache_USVᴴ, dUSVᴴ, ind)
```
Jutho: Same here: we might be storing `nothing` in `cache_USVᴴ`:

```suggestion
    cache_A, cache_USVᴴ, dUSVᴴ, ind = cache
    Aval = something(cache_A, A.val)
    USVᴴval = something(cache_USVᴴ, USVᴴ.val)
    if !isa(A, Const)
        svd_pullback!(A.dval, Aval, USVᴴval, dUSVᴴ, ind)
```

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,410 @@
...
+    @eval begin
+        function EnzymeRules.augmented_primal(
+                config::EnzymeRules.RevConfigWidth{1},
+                func::Const{typeof($f!)},
+                ::Type{RT},
+                A::Annotation,
+                D::Annotation,
+                alg::Annotation{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT}
+            cache_A = !isa(A, Const) && EnzymeRules.overwritten(config)[2] ? copy(A.val) : nothing
```
Jutho: The ` EnzymeRules.overwritten(config)[2]` for deciding on caching `A` is not used consistently. Can we have this everywhere?

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,410 @@
...
+        func::Const{typeof(svd_vals!)},
+        ::Type{RT},
+        cache,
+        A::Annotation,
+        S::Annotation,
+        alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+    ) where {RT}
+    cache_A, cache_S, dS, U, Vᴴ = cache
+    Sval = something(cache_S, S.val)
+    Aval = something(cache_A, A.val)
```
Jutho: Me being me:
```suggestion
    Aval = something(cache_A, A.val)
    Sval = something(cache_S, S.val)
```

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
kshyatt: Going off their internal rules https://github.com/EnzymeAD/Enzyme.jl/blob/main/src/internal_rules/linalg.jl#L35 it seems we can remove the `Const` condition. I think this again an artifact of an older Enzyme version that hung around because I was loath to change things that were working and didn't impact performance much.

kshyatt: Removed a bunch of the `isa(A, Const)` checks in favour of `overwritten`

kshyatt:
If the return is e.g. a `Duplicated` and freshly allocated we can put a _reference_ to its derivative on the tape so that we can peek into it later for the pullback, which is what we're currently doing

kshyatt:
OK, I think we must always copy `A.val` for the `qr_null` or `lq_null` cases as those pullbacks do depend directly on `A`, NOT just on `dA`, `N`, and `dN`.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,455 @@
...
+    USVᴴ′, ind = MatrixAlgebraKit.truncate(svd_trunc!, ret, alg.val.trunc)
+    primal = EnzymeRules.needs_primal(config) ? USVᴴ′ : nothing
+    # This creates new output shadow matrices, we use USVᴴ′ to ensure the
+    # eltypes and dimensions are correct.
+    # These new shadow matrices are "filled in" with the accumulated
+    # results from earlier in reverse-mode AD after this function exits
+    # and before `reverse` is called.
+    dret = (zero(USVᴴ′[1]), Diagonal(zero(USVᴴ′[2].diag)), zero(USVᴴ′[3]))
+    shadow = EnzymeRules.needs_shadow(config) ? dret : nothing
+    return EnzymeRules.AugmentedReturn(primal, shadow, (cache_USVᴴ, dret, ind))
```
Jutho:
```suggestion
    dret = if EnzymeRules.needs_shadow(config)
        (zero(USVᴴ′[1]), Diagonal(zero(USVᴴ′[2].diag)), zero(USVᴴ′[3]))
    else
        nothing
    end
    return EnzymeRules.AugmentedReturn(primal, dret, (cache_USVᴴ, dret, ind))
```

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -0,0 +1,458 @@
...
+        # A is overwritten in the primal, but NOT used in the pullback,
+        # so we do not need to cache it. This may change if future pullbacks
+        # depend directly on A!
+        ret = $full_f(A.val, DV.val, alg.val.alg)
+        cache_DV = (DV.val !== ret) || EnzymeRules.overwritten(config)[3] ? copy.(ret) : nothing
+        DV′, ind = truncate($trunc_f, ret, alg.val.trunc)
+        primal = EnzymeRules.needs_primal(config) ? DV′ : nothing
+        dret = (Diagonal(zero(diagview(DV′[1]))), zero(DV′[2]))
+        shadow = EnzymeRules.needs_shadow(config) ? dret : nothing
+        return EnzymeRules.AugmentedReturn(primal, shadow, (cache_DV, dret, ind))
```
Jutho:
```suggestion
        dret = if EnzymeRules.needs_shadow(config)
            (Diagonal(zero(diagview(DV′[1]))), zero(DV′[2]))
        else
            nothing
        end
        return EnzymeRules.AugmentedReturn(primal, dret, (cache_DV, dret, ind))
```

PR #87, opened 2025-10-29
add support for BigFloats via a new extension

sanderdemeyer:
Add an extension using GenericLinearAlgebra and GenericSchur to be able to deal with `BigFloat`s. Some comments are in order.

- Currently, only the QR, LQ, SVD, eig, and eigh decompositions are defined, the rest is not. Also `qr_null` and `lq_null` are not implemented.
- Both qr_full and qr_compact are defined. In this implementation, the former is based on the latter, where a unitary is constructed via a Gram-Schmidt procedure. Hence, the Q resulting from `qr_full` is not unique, and in the tests, I only cover the columns of `Q` arising from the `qr_compact`. The same holds for the LQ decomposition.
- In the QR and LQ decompositions, only `pivoted = false` and `blocksize = 1` are supported. The other options throw an explicit error.
- In some tests, I had to change `===` to `\approx`, since they would otherwis fail. I don't know whether this has an easy solution, or whether this is something that can not be enforced in the generic case.
- GenericLinearAlgebra.jl has all required functions, apart from the eigendecomposition for non-hermitian matrices (although eigvals is supported for non-hermitian matrices). Hence, I added GenericSchur.jl to cover this.
- One of the main reasons I want to have support for BigFloats in MatrixAlgebraKit is to have support for the exponential of a matrix of BigFloats. This functionality is currently not defined here, but in TensorKit.jl. The only thing that is called there is `LinearAlgebra.exp!`. Since it might be cleaner to only define an extension for MatrixAlgebraKit and not for TensorKit, it might be beneficial to move the definition of exp! to MatrixAlgebraKit. If there are alternatives or suggestions, please let me know.

### DIFF
--- a/test/genericlinearalgebra/lq.jl
+++ b/test/genericlinearalgebra/lq.jl
@@ -122,28 +122,3 @@ end
         @test Q[1:minmn, n] ≈ Q2[1:minmn, n]
     end
 end
-
-@testset "lq_compact for Diagonal{$T}" for T in eltypes
-    rng = StableRNG(123)
-    atol = eps(real(T))^(3 / 4)
-    for m in (54, 0)
-        Ad = randn(rng, T, m)
-        A = Diagonal(Ad)
-
-        # compact
-        L, Q = @constinferred lq_compact(A)
-        @test Q isa Diagonal{T} && size(Q) == (m, m)
-        @test L isa Diagonal{T} && size(L) == (m, m)
-        @test L * Q ≈ A
-        @test isunitary(Q)
-
-        # compact and positive
-        Lp, Qp = @constinferred lq_compact(A; positive = true)
-        @test Qp isa Diagonal{T} && size(Qp) == (m, m)
-        @test Lp isa Diagonal{T} && size(Lp) == (m, m)
-        @test Lp * Qp ≈ A
-        @test isunitary(Qp)
-        @test all(≥(zero(real(T))), real(diag(Lp))) &&
-            all(≈(zero(real(T)); atol), imag(diag(Lp)))
-    end
-end
--- a/test/genericlinearalgebra/qr.jl
+++ b/test/genericlinearalgebra/qr.jl
@@ -107,28 +107,3 @@ end
         end
     end
 end
-
-@testset "qr_compact for Diagonal{$T}" for T in eltypes
-    rng = StableRNG(123)
-    atol = eps(real(T))^(3 / 4)
-    for m in (54, 0)
-        Ad = randn(rng, T, m)
-        A = Diagonal(Ad)
-
-        # compact
-        Q, R = @constinferred qr_compact(A)
-        @test Q isa Diagonal{T} && size(Q) == (m, m)
-        @test R isa Diagonal{T} && size(R) == (m, m)
-        @test Q * R ≈ A
-        @test isunitary(Q)
-
-        # compact and positive
-        Qp, Rp = @constinferred qr_compact(A; positive = true)
-        @test Qp isa Diagonal{T} && size(Qp) == (m, m)
-        @test Rp isa Diagonal{T} && size(Rp) == (m, m)
-        @test Qp * Rp ≈ A
-        @test isunitary(Qp)
-        @test all(≥(zero(real(T))), real(diag(Rp))) &&
-            all(≈(zero(real(T)); atol), imag(diag(Rp)))
-    end
-end
--- a/test/eig.jl
+++ b/test/eig.jl
@@ -5,7 +5,8 @@ using StableRNGs
 using LinearAlgebra: Diagonal
 using MatrixAlgebraKit: TruncatedAlgorithm, diagview, norm
 
-const BLASFloats = (Float32, Float64, ComplexF32, ComplexF64)
+BLASFloats = (Float32, Float64, ComplexF32, ComplexF64)
+GenericFloats = (Float16, BigFloat, Complex{BigFloat})
 
 @testset "eig_full! for T = $T" for T in BLASFloats
     rng = StableRNG(123)
@@ -91,7 +92,7 @@ end
     @test ϵ3 ≈ norm(diagview(D)[3:4]) atol = atol
 end
 
-@testset "eig for Diagonal{$T}" for T in BLASFloats
+@testset "eig for Diagonal{$T}" for T in (BLASFloats..., GenericFloats...)
     rng = StableRNG(123)
     m = 54
     Ad = randn(rng, T, m)
--- a/test/eigh.jl
+++ b/test/eigh.jl
@@ -5,7 +5,8 @@ using StableRNGs
 using LinearAlgebra: LinearAlgebra, Diagonal, I
 using MatrixAlgebraKit: TruncatedAlgorithm, diagview, norm
 
-const BLASFloats = (Float32, Float64, ComplexF32, ComplexF64)
+BLASFloats = (Float32, Float64, ComplexF32, ComplexF64)
+GenericFloats = (Float16, BigFloat, Complex{BigFloat})
 
 @testset "eigh_full! for T = $T" for T in BLASFloats
     rng = StableRNG(123)
@@ -100,7 +101,7 @@ end
     @test ϵ3 ≈ norm(diagview(D)[3:4]) atol = atol
 end
 
-@testset "eigh for Diagonal{$T}" for T in BLASFloats
+@testset "eigh for Diagonal{$T}" for T in (BLASFloats..., GenericFloats...)
     rng = StableRNG(123)
     m = 54
     Ad = randn(rng, T, m)
--- a/test/genericlinearalgebra/eigh.jl
+++ b/test/genericlinearalgebra/eigh.jl
@@ -91,28 +91,3 @@ end
     @test diagview(D3) ≈ diagview(D)[1:2]
     @test ϵ3 ≈ norm(diagview(D)[3:4]) atol = atol
 end
-
-@testset "eigh for Diagonal{$T}" for T in eltypes
-    rng = StableRNG(123)
-    m = 54
-    Ad = randn(rng, T, m)
-    Ad .+= conj.(Ad)
-    A = Diagonal(Ad)
-    atol = sqrt(eps(real(T)))
-
-    D, V = @constinferred eigh_full(A)
-    @test D isa Diagonal{real(T)} && size(D) == size(A)
-    @test V isa Diagonal{T} && size(V) == size(A)
-    @test A * V ≈ V * D
-
-    D2 = @constinferred eigh_vals(A)
-    @test D2 isa AbstractVector{real(T)} && length(D2) == m
-    @test diagview(D) ≈ D2
-
-    A2 = Diagonal(T[0.9, 0.3, 0.1, 0.01])
-    alg = TruncatedAlgorithm(DiagonalAlgorithm(), truncrank(2))
-    D2, V2, ϵ2 = @constinferred eigh_trunc(A2; alg)
-    @test diagview(D2) ≈ diagview(A2)[1:2]
-    @test ϵ2 ≈ norm(diagview(A2)[3:4]) atol = atol
-
-end
--- a/test/genericlinearalgebra/svd.jl
+++ b/test/genericlinearalgebra/svd.jl
@@ -169,39 +169,3 @@ end
     @test ϵ2 ≈ norm(diagview(S)[3:4]) atol = atol
     @test_throws ArgumentError svd_trunc(A; alg, trunc = (; maxrank = 2))
 end
-
-@testset "svd for Diagonal{$T}" for T in eltypes
-    rng = StableRNG(123)
-    atol = sqrt(eps(real(T)))
-    for m in (54, 0)
-        Ad = randn(T, m)
-        A = Diagonal(Ad)
-
-        U, S, Vᴴ = @constinferred svd_compact(A)
-        @test U isa AbstractMatrix{T} && size(U) == size(A)
-        @test Vᴴ isa AbstractMatrix{T} && size(Vᴴ) == size(A)
-        @test S isa Diagonal{real(T)} && size(S) == size(A)
-        @test isunitary(U)
-        @test isunitary(Vᴴ)
-        @test all(≥(0), diagview(S))
-        @test A ≈ U * S * Vᴴ
-
-        U, S, Vᴴ = @constinferred svd_full(A)
-        @test U isa AbstractMatrix{T} && size(U) == size(A)
-        @test Vᴴ isa AbstractMatrix{T} && size(Vᴴ) == size(A)
-        @test S isa Diagonal{real(T)} && size(S) == size(A)
-        @test isunitary(U)
-        @test isunitary(Vᴴ)
-        @test all(≥(0), diagview(S))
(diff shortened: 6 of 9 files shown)


ext/MatrixAlgebraKitGenericExt.jl
```diff
@@ -0,0 +1,230 @@
...
+    return BigFloat_svd_QRIteration()
+end
+
+function MatrixAlgebraKit.svd_compact!(A::AbstractMatrix{T}, USVᴴ, alg::BigFloat_svd_QRIteration) where {T <: Union{BigFloat, Complex{BigFloat}}}
+    check_input(svd_compact!, A, USVᴴ, alg)
+    U, S, V = GenericLinearAlgebra.svd(A)
+    return U, Diagonal(S), V' # conjugation to account for difference in convention
+end
+
+function MatrixAlgebraKit.svd_full!(A::AbstractMatrix{T}, USVᴴ, alg::BigFloat_svd_QRIteration)::Tuple{Matrix{T}, Matrix{BigFloat}, Matrix{T}} where {T <: Union{BigFloat, Complex{BigFloat}}}
```
lkdvos: Are the output type annotations required here (and below)? Is this because `GenericLinearAlgebra.svd` isn't type stable?

ext/MatrixAlgebraKitGenericExt.jl
```diff
@@ -0,0 +1,230 @@
...
+        return USVᴴ
+    end
+    S̃ = zeros(eltype(S), size(S))
+    U_compact, S_compact, V_compact = GenericLinearAlgebra.svd(A)
+    S̃[1:minmn, 1:minmn] .= Diagonal(S_compact)
+    Ũ = _gram_schmidt(U_compact)
+    Ṽ = _gram_schmidt(V_compact)
+
+    copyto!(U, Ũ)
+    copyto!(S, S̃)
```
lkdvos: Would it be possible to have:

```suggestion
    U = _gram_schmidt!(U, U_compact)
    V = _gram_schmidt!(V, V_compact)
```

ext/MatrixAlgebraKitGenericExt.jl
```diff
@@ -0,0 +1,230 @@
...
+    computeR = length(R) > 0
+
+    Q_zero = zeros(eltype(Q), (m, minmn))
+    R_zero = zeros(eltype(R), (minmn, n))
+    Q_compact, R_compact = _bigfloat_householder_qr!(A, Q_zero, R_zero; alg.kwargs...)
+    copyto!(Q, _gram_schmidt(Q_compact[:, 1:min(m, n)]))
+    if computeR
+        R̃ = zeros(eltype(R), m, n)
+        R̃[1:minmn, 1:n] .= R_compact
+        copyto!(R, R̃)
```
lkdvos: Can this be done directly in the correct destination?
```suggestion
        R = fill!(R, zero(eltype(R)))
        R[1:minmn, 1:n] .= R_compact
```

lkdvos (commented):
Looks really cool, very nice work!

I have a couple questions that are mostly due to the fact that I am less familiar with these packages:

Do these packages have to come together, or are they really separate entities? From a quick glance it seems like `GenericSchur` is being used for `eig`, while `GenericLinearAlgebra` is used for the rest. I was under the impression that `GenericLinearAlgebra` also had a non-symmetric eigenvalue implementation but that might be wrong.
Anyways, the reason I'm asking is because I was wondering if it would make sense to simply have one extension for each, rather than requiring both.

It looks slightly odd to me to call these algorithms `BigFloat_*`, since IIRC these should work for generic element types, right? How would you feel about `GenericLinearAlgebra_HouseholderQR`, or simply `GLA_HouseholderQR` etc? Similarly, I think there are some type restrictions on the functions that are might not be required.

Considering the allocation of the output, do you know if we can pass the output to the implementations of these libraries? If not, that's definitely also okay, but in that case we should probably intercept allocating them in the first place, if they will be ignored. This can be achieved by overloading `initialize_output(...) -> Nothing` and altering the checks accordingly.

---

For the tests, should we try and invest in reducing the code duplication a bit for the different algorithms? Would it make sense to simply add the eltypes to the already existing tests? This also relates to the GPU tests, @kshyatt might be interested in this as well.

---

I left some other comments throughout the code, but overall this looks like an amazing addition.

ext/MatrixAlgebraKitGenericExt.jl
```diff
@@ -0,0 +1,230 @@
...
+    U, S, Vᴴ = USVᴴ
+    m, n = size(A)
+    minmn = min(m, n)
+    if minmn == 0
+        MatrixAlgebraKit.one!(U)
+        MatrixAlgebraKit.zero!(S)
+        MatrixAlgebraKit.one!(Vᴴ)
+        return USVᴴ
+    end
+    S̃ = zeros(eltype(S), size(S))
```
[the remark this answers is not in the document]
sanderdemeyer: Yes, good catch.

kshyatt:
Re the test duplications, I think it could well make sense to do something similar to what `GPUArrays` does with its test suite: define compatible array and element type pairs and run the generically written testsuite on them. But that shouldn't hold this PR back, should we maybe open an issue for it?

kshyatt:
> In some tests, I had to change === to \approx

Does this mean that copies are happening where they shouldn't? What happens if you use `==` rather than `===`?

ext/MatrixAlgebraKitGenericExt.jl
sanderdemeyer: I was apparently too strict. For the svd it is indeed not necessary. The only function that requires it is `MatrixAlgebraKit.eigh_full!`, since that uses `GenericLinearAlgebra.eigen`, which is indeed type unstable. I guess that arises from the fact that for hermitian and non-hermitian matrices, they use different subroutines, so they don't know beforehand whether the eigenvalues will be real or not (and we require the former in `eigh_full`)

ext/MatrixAlgebraKitGenericLinearAlgebraExt.jl
```diff
@@ -0,0 +1,161 @@
...
+end
+
+function _gla_householder_qr!(A::AbstractMatrix{T}, Q, R; positive = false, blocksize = 1, pivoted = false) where {T}
+    pivoted && throw(ArgumentError("Only pivoted = false implemented for GLA_QR_Householder."))
+    (blocksize == 1) || throw(ArgumentError("Only blocksize = 1 implemented for GLA_QR_Householder."))
+
+    m, n = size(A)
+    k = min(m, n)
+    computeR = length(R) > 0
+    Q̃, R̃ = qr(A)
```
Jutho: This can definitely be improved. GenericLinearAlgebra.jl has both a `qrUnblocked!` and a `qrBlocked!`. They act similar to Lapack, i.e. after calling `F = qrUnblocked!(A)`, the elements on and above the diagonal of `A` (with `F.factors === A`) encode `R`, whereas the elements below the diagonal of `A` as well as a new vector `F.τ` encode the Householder reflectors that encode `Q`. From this, you can reconstruct both `qr_compact!`, `qr_full!` as well as `qr_null!`.

However, all of this is quite cryptic if you are not familiar with these ancient LAPACK storage strategies.

ext/MatrixAlgebraKitGenericLinearAlgebraExt.jl
```diff
@@ -0,0 +1,161 @@
...
+    minmn = min(m, n)
+    computeR = length(R) > 0
+
+    Q_zero = zeros(eltype(Q), (m, minmn))
+    R_zero = zeros(eltype(R), (minmn, n))
+    Q_compact, R_compact = _gla_householder_qr!(A, Q_zero, R_zero; alg.kwargs...)
+    Q = _gram_schmidt!(Q, Q_compact[:, 1:min(m, n)])
+    if computeR
+        R = fill!(R, zero(eltype(R)))
+        R[1:minmn, 1:n] .= R_compact
```
lkdvos:
```suggestion
        R[1:minmn, :] .= R_compact
        R[(minmn + 1):end, :] .= zero(eltype(R))
```

src/interface/decompositions.jl
```diff
@@ -100,6 +104,9 @@ const LAPACK_EighAlgorithm = Union{
     LAPACK_MultipleRelativelyRobustRepresentations,
 }
 
+# TODO:
+@algdef GLA_eigh_Francis
```
lkdvos:
```suggestion
@algdef GLA_Francis
```

ext/MatrixAlgebraKitGenericLinearAlgebraExt.jl
```diff
@@ -0,0 +1,161 @@
...
+using MatrixAlgebraKit
+using MatrixAlgebraKit: sign_safe, check_input
+using GenericLinearAlgebra: svd, svdvals!, eigen, eigvals, Hermitian, qr
+using LinearAlgebra: I, Diagonal
+
+function MatrixAlgebraKit.default_svd_algorithm(::Type{T}; kwargs...) where {T <: StridedMatrix{<:Union{BigFloat, Complex{BigFloat}}}}
+    return GLA_svd_QRIteration()
+end
+
+function MatrixAlgebraKit.svd_compact!(A::AbstractMatrix{T}, USVᴴ, alg::GLA_svd_QRIteration) where {T}
```
lkdvos:
```suggestion
function MatrixAlgebraKit.svd_compact!(A, USVᴴ, alg::GLA_svd_QRIteration)
```

Jutho:
I had a look at GenericSchur.jl, which I was much less familiar with than GenericLinearAlgebra.jl. That seems to reimplement a whole lot of LAPACK functionality related to Schur and eigenvalue decompositions, and so it might be possible to write the wrappers for `schur_...`, `eig_` and `eigh_` in a manner that is much closer in spirit to the LAPACK algorithms, and in particular writes directly to the given output arrays.

So I think it would definitely be good to split this PR into two parts corresponding to the two extensions.

GenericSchur.jl seems to provide both a `QRIteration` (which equals the current `Francis` name) and a `DivideAndConquer` implementation for the hermitian eigenvalue problem, and the standard `QRIteration` algorithm for the nonhermitian one (which for one or the other reason is not called as such in LAPACK). So we might want to think about the best naming strategy for these algorithms as well. I would definitely not call them `GLA_...` if they come from GenericSchur.jl, as that package might be worth while using in its own.

In case anyone is interested in reading more about the QR iteration method, Chapter 4 in https://people.inf.ethz.ch/arbenz/ewp/Lnotes/lsevp.pdf is really good.
My old implementation (which I still plan to include as well), is based on that.

ext/MatrixAlgebraKitGenericLinearAlgebraExt.jl
Jutho:
```suggestion
    Q̃, R̃ = qr!(A)
```

ext/MatrixAlgebraKitGenericLinearAlgebraExt.jl
```diff
@@ -0,0 +1,161 @@
...
+    eigval, eigvec = eigen(Hermitian(A); sortby = real)
+    copyto!(D, Diagonal(eigval))
+    copyto!(V, eigvec)
+    return D, V
+end
+
+function MatrixAlgebraKit.eigh_vals!(A::AbstractMatrix{T}, D, alg::GLA_eigh_Francis) where {T}
+    check_input(eigh_vals!, A, D, alg)
+    D = eigvals(A; sortby = real)
+    return real.(D)
```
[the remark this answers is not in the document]
sanderdemeyer: Indeed `Hermitian` fixes this. `eigvals!` also exists and is now used.

src/interface/decompositions.jl
```diff
@@ -32,6 +32,16 @@ elements of `L` are non-negative.
 """
 @algdef LAPACK_HouseholderLQ
 
+"""
+    GLA_HouseholderQR(; positive = false)
+
+Algorithm type to denote the GenericLinearAlgebra.jl implementation for computing the QR decomposition 
+of a matrix using Householder reflectors. Currenlty, only `blocksize = 1` and `pivoted == false`
```
Jutho:
```suggestion
of a matrix using Householder reflectors. Currently, only `blocksize = 1` and `pivoted == false` 
```

test/genericlinearalgebra/lq.jl
```diff
@@ -0,0 +1,149 @@
...
+        lq_full!(copy!(Ac, A), (L, Q); positive = true, blocksize = 1)
+        @test L * Q ≈ A
+        @test isunitary(Q)
+        @test all(>=(zero(real(T))), real(diag(L)))
+        lq_full!(copy!(Ac, A), (noL, Q2); positive = true, blocksize = 1)
+        @test Q[1:minmn, n] ≈ Q2[1:minmn, n]
+    end
+end
+
+@testset "lq_compact for Diagonal{$T}" for T in eltypes
```
Jutho: This `testset` is not testing anything related to GenericLinearAlgebra.jl.

test/genericlinearalgebra/qr.jl
```diff
@@ -0,0 +1,134 @@
...
+            # element
+            for j in 1:n
+                i = findlast(!iszero, view(R, :, j))
+                @test real(R[i, j]) >= zero(real(T))
+            end
+        end
+    end
+end
+
+@testset "qr_compact for Diagonal{$T}" for T in eltypes
```
Jutho: This `testset` is not testing anything related to GenericLinearAlgebra.jl.

Jutho:
Aside from the small comments above, I am a bit torn about is about registering these methods as the `default_....` for `StridedMatrix{BigFloat}` and `StridedMatrix{Complex{BigFloat}}`.

On the one hand, they probably work for a larger set of types, such as `Float16` eltype, or non-strided matrices.

On the other hand, especially with having in mind that we might have alternative `Native_` algorithms for some of those, I am wondering if this is the best choice.

However, without these defaults in place, you would not be able to compute a matrix factorization with `BigFloat` entries without specifying the `GLA_...` algorithm explicitly, which is probably not what you want in generic code. If you explicitly load `GenericLinearAlgebra`, you probably want to use it.

I am however wondering what happens if you are `using MatrixAlgebraKit` and some other packages, some of which have GenericLinearAlgebra.jl as a dependency. Then also the package extension is loaded without explicit consent I assume?

lkdvos:
I addressed the remainder of the comments since I was already busy anyways.

Considering the package dependency being loaded, I think your assumption is true, if anyone has this as a dependency somewhere, the code will be available and the default algorithms will be registered.
While I would in time also prefer to have our own native implementation as the default, I don't see too much harm in registering these default algorithms either.
We can still swap them afterwards?

What is interesting though is that GenericLinearAlgebra is definitely doing quite strong type piracy on the regular LinearAlgebra methods, and is actually simply hijacking them.
One alternative strategy therefore could be to, instead of extending GenericLinearAlgebra, simply extending LinearAlgebra itself, and using that as a fallback.
(IIRC, `GenericLinearAlgebra.svd` is actually just `LinearAlgebra.svd` imported).
There could be an argument about that being fair as a fallback too...

In any case, I don't think changing this in upcoming releases really matters too much, and even if this counts as a breaking change, it shouldn't really trickle too far downstream anyways.

Jutho:
Ok, I think this is ready so I will merge.

PR #90, opened 2025-10-31
WIP: native algorithms

Jutho:
This is currently just a proof of principle, but I think it is not too hard, using my old attempt at a generic linear algebra library (never published) to implement native algorithms. Currently only has a QR (which is easy), but with a performance that is surprisingly identical to GenericLinearAlgebra. 

Hence, this is an alternative to #87 , but we can easily have both, as it would take some time to bring everything up to date.

### DIFF
--- a/test/lq.jl
+++ b/test/lq.jl
@@ -52,7 +52,7 @@ for T in (BLASFloats..., GenericFloats...), n in (37, m, 63)
             TestSuite.test_lq_algs(T, (m, n), LAPACK_LQ_ALGS)
         elseif T ∈ GenericFloats
             TestSuite.test_lq(T, (m, n); test_pivoted = false, test_blocksize = false)
-            GENERIC_LQ_ALGS = (Native_HouseholderLQ(), GLA_HouseholderLQ())
+            GENERIC_LQ_ALGS = (Native_HouseholderLQ(), LQViaTransposedQR(GLA_HouseholderQR()))
             TestSuite.test_lq_algs(T, (m, n), GENERIC_LQ_ALGS)
         end
         if m == n


kshyatt:
I think having the native AD can be helpful for testing since the compilers in both Mooncake and Enzyme are pretty good and the guidance is to only write a custom rule if you *really* have to -- such as when you have a call to a foreign library like BLAS.

Jutho:
This now contains a fully functional QR and LQ and tests, and could in principle be merged as is, with other factorizations be done in separate PRs. However, going forward, this raises a number of interesting questions: @lkdvos and @kshyatt 

1) Currently, the `Native_HouseholderQR/LQ` algorithm is not registered as default. Should this be the default for `AbstractMatrix{BigFloat}` and `AbstractMatrix{Complex{BigFloat}}`, with then other scalar types (e.g. https://github.com/JuliaMath/DoubleFloats.jl) needing separate registration (possibly in package extensions). Or do we register the native algorithms as a default for all `AbstractMatrix`, hoping that the LAPACK / GPU stuff is always correctly selected by their more specific default registration signature. Also, how should a non-strided `AbstractMatrix{Float64}` be handled in a non-mutating method so that it is anyway copied?

2) In principle, we can natively AD through these implementations. Is this interesting for testing purposes? Or do we even want to have the native AD as the default behavior?

lkdvos:
For 1, I think that since the arbitrary element types are the primary candidates for these native implementations, it would make sense to just register the native implementations as the defaults. I would prefer to avoid having to define overloads for every possible combination of weird element type and weird array type, and this seems like a solution that might often just work?

For 2, I don't really know. We don't really have a good interface for selecting different AD modes right now, (nor for selecting tolerances etc), so while I don't mind having the option to use native AD, I would also just want to see what the performances are like before we invest into trying to make this accessible?

kshyatt:
Since this provides LQ and QR could we also test the `left_orth` and `left_null` and right sided versions using these?

kshyatt:
I think like the testsuite stuff, we can merge things piecemeal as we go

test/qr.jl
```diff
@@ -51,8 +51,8 @@ for T in (BLASFloats..., GenericFloats...), n in (37, m, 63)
             TestSuite.test_qr_algs(T, (m, n), LAPACK_QR_ALGS)
         elseif T ∈ GenericFloats
             TestSuite.test_qr(T, (m, n); test_pivoted = false, test_blocksize = false)
-            GLA_QR_ALGS = (GLA_HouseholderQR(),)
-            TestSuite.test_qr_algs(T, (m, n), GLA_QR_ALGS)
+            NATIVE_QR_ALGS = (Native_HouseholderQR(),)
+            TestSuite.test_qr_algs(T, (m, n), NATIVE_QR_ALGS)
```
Jutho: Same here

Jutho:
I cannot approve my own PR but I nonetheless approve!

Issue #89, opened 2025-10-30
[Bug] SVD pullback does not properly take into account the norm of the initial matrix


leburgel:
When running variational optimization using a cost function which uses `MatrixAlgebraKit.svd_trunc` in an intermediate step, and whose value is independent of the normalization of the input, the current choice of tolerances for the `svd_trunc` pullback can give rise to very inaccurate gradients depending on the initial normalization. This came up for example in QuantumKitHub/PEPSKit.jl#276, but the problem persists with the new implementations in MatrixAlgebraKit.jl.

An example would be

```julia
using Random
using LinearAlgebra
using MatrixAlgebraKit
using OptimKit
using Zygote

Random.seed!(123)

# some random matrix
O = randn(ComplexF64, 10, 10)

# some cost function which uses svd_trunc
function fg(t)
    e, g = Zygote.withgradient(t) do x
        _, S, _ = svd_trunc(x; trunc = truncrank(10))
        return real(tr(O * (S / norm(S))))
    end
    return e, only(g)
end

# prepare a random input
t0 = randn(ComplexF64, 20, 20)
# but artificially make singular values decay exponentially
U, S, V = svd_full(t0)
S´ = S .* diagm(logrange(1.0e0, 1.0e-12, size(S, 1)))
t = U * S´ * V
```

Clearly, the value of the cost function is independent of the initial norm of `t`. Checking the gradient without any prior normalization on `t` gives a good result,
```julia
@show fg(t´)[1]
_, _, dfs1, dfs2 = OptimKit.optimtest(
    fg, t´;
    alpha = LinRange(-1.0e-5, 1.0e-5, 2),
)
@show dfs1
@show dfs2
```
```
(fg(t´))[1] = -0.621052432092075
dfs1 = [0.06446034588325489]
dfs2 = [0.06446034587777202]
```

Running the same for an input with unit norm already gives a significantly less accurate result,
```julia
rescaling_factor = norm(t´)
@show fg(t´ / rescaling_factor)[1]
_, _, dfs1, dfs2 = OptimKit.optimtest(
    fg, t´ / rescaling_factor;
    alpha = LinRange(-1.0e-5, 1.0e-5, 2),
)
@show dfs1
@show dfs2
```
```
(fg(t´ / rescaling_factor))[1] = -0.6210524320920751
dfs1 = [4.3408483775597695]
dfs2 = [4.909645149555295]
```

[...]

PR #92, opened 2025-11-05
update pullback tolerances

lkdvos:
Attempt to address #89 

I've updated the tolerances across all pullbacks to follow this logic:
- `degeneracy` and `rank` are determined by default relative to the norm
- `gauge` are determined relative to 1

Or, in other words, gauge tolerances are absolute while degeneracy and rank tolerances are relative.
I would have really liked to have a better relative criterion for the gauge tolerance as well, but can't really come up with a satisfactory solution.
I'm guessing that the idea is that the cotangent in the direction of the gauge should be small compared to the total norm of the cotangent (is that even true?), so possibly we could consider an approach like that, but already this should be an improvement.

### DIFF
--- a/src/common/defaults.jl
+++ b/src/common/defaults.jl
@@ -16,7 +16,10 @@ Default tolerance for deciding to warn if incoming adjoints of a pullback rule
 has components that are not gauge-invariant.
 """
 default_pullback_gauge_atol(A) = iszerotangent(A) ? 0 : eps(norm(A, Inf))^(3 / 4)
-default_pullback_gauge_atol(A, As...) = maximum(default_pullback_gauge_atol, (A, As...))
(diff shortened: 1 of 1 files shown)


Jutho:
I do think it makes sense if `gauge_tol` would be relative to the norm of the associated pullback variable that can have gauge-dependent components, i.e. `ΔQ` in the QR/LQ, `ΔV` in eig/eigh and something like `max(norm(ΔU, Inf), norm(ΔVᴴ))` for svd.

lkdvos:
I tried again, I think this should now be type stable on the condition that `iszerotangent` constant folds for the given types, which I think should be fine to assume.

Jutho:
`@code_warntype` seems to be happy with the new version.

Issue #91, opened 2025-11-03
[Feature] Test suite


lkdvos:
Given the current explosion of code duplication in the tests, it would be nice to streamline some of this, presumably in the form of a reusable testsuite.
A good starting point might be (suggested by @kshyatt) to check [the GPUArrays.jl testsuite](https://github.com/JuliaGPU/GPUArrays.jl/tree/master/test/testsuite).

PR #93, opened 2025-11-06
[WIP] Testsuite

lkdvos:
This is an attempt to tackle #91 .

The main goal is to provide some way to reduce the amount of code duplication in our tests here, but additionally to think about how this might affect downstream adopters of this package.

I've set up the QR tests to give an idea of what I have in mind, feedback greatly appreciated!

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -46,6 +46,7 @@ AMDGPU = "21141c5a-9bdb-4563-92ae-f87d6854732e"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
 CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"
+GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
 JET = "c3a54625-cd67-489e-a8e7-0a5a0ff4e31b"
 Mooncake = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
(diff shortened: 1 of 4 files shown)


kshyatt:
I think this should be relatively easy to make work with the GPU code, but we might want to break up some of the files into multiple functions that implement subsets of the tests to avoid scalar indexing/unsupported operations in some cases.

test/testsuite/TestSuite.jl
```diff
@@ -0,0 +1,69 @@
...
+precision(::Type{T}) where {T} = precision(eltype(T))
+
+function has_positive_diagonal(A)
+    T = eltype(A)
+    return if T <: Real
+        all(≥(zero(T)), diagview(A))
+    else
+        all(≥(zero(real(T))), real(diagview(A))) &&
+            all(≈(zero(real(T))), imag(diagview(A)))
+    end
```
Jutho:
```suggestion
    all(x -> x ≈ abs(x), diagview(A))
```

lkdvos: We can do this, but it does change the tests, `-eps()` would have this property but it wouldn't have been properly "positive". `sign(x) == 1` might work but I'm not sure how accurate that is

Jutho: Well indeed, I also think we currently require strict positivity, but I was confused by `all(≈(zero(real(T))), imag(diagview(A)))`. Is this equivalent to `all(iszero, imag(diagview(A)))`. If you want the latter, then clearly we want `x == abs(x)`.

Jutho: or simply `all(isposdef, diagview(A))` (but this requires LinearAlgebra.isposdef)

kshyatt:
I think the only factorization left to cover is `schur`, and then of course the AD tests

PR #94, opened 2025-11-12
Exponential

sanderdemeyer:
This implements the exponential of a matrix for both `BLASFloats` and `BigFloats`.

I have named these functions `exponential` and `exponential!`, instead of the usual `exp` and `exp!` from `LinearAlgebra`. Extending these methods while keeping the current structure using @algdef and @ functiondef results in some naming conflicts. The default for BLASFloats is to use `LinearAlgebra.exp!`. In `TensorKit`, we can still stick to the `exp` naming convention.


src/implementations/exponential.jl
```diff
@@ -0,0 +1,50 @@
...
+
+# Implementation
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaLA)
+    copyto!(expA, LinearAlgebra.exp(A))
+    return expA
+end
+
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaEigh)
+    D, V = eigh_full(A, alg.eigh_alg)
```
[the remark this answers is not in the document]
lkdvos: (recommend doing this in a separate PR, easier to get merged 😉 )

src/interface/decompositions.jl
```diff
@@ -319,6 +319,47 @@ Divide and Conquer algorithm.
...
 const ROCSOLVER_SVDAlgorithm = Union{ROCSOLVER_QRIteration, ROCSOLVER_Jacobi}
 
+# ================================
+# EXPONENTIAL ALGORITHMS
+# ================================
+"""
+    ExponentialViaLA()
+
+Algorithm type to denote finding the LQ decomposition of `A` by computing the QR decomposition of `Aᵀ`.
+The `qr_alg` specifies which QR-decomposition implementation to use.
```
lkdvos: todo

src/interface/decompositions.jl
```diff
@@ -319,6 +319,47 @@ Divide and Conquer algorithm.
...
+"""
+@algdef ExponentialViaLA
+
+"""
+    ExponentialViaEigh()
+
+Algorithm type to denote finding the LQ decomposition of `A` by computing the QR decomposition of `Aᵀ`.
+The `qr_alg` specifies which QR-decomposition implementation to use.
```
lkdvos: todo

src/interface/decompositions.jl
```diff
@@ -319,6 +319,47 @@ Divide and Conquer algorithm.
...
+    print(io, "ExponentialViaEigh(")
+    _show_alg(io, alg.eigh_alg)
+    return print(io, ")")
+end
+
+"""
+    ExponentialViaEig()
+
+Algorithm type to denote finding the LQ decomposition of `A` by computing the QR decomposition of `Aᵀ`.
+The `qr_alg` specifies which QR-decomposition implementation to use.
```
lkdvos: todo

src/implementations/exponential.jl
```diff
@@ -0,0 +1,50 @@
...
+    D, V = eig_full(A, alg.eig_alg)
+    copyto!(expA, V * Diagonal(exp.(diagview(D))) * inv(V))
+    return expA
+end
+
+# Diagonal logic
+# --------------
+function exponential!(A::Diagonal, expA, alg::DiagonalAlgorithm)
+    check_input(exponential!, A, expA, alg)
+    copyto!(expA, Diagonal(LinearAlgebra.exp.(diagview(A))))
```
Jutho: The `exp.` call creates a new vector which is then copied into `expA`.
```suggestion
    map!(exp, diagview(expA), diagview(A))
```

src/interface/decompositions.jl
Jutho: Something is off with the docstrings.

src/implementations/exponential.jl
```diff
@@ -0,0 +1,50 @@
...
+# Implementation
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaLA)
+    copyto!(expA, LinearAlgebra.exp(A))
+    return expA
+end
+
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaEigh)
+    D, V = eigh_full(A, alg.eigh_alg)
+    copyto!(expA, V * Diagonal(exp.(diagview(D))) * inv(V))
```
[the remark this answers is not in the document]
sanderdemeyer: It has to be `map!(exp, diagview(D), diagview(D))` instead of ` map!(exp, diagview(D))`, but good suggestion otherwise. I have also added it for the `ExponentialViaEig`.

src/implementations/exponential.jl
```diff
@@ -0,0 +1,66 @@
...
+# Implementation
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaLA)
+    copyto!(expA, LinearAlgebra.exp(A))
+    return expA
+end
+
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaEigh)
+    D, V = eigh_full(A, alg.eigh_alg)
+    iV = inv(V)
```
lkdvos:
```suggestion
    iV = adjoint(V)
```
for the hermitian case this shouldn't be a full inverse, but you then do have to adapt the remainder of the implementation to avoid overwriting it

src/implementations/exponential.jl
```diff
@@ -0,0 +1,66 @@
...
+    return expA
+end
+
+function initialize_output(::typeof(exponential!), A::Diagonal, ::DiagonalAlgorithm)
+    return similar(A)
+end
+
+# Implementation
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::ExponentialViaLA)
```
lkdvos: This implementation is a bit strange, LinearAlgebra actually uses `exp!(A)` as its expert driver, so at least this is allowed to destroy the input matrix, but additionally you should also try and avoid allocating `expA`, and simply have `expA === A`

lkdvos (commented):
Overall I'm not fully convinced by the interface of `exponential(!)`, especially in its current form and implementation this looks slightly strange.

LinearAlgebra uses an in-place version, i.e. it reuses the input array to return `exp!`, and looking at the different implementations you have here, it is not obvious that trying to fit this into a `exponentiate!(A, expA, alg)` signature is really helping us - on the contrary, all this is really doing is creating an additional copy at the end just to make sure that it is allocated in the provided output.
As we discussed for your previous PR, this really is not the purpose of being able to provide the output argument.

For the algorithms, thinking a bit ahead, it might be appropriate to just call these something along the lines of matrix functions via eig, since presumably these approaches are actually generic for all of these implementations.

src/interface/decompositions.jl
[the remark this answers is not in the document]
Jutho: I agree with Lukas, in that I have always had a bit of mixed feelings about the interface vs implementations split, and I had something like a `decompositions` folder in the very early commits (pre v0.1) of this package:
https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/commit/5702d55cd5707fc2d0aab9e0a5f4840eb2e37af8

On the other hand, I have gotten used to it by now, so a `matrixfunctions.jl` file would definitely also be ok for me.

lkdvos:
To comment on the TensorKit interaction, I definitely agree with the purpose, but this is not actually currently the design we ended up with.
Our interface very clearly states that we _may_ use the provided output, but are in no way bound to do so.
This is reflected in the TensorKit implementation, see e.g. https://github.com/QuantumKitHub/TensorKit.jl/blob/feab2072b2909b9a127d871ef58c36e83623c2fc/src/factorizations/matrixalgebrakit.jl#L39
I'm of course happy to rediscuss that design decision, but I'm actually quite happy with how that is shaping out.

So basically there are two comments I have:

On the one hand, there is the question about whether or not there are implementations that benefit from providing an additional output array.
From looking at the implementations here, this is not obvious at all, actually it seems like instead we are simply allocating an additional array at the end to copy the results into, which we might as well have reused the `A` for, since we already "destroyed" that data anyways.
I am not saying that the implementations have to be fully allocation-free (the LAPACK routines still have workspaces too), I just think that if providing the output is a way to avoid having to do an allocation, it really shouldn't be _increasing_ the allocations instead.
If we can't come up with a case where having this additional array actually helps us out, it might be reasonable to simplify define `exponentiate!(A, alg) -> expA` where `expA === A` instead, which would still work for TensorKit in the exact same way.
I definitely agree that this can be achieved by hacking it into the current interface as well, by simply having `initialize_output(exponentiate!, A) = A`, I just wanted to open up the discussion about whether or not that is something we like, or think this is just boilerplate we aren't using.
Giving it some thought, I'm definitely okay with the `initialize_output(...) = A` default and going from there, i.e. keeping `@functiondef` and "hacking" this into that.

On the other hand, given that interface, if there is no way of naturally making the output end up in the provided destination, I would really like to avoid ending up with a final `copy!(provided_dest, computed_exp)` at the end, even as a "for now" implementation that we can improve on later.
Given that we have to check for in-placeness in TensorKit anyways, I would in this case much rather have `initialize_output() = nothing`, and simply return the `computed_exp` instead.
This is the exact same comment and solution as for the generic linear algebra wrappers.

sanderdemeyer:
The last few commits added the functions `exponentiali` and `exponentiali!`, for which it holds that `exponentiali(t, A) = exponential(im*t*A)`. This can be useful for hermitian matrices A, where you still want to use the hermitian eigendecomposition for the matrix A, even though the matrix `im*t*A` is not per se hermitian.

Jutho:
I think I finally managed to work away (some of) my backlog of PRs that I had to review, so this one is among the next on my list. I hope to go through it (and if necessary commit changes myself directly) before the end of the week.

src/implementations/exponential.jl
```diff
@@ -0,0 +1,136 @@
...
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
+    @check_size(expA, (m, m))
+    return @check_scalar(expA, A)
+end
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    if !ishermitian(A)
+        throw(DomainError(A, "Hermitian matrix was expected. Use `project_hermitian` to project onto the nearest hermitian matrix)"))
+    end
```
lkdvos: TODO: do we want to keep this check here, knowing that it will again be checked in the iimplementation of `eigh_full`?

lkdvos (commented):
I cleaned up part of the implementation, with two major changes:
- `expA` is only allocated if a scalartype change is required
- I refactored/added `map_diagonal(!)` in an attempt to have a specialization hook, thinking ahead a bit about extending this to different types and different matrix functions

I feel like with this change, the specific difference for `exponentiali` feels even smaller.
I would still argue that it is a bit strange to default to an imaginary scalar while not supporting a real one, and instead would argue that it could be more convenient to have a generic scalar for which you could simply fill in an imaginary value.
Thinking about that a bit more, I even would say that the reason we need to have the `imaginary_evolution` flag in MPSKit for time evolution is precisely because we defaulted to evolving with `exp(im * t * H)` instead of `exp(t * H)`.
It is easy to detect real vs complex scalars in a type stable manner, but less so to detect purely imaginary vs complex.
(I hope I'm explaining this somewhat decently)

Trying to come up with some solutions I think could be better, I can think of some alternative approaches:

- We could introduce a `ScaledOperator(A, tau)` dedicated type, and then (re)use `exponential(ScaledOperator(A, tau), ...)`. This might remove a little bit of overhead, and a default implementation could just instantiate the scaled operator. This avoids the need for `@functiondef 2 ...`
- We could simply always have a scale factor in the `exponential` function, which by default takes the value `true` or `VectorInterface.One()`, and just correctly handle the complex/real case by checking if this scalar type is real/complex.

---

Some more discussion points/things to consider, some I recall briefly bringing up but might be good to write down here:

These matrix functions should have the same output types, no matter whether they are `ishermitian` or not.
Instead, most matrix functions distinguish between having support for the entire `\bbR -> \bbR`, or require complex extensions to do so (this is of course not relevant for `exp`).
Therefore, it might actually be reasonable to instead of splitting `MatrixFunctionViaEig` and `MatrixFunctionViaEigh` into separate types, simply having a `MatrixFunctionViaEigen` with a runtime flag for hermitianness, rather than a compile time flag. (in essence, manually union-splitting with an `if` statement) This could potentially cut down on compile-times, and the number of required algorithms?

It might be reasonable to have a global `LinearAlgebraAlgorithm` backend, which functions similar to `DiagonalAlgorithm` in the sense that it simply dispatches to the respective LinearAlgebra routines.
With this, I think we could deprecate most of the GenericLinearAlgebra extension, as this would simply coincide.

src/implementations/exponential.jl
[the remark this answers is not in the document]
lkdvos: Closing this since I just went and checked myself, this syntax already exists, just needed to add a `!` there.

lkdvos:
> I don't get why you say the current code would not work with a real scalar.

The difference is a bit subtle, but as I mentioned before this is about type stability.
(It is easy to detect real vs complex scalars in a type stable manner, but less so to detect purely imaginary vs complex.)
Maybe it is more obvious with an example:

```julia
function exponentialr(tau, A)
    expA = tau isa Complex ? complex(A) : A
    return exponentialr!(tau, A, expA) # this computes expA .= exp(A * tau) in whatever way
end
function exponentiali(t, A)
    expA = iszero(real(t)) ? A : complex(A)
    return exponentiali!(t, A, expA) # this computes expA .= exp(A * t * im) in whatever way
end
```
This first implementation is type stable for both purely real, purely imaginary and complex values of tau, while the second isn't.

---

> In terms of differentiating between eigh and eig during runtime: Then why do we not do this for eigh and eig themselves?

The reason we opted for `eigh` and `eig` is that, unlike `LinearAlgebra.eigen`, this can be implemented in a type-stable manner for real input matrices. `eigh` will return real `D` and `V`, while `eig` will always return complex `D` and `V`.

For `exponential` we can revisit this, since both hermitian and non-hermitian real inputs get real outputs.

---

> (again, because I would like to have this in the next TensorKit version)

I’m afraid that might be difficult: we’re trying to get a version released this week, so I don’t think integrating and polishing this in time is very realistic.

That said, if your main goal is to get something working for your use case in the next TensorKit release, you could already move ahead with something along the following lines:

```julia
function myexp(A::AbstractTensorMap)
    D, V = eig_full(A)
    D.data .= exp.(D.data)
    return V * D * inv(V)
end
```

From my side, I think it’s important that we take a bit of time here to come up with a good, consistent interface and to think through allocations and long-term maintenance, especially since this will likely influence all of the matrix functions. I’d really like to avoid having to introduce multiple breaking changes later.

But that shouldn’t block you from experimenting or using a simpler version for your immediate needs.

sanderdemeyer:
Revisiting this after quite some time. I think the last main issue was about the type stability of the different methods. Indeed, we can rewrite `exponential` in a type-stable manner, as shown by @lkdvos above. Since this is not possible for `exponentiali` (or whatever name we will give the function that computes `exp(im*tau*A))`, I guess we have two options

- We remove `exponentiali`. 
- We define `exponentiali` in such a way that it always returns a complex matrix, even in the case where tau is purely imaginary. This means we could still use `eigh` to compute the exponential of skew-hermitian matrices. It's indeed suboptimal when tau purely imaginary, but maybe then you should use `exponential(im*tau, A)` anyway.

Let me know which of the two you prefer, or whether there is another option that I'm missing here. Additionally, should there be any changes to this PR in light of #QuantumKitHub/TensorKit.jl#342?

Jutho:
I have been reading a bit about matrix exponentials recently, and the conclusion seems to be that, for higher precision, you basically want to evaluate it as a Taylor series (potentially after some rescaling, followed by squaring the result, i.e.

exp(A) = (exp_via_taylor_up_to_order_m(A/2^n))^(2^n)

where order m and scaling exponent n is chosen according to some cost function. I hope to implement this soon.

lkdvos:
I think you left out the third option here, which is to add an `exponentialr` where you give an imaginary factor, which would solve the issues of type stability while keeping the features.
However, in the interest of keeping things moving, it probably makes sense to just stick with `exponential` for now, and deal with the inefficiency for now, especially since I think this does not apply as much to @Jutho's Taylor suggestion?

sanderdemeyer:
@lkdvos I guess this is equivalent to supporting `exponential(A)` and `exponential(tau, A)`, where we dispatch on whether `tau` is real or not (maybe this is what you meant?). Then we would not have to support `exponentiali`, and everything should be type stable. Is that an approach you would be okay with?

lkdvos:
yes, this is exactly what I meant! :)

sanderdemeyer:
I finally got around to working on this. Some comments

- Instead of having `exponentiali` encoding `exp(im*tau*A)`, we now have `exponentialr` encoding `exp(tau*A)`. I don't know whether we want to keep this name, but changing it to just `exponential` is not possible withing the current framework of `n_args`.
- I added the cases where the eltype of `tau` and `A` can be different to the tests.
- One previous commit trying to avoid an allocation gave incorrect results, so I changed this back (see below). I don't know if there is a way to avoid this.

src/implementations/exponential.jl
```diff
@@ -95,32 +96,36 @@ function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFuncti
...
-        VexpD = V * expD
+    expD = map_diagonal(x -> exp(x * τ), D)
+    VexpD = V * expD
+    if eltype(A) <: Real && eltype(τ) <: Real
         return expA .= real.(VexpD * V')
     else
-        VexpD = rmul!(V, expD)
         return mul!(expA, VexpD, V')
     end
 end
```
sanderdemeyer: In the complex case, `V` was overwritten, such that `V'` was not the 'original' `V'`, but also contained the exponential. This could be resolved where `tau` is real, where we could absorb a factor `exp(tau * D / 2)`. Is this something that would be preferred? Then we would have a case where `tau` is complex and a case where `tau` is real, where the latter will be further divided based on the eltype of A.

ext/MatrixAlgebraKitGenericSchurExt.jl
```diff
@@ -14,6 +14,13 @@ function MatrixAlgebraKit.default_eig_algorithm(
     return QRIteration(; driver, kwargs...)
 end
 
+function MatrixAlgebraKit.default_exponential_algorithm(
+        type::Type{T}; driver::Driver = GS(), kwargs...
+    ) where {T <: StridedMatrix{<:GSFloat}}
+    eig_alg = MatrixAlgebraKit.default_eig_algorithm(type; driver, kwargs...)
```
lkdvos:
```suggestion
        type::Type{T}; kwargs...
    ) where {T <: StridedMatrix{<:GSFloat}}
    eig_alg = MatrixAlgebraKit.default_eig_algorithm(type; kwargs...)
```
I think it might be fine to just keep the driver unspecified, since `default_eig_algorithm` should select the appropriate default, and then this is one fewer spot where this has to be kept into account.

src/implementations/exponential.jl
```diff
@@ -0,0 +1,141 @@
...
+initialize_output(::typeof(exponential!), A::AbstractMatrix, ::AbstractAlgorithm) = A
+initialize_output(::typeof(exponentialr!), τ::T, A::AbstractMatrix, ::AbstractAlgorithm) where {T <: Real} = A
+initialize_output(::typeof(exponentialr!), τ::Number, A::AbstractMatrix, ::AbstractAlgorithm) =
+    complex(A)
+
+# Implementation
+# --------------
+function exponential!(A, expA, alg::MatrixFunctionViaLA)
+    check_input(exponential!, A, expA, alg)
+    return LinearAlgebra.exp!(A)
```
lkdvos: I know we had a lot of discussions about this in the past, but I think we now settled on actually copying over the data (see also the GLA extension, I think), so this should probably become:

```suggestion
    LinearAlgebra.exp!(A)
    A === expA || copy!(expA, A)
    return expA
```

test/genericlinearalgebra/exponential.jl
```diff
@@ -0,0 +1,48 @@
...
+using GenericLinearAlgebra
+
+GenericFloats = (BigFloat, Complex{BigFloat})
+
+@testset "exponential! for T = $T" for T in GenericFloats
+    rng = StableRNG(123)
+    m = 54
+
+    A = randn(rng, T, m, m)
+    A = (A + A') / 2
```
lkdvos:
```suggestion
    A = project_hermitian!(randn(rng, T, m, m))
```

test/genericlinearalgebra/exponential.jl
```diff
@@ -0,0 +1,48 @@
...
+        @test diagview(Dexp) ≈ LinearAlgebra.exp.(diagview(D))
+    end
+end
+
+using GenericSchur
+@testset "exponentialr! for T1 = $T1, T2 = $T2" for T1 in GenericFloats, T2 in GenericFloats
+    rng = StableRNG(123)
+    m = 54
+    A = randn(rng, T1, m, m)
+    A = (A + A') / 2
```
lkdvos:
```suggestion
    A = project_hermitian!(randn(rng, T1, m, m))
```

lkdvos (commented):
Overall this is looking quite nice! Apologies if this is a premature review, just showed up in my notifications and figured I'd have a look.

The only thing I'm still wondering is if we want to just replace `exponentialr(t, A) = exponential((t, A))` and not have the two functions. I know we had this discussion before, I'm not sure what came out of that to be honest.

src/implementations/exponential.jl
sanderdemeyer: > I know we had a lot of discussions about this in the past, but I think we now settled on actually copying over the data (see also the GLA extension, I think), so this should probably become:

This does not work, for the very bizarre reason that `LinearAlgebra.exp!(A)` does not guarantee that the result is directly stored in the matrix `A` (at least when I test it locally with version 1.12.0). Everything is fine if I replace line 74 with `A = LinearAlgebra.exp!(A)`, but probably there is a better way to get around this?

sanderdemeyer:
> The only thing I'm still wondering is if we want to just replace `exponentialr(t, A) = exponential((t, A))` and not have the two functions. I know we had this discussion before, I'm not sure what came out of that to be honest.

I also don't know whether there was a consensus. I would be in favour of indeed just using `exponential((t,A))`, since that seems to be the most convenient/logical option. Especially if we don't support `exponentiali`, it feels weird to call it 'exponentialr'. So if you agree with this, I will make the necessary changes.

lkdvos:
Yes, let's do this and then push this PR over the edge 🎉

src/implementations/exponential.jl
```diff
@@ -0,0 +1,124 @@
...
+copy_input(::typeof(exponential), (τ, A)::Tuple{Number, AbstractMatrix}) = (τ, copy!(similar(A, float(eltype(A))), A))
+copy_input(::typeof(exponential), (τ, A)::Tuple{Number, Diagonal}) = τ, copy(A)
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::AbstractAlgorithm; scalar_check = true)
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
+    @check_size(expA, (m, m))
+    if scalar_check
+        @check_scalar(expA, A)
+    end
```
sanderdemeyer: I added these lines to be able to differentiate the `check_input` test between `exponential(A)` and `exponential(tau,A)`, since the latter does not require `@check_scalar`. Feel free to make suggestions to do this in a different way.

lkdvos: Why does the latter not require that? If A is real and tau is complex, expA has to be complex right?

sanderdemeyer: You're right, I forgot about that. I added a method `check_input` for `exponential((tau,A))`, where I dispatch on whether tau is real.

src/implementations/exponential.jl
```diff
@@ -8,40 +8,43 @@ copy_input(::typeof(exponential), A::Diagonal) = copy(A)
...
+    @check_scalar(expA, A)
     return nothing
 end
 
+function check_input(type::typeof(exponential!), τ::Real, A::AbstractMatrix, expA::AbstractMatrix, alg)
+    return check_input(type, A, expA, alg)
+end
+
+function check_input(type::typeof(exponential!), τ, A::AbstractMatrix, expA::AbstractMatrix, alg)
+    return check_input(type, complex(A), expA, alg)
```
lkdvos: This is probably not entirely how we should handle this, as we don't want to copy `A` to make the check. I think the easiest is just to copy some of the code, and I think `@check_scalar` has the option to give `complex` as an extra parameter. See e.g. the `eig_` implementations, where we have a similar thing happening

src/implementations/exponential.jl
```diff
@@ -0,0 +1,127 @@
...
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
+    @check_size(expA, (m, m))
+    @check_scalar(expA, A)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    if !ishermitian(A)
+        throw(DomainError(A, "Hermitian matrix was expected. Use `project_hermitian` to project onto the nearest hermitian matrix)"))
+    end
```
lkdvos:
```suggestion
```

sanderdemeyer: Why do you want to remove this? If you're calculating the eigenvalue decomposition via `eigh`, you still require `A` to be Hermitian, right?

src/implementations/exponential.jl
sanderdemeyer: Now, I copy some of the code, and changed `@check_scalar(expA, A)` to `(τ isa Real) ? @check_scalar(expA, A) : @check_scalar(expA, A, complex)`

src/implementations/exponential.jl
lkdvos: Yes, but `eigh` will check this itself, so this check is duplicated. (Additionally, otherwise we'd have to repeat the logic about the tolerance, since this now only works for exactly hermitian inputs, which typically is too strict.

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    return nothing
+end
+
+function check_input(::typeof(exponential!), τ::Number, A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
+    @check_size(expA, (m, m))
+    (τ isa Real) ? @check_scalar(expA, A) : @check_scalar(expA, A, complex)
+    return nothing
+end
```
lkdvos:
```suggestion
```

I think this is covered automatically by the case above right, so no need for an extra specialization? (Just looking at this, I might be missing some ambiguity)

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    return copy!(similar(A, float(eltype(A))), A)
+end
+
+copy_input(::typeof(exponential), A::Diagonal) = copy(A)
+copy_input(::typeof(exponential), (τ, A)::Tuple{Number, AbstractMatrix}) = (τ, copy!(similar(A, float(eltype(A))), A))
+copy_input(::typeof(exponential), (τ, A)::Tuple{Number, Diagonal}) = τ, copy(A)
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::AbstractAlgorithm)
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
```
lkdvos:
```suggestion
    m = LinearAlgebra.checksquare(A)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    @check_size(expA, (m, m))
+    @check_scalar(expA, A)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
```
lkdvos:
```suggestion
    m = LinearAlgebra.checksquare(A)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    @check_size(expA, (m, m))
+    @check_scalar(expA, A)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, ::DiagonalAlgorithm)
+    m, n = size(A)
+    @assert m == n && isdiag(A)
```
lkdvos:
```suggestion
    m = LinearAlgebra.checksquare(A)
    @assert isdiag(A)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    @assert expA isa Diagonal
+    @check_size(expA, (m, m))
+    @check_scalar(expA, A)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), τ::Number, A::AbstractMatrix, expA::AbstractMatrix, alg::AbstractAlgorithm)
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
```
lkdvos:
```suggestion
    m = LinearAlgebra.checksquare(A)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    @check_size(expA, (m, m))
+    (τ isa Real) ? @check_scalar(expA, A) : @check_scalar(expA, A, complex)
```
lkdvos:
```suggestion
    @check_scalar(expA, A, (τ isa Real) ? identity : complex)
```

Not sure if this is more readable, just a suggestion to try out. Feel free to ignore

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    return nothing
+end
+
+function check_input(::typeof(exponential!), τ::Number, A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
```
lkdvos:
```suggestion
    m = LinearAlgebra.checksquare(A)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    @check_size(expA, (m, m))
+    (τ isa Real) ? @check_scalar(expA, A) : @check_scalar(expA, A, complex)
```
lkdvos:
```suggestion
    @check_scalar(expA, A, (τ isa Real) ? identity : complex)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    return nothing
+end
+
+function check_input(::typeof(exponential!), τ::Number, A::AbstractMatrix, expA::AbstractMatrix, ::DiagonalAlgorithm)
+    m, n = size(A)
+    @assert m == n && isdiag(A)
+    @assert expA isa Diagonal
+    @check_size(expA, (m, m))
+    (τ isa Real) ? @check_scalar(expA, A) : @check_scalar(expA, A, complex)
```
lkdvos:
```suggestion
    @check_scalar(expA, A, (τ isa Real) ? identity : complex)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+    m, n = size(A)
+    m == n || throw(DimensionMismatch("square input matrix expected. Got ($m,$n)"))
+    @check_size(expA, (m, m))
+    (τ isa Real) ? @check_scalar(expA, A) : @check_scalar(expA, A, complex)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), τ::Number, A::AbstractMatrix, expA::AbstractMatrix, ::DiagonalAlgorithm)
+    m, n = size(A)
+    @assert m == n && isdiag(A)
```
lkdvos:
```suggestion
    m = LinearAlgebra.checksquare(A)
    @assert isdiag(A)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,140 @@
...
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, ::DiagonalAlgorithm)
+    m, n = size(A)
+    @assert m == n && isdiag(A)
+    @assert expA isa Diagonal
+    @check_size(expA, (m, m))
+    @check_scalar(expA, A)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), τ::Number, A::AbstractMatrix, expA::AbstractMatrix, alg::AbstractAlgorithm)
```
lkdvos: Another question that is kind of irrelevant, but just want to check if you have a specific reasoning for choosing `check_input(exponential!, (τ, A), expA, alg)` vs `check_input(exponential!, τ, A, expA, alg)`. I guess there is some annoyance with the tuple and dispatch, but this is somewhat less in line with the other single-argument functions where `f!(in, out, alg)` calls `check_input(f!, in, out, alg)`, instead of here `check_input(f!, in..., out, alg)`. I assume this is a remnant of using the 2-arg `@functiondef` from before?

lkdvos (commented):
I think I left a final set of comments, otherwise this is good to go for me. You'll see it's quite nitpicky, so feel free to share your opinions as well!

src/implementations/exponential.jl
sanderdemeyer: I agree that `check_input(exponential!, (τ, A), expA, alg)` is the more logical choice. I've changed that now.

sanderdemeyer:
> I think I left a final set of comments, otherwise this is good to go for me. You'll see it's quite nitpicky, so feel free to share your opinions as well!

Thanks for the final comments. I agreed with all of them!

lkdvos (approved):
Thanks for all the work @sanderdemeyer, this looks like a great addition!

@Jutho, @kshyatt I'll leave this open in case you still want to have a look, and at the latest I'll merge this Monday evening to keep things moving, while not imposing on weekends 😉

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+    @check_scalar(expA, A)
+    return nothing
+end
+
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    m = LinearAlgebra.checksquare(A)
+    @check_size(expA, (m, m))
+    @check_scalar(expA, A)
+    return nothing
+end
```
Jutho: Isn't this one covered by the previous definition with `alg::AbstractAlgorithm`?

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+    A === expA || copy!(expA, A)
+    return expA
+end
+
+function exponential!(A, expA, alg::MatrixFunctionViaEigh)
+    check_input(exponential!, A, expA, alg)
+    D, V = eigh_full!(A, alg.eigh_alg)
+    expD = map_diagonal!(x -> exp(x / 2), D, D)
+    VexpD = rmul!(V, expD)
+    return mul!(expA, VexpD, V')
```
Jutho: This doesn't seem right. I think you do `x -> exp(x / 2)` with the intention to then do `mul(expA, VexpD, VexpD')` ?

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    check_input(exponential!, (τ, A), expA, alg)
+    D, V = eigh_full!(A, alg.eigh_alg)
+    expD = map_diagonal(x -> exp(x * τ), D)
+    VexpD = V * expD
+    if eltype(A) <: Real && eltype(τ) <: Real
+        return expA .= real.(VexpD * V')
+    else
+        return mul!(expA, VexpD, V')
+    end
```
Jutho: `if eltype(A) <: Real && eltype(τ) <: Real`, everything should remain real, so no `real` call is necessary.


```suggestion
    if eltype(A) <: Real && eltype(τ) <: Real
        V .*= exp.(transpose(diagview(D)) .* τ ./ 2)
        return mul!(expA, V, V')
    else
        VexpD = V .* exp.(transpose(diagview(D)) .* τ)
        return mul!(expA, VexpD, V')
    end
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+        mul!(expA, VexpD, iV)
+    end
+    return expA
+end
+
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::MatrixFunctionViaLA)
+    check_input(exponential!, (τ, A), expA, alg)
+    expA .= A .* τ
+    return LinearAlgebra.exp!(expA)
+end
```
Jutho: To avoid code duplication:

```suggestion
exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::MatrixFunctionViaLA) = exponential!(A * τ, expA, alg)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+    expD = map_diagonal!(exp, D, D)
+    iV = inv(V)
+    VexpD = rmul!(V, expD)
+    if eltype(A) <: Real
+        expA .= real.(VexpD * iV)
+    else
+        mul!(expA, VexpD, iV)
+    end
+    return expA
+end
```
Jutho:
```suggestion
exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEig) = exponential!((1, A), expA, alg)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+
+function exponential!(A, expA, alg::MatrixFunctionViaEigh)
+    check_input(exponential!, A, expA, alg)
+    D, V = eigh_full!(A, alg.eigh_alg)
+    expD = map_diagonal!(x -> exp(x / 2), D, D)
+    VexpD = rmul!(V, expD)
+    return mul!(expA, VexpD, V')
+end
```
Jutho:
```suggestion
exponential!(A, expA, alg::MatrixFunctionViaEigh) = exponential!((1, A), expA, alg)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+    D, V = eig_full!(A, alg.eig_alg)
+    expD = map_diagonal!(x -> exp(x * τ), D, D)
+    iV = inv(V)
+    VexpD = rmul!(V, expD)
+    if eltype(A) <: Real && eltype(τ) <: Real
+        expA .= real.(VexpD * iV)
+        return expA
+    else
+        return mul!(expA, VexpD, iV)
+    end
```
Jutho:
```suggestion
    if eltype(A) <: Real && eltype(τ) <: Real
        VexpD = V .* exp.(transpose(diagview(D)) .* τ)
        expAc = rdiv!(VexpD, LinearAlgebra.lu!(V))
        return expA .= real.(expAc)
    else
        expA .= V .* exp.(transpose(D) .* τ)
        return rdiv!(expA, LinearAlgebra.lu!(V))
    end
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+end
+
+# Diagonal logic
+# --------------
+function exponential!(A, expA, alg::DiagonalAlgorithm)
+    check_input(exponential!, A, expA, alg)
+    return map_diagonal!(exp, expA, A)
```
Jutho, line 103: I actually like the broadcasting version more. Together with the suggestions above, I think this removes the need for `map_diagonal`:


```suggestion
    diagview(expA) .= exp.(diagview(A))
    return expA
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,129 @@
...
+end
+
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA, alg::DiagonalAlgorithm)
+    check_input(exponential!, (τ, A), expA, alg)
+    return map_diagonal!(x -> exp(x * τ), expA, A)
```
Jutho, line 108:
```suggestion
    diagview(expA) .= exp.(diagview(A) .* τ)
    return expA
```

src/interface/exponential.jl
```diff
@@ -0,0 +1,43 @@
...
+    @eval function default_algorithm(::typeof($f), ::Type{A}; kwargs...) where {A}
+        return default_exponential_algorithm(A; kwargs...)
+    end
+end
+
+for f in (:exponential!,)
+    @eval function default_algorithm(::typeof($f), ::Tuple{A, B}; kwargs...) where {A, B}
+        return default_exponential_algorithm(B; kwargs...)
+    end
+end
```
Jutho: No need to do these via an `@eval` block since it is only one function:
```suggestion
function default_algorithm(::typeof(exponential!), ::Type{A}; kwargs...) where {A}
    return default_exponential_algorithm(A; kwargs...)
end

function default_algorithm(::typeof(exponential!), ::Tuple{A, B}; kwargs...) where {A, B}
    return default_exponential_algorithm(B; kwargs...)
end
```

src/interface/matrixfunctions.jl
```diff
@@ -0,0 +1,39 @@
...
+    MatrixFunctionViaLA()
+
+Algorithm type to denote finding the exponential of `A` via the implementation of `LinearAlgebra`.
+"""
+@algdef MatrixFunctionViaLA
+
+"""
+    MatrixFunctionViaEigh()
+
+Algorithm type to denote finding the exponential `A` by computing the hermitian eigendecomposition of `A`.
```
Jutho:
```suggestion
Algorithm type for computing a function of a matrix by computing its hermitian eigenvalue decomposition and applying the function to the eigenvalues.
```

src/interface/matrixfunctions.jl
```diff
@@ -0,0 +1,39 @@
...
+# ================================
+"""
+    MatrixFunctionViaLA()
+
+Algorithm type to denote finding the exponential of `A` via the implementation of `LinearAlgebra`.
+"""
+@algdef MatrixFunctionViaLA
+
+"""
+    MatrixFunctionViaEigh()
```
Jutho:
```suggestion
    MatrixFunctionViaEigh(eigh_alg)
```

src/interface/matrixfunctions.jl
```diff
@@ -0,0 +1,39 @@
...
+    print(io, "MatrixFunctionViaEigh(")
+    _show_alg(io, alg.eigh_alg)
+    return print(io, ")")
+end
+
+"""
+    MatrixFunctionViaEig()
+
+Algorithm type to denote finding the exponential `A` by computing the eigendecomposition of `A`.
+The `eig_alg` specifies which eigendecomposition implementation to use.
```
Jutho:
```suggestion
    MatrixFunctionViaEig(eig_alg)

Algorithm type for computing a function of a matrix by computing its eigenvalue decomposition and applying the function to the eigenvalues.
The `eig_alg` specifies which eigendecomposition implementation to use.
```

Jutho (changes requested):
I think there some more required changes that could improve allocations, code duplication and in one case correctness.

src/implementations/exponential.jl
lkdvos: We should probably rename the variables slightly, but V and expD share memory so the two are in fact the same

src/implementations/exponential.jl
lkdvos:
```suggestion
function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::AbstractAlgorithm)
    expA .= A .* τ
    return exponential!(expA, expA, alg)
end
```

I think this way we avoid one more allocation, and I immediately relaxed the `alg::AbstractAlgortihm` signature since I guess this is a good default implementation? (Note that while we could modify `A` itself, this might fail because of `eltype` issues, so it has to be `expA`.

src/implementations/exponential.jl
lkdvos:
```suggestion
    return mul!(expA, VexpD, VexpD')
```

src/implementations/exponential.jl
lkdvos: Should we do `1`, `one(eltype(A))`, or `true`?

src/implementations/exponential.jl
lkdvos: I think I suggested/introduced the `map_diagonal` to have a specialization point in case we end up using this for `SectorVector`/`DiagonalTensorMap` as well. (Currently broadcasting for those doesn't actually work on GPU, since it falls back to scalar indexing). Happy to remove this if you prefer though, but then we probably should investigate properly implementing the broadcasting :)

Jutho: I am fine with `map_diagonal!` also, and then we should probably use it more consistently. We have to make sure that this one then works well on the GPU, and supports overwriting input, e.g. like `map_diagonal!(exp, D, D)`.

But maybe we should actually just call `exponential!(D, D, DiagonalAlgorithm())`?

lkdvos: That's a great idea, that way I guess we can fully remove any array-specific part of this code and have it be generic?

src/implementations/exponential.jl
sanderdemeyer: I have a slight preference for `one(eltype(A))`.

src/implementations/exponential.jl
sanderdemeyer: This is now added in the latest commit.

src/implementations/exponential.jl
sanderdemeyer: This is now outdated due to a previous comment.

sanderdemeyer:
I think all comments are now taken care of (except for the discussion of whether to use `1` or `one(eltype(A))`, where I've chosen the latter.)

src/implementations/exponential.jl
```diff
@@ -0,0 +1,102 @@
...
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::AbstractAlgorithm)
+    expA .= A .* τ
+    return exponential!(expA, expA, alg)
+end
+
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::MatrixFunctionViaEigh)
+    check_input(exponential!, (τ, A), expA, alg)
+    D, V = eigh_full!(A, alg.eigh_alg)
+    if eltype(A) <: Real && eltype(τ) <: Real
+        V .*= transpose(diagview(exponential!((τ / 2, D), D, DiagonalAlgorithm())))
```
lkdvos:
```suggestion
        expD = exponential!((τ / 2, D), D)
        rmul!(V, expD)
```

This is slightly easier to read for me :) I think specifying the `DiagonalAlgorithm` shouldn't be required, and we can simply use the default selection mechanism here :).

src/implementations/exponential.jl
```diff
@@ -0,0 +1,102 @@
...
+        return mul!(expA, V, V')
+    else
+        VexpD = V .* exp.(transpose(diagview(D)) .* τ)
+        return mul!(expA, VexpD, V')
```
lkdvos: I was looking at this and slightly breaking my brain, if everything is complex than `D` is real (because hermitian), but we can still get away with `rmul!(V, exponential!((τ / 2, D), D))` since `V` should be complex, and then we basically recover the path from above. Only in the case where `A` is real, `tau` is complex do we run into trouble because the in-place `rmul!` has to become a new allocation due to the element type changing.
However, even in that case it's probably better to use the `tau/2` way because `mul!(expA, VexpD, VexpD')` can then use BLAS, since all matrices involved have complex entries. (I think in principle we can do `mul!(complexC, complexA, realB)`, but not sure if that is automatically implemented?)

src/implementations/exponential.jl
```diff
@@ -0,0 +1,102 @@
...
+    end
+end
+
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::MatrixFunctionViaEig)
+    check_input(exponential!, (τ, A), expA, alg)
+    D, V = eig_full!(A, alg.eig_alg)
+    if eltype(A) <: Real && eltype(τ) <: Real
+        VexpD = V .* transpose(diagview(exponential!((τ, D), D, DiagonalAlgorithm())))
```
lkdvos:
```suggestion
        VexpD = V * exponential!((τ, D), D)))
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,102 @@
...
+        expAc = rdiv!(VexpD, LinearAlgebra.lu!(V))
+        return expA .= real.(expAc)
+    else
+        expA .= V .* transpose(diagview(exponential!((τ, D), D, DiagonalAlgorithm())))
```
lkdvos:
```suggestion
        expA .= V .* transpose(diagview(exponential!((τ, D), D)))
```

lkdvos (commented):
sorry for being a bit slow on this, but I just realized that probably `exponential!((tau, D))` is enough instead of the `exponential!((tau, D), D, DiagonalAlgorithm())`, which should make a lot of the code more generic and might allow us to just have it work for the TensorMap's instead.

src/implementations/exponential.jl
Jutho: If `tau` is complex but `A` real and symmetric, we shouldn't be doing `VexpD * VexpD'` cause that gives something different, but `VexpD * transpose(VexpD)` might work.

It was to avoid these different eltype combinations that I used broadcasting, but I think the point is now to avoid broadcasting altogether, right?

lkdvos: I think by avoiding broadcasting it would work with `TensorMap` instances out of the box, but this might also really be taking it too far and it might not matter all that much for a first implementation, and we can always revisit to optimize

sanderdemeyer: Is the consensus then to keep the broadcasting here? There are a few other instances where broadcasting is used, and thus in order for `exponential(!)` to work with `TensorMap`s, there will anyway be some work needed in TensorKit.

Jutho: But now it is completely inconsistent no? In one branch of an `if else` we do call `exponential!(::Diagonal)` and in the other we do broadcasting?

I think we either want the ``::ExponentialViaEig(h)`` to either be fully written using `exponential!(::Diagonal)` and generic operations such as `mul!` and `rmul!`, so that it is directly useful for other types.

If we do not succeed, and anyway use broadcasting in part of it, I preferred my older version that just went all in with broadcasting.

sanderdemeyer: I changed it such that it avoids broadcasting altogether. To avoid allocations as much as possible, I just consider the 4 different cases now.

src/implementations/exponential.jl
```diff
@@ -0,0 +1,110 @@
...
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::DiagonalAlgorithm)
+    check_input(exponential!, A, expA, alg)
+    return map_diagonal!(exp, expA, A)
+end
+
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::DiagonalAlgorithm)
+    check_input(exponential!, (τ, A), expA, alg)
+    diagview(expA) .= exp.(diagview(A) .* τ)
+    return expA
```
lkdvos:
```suggestion
    return map_diagonal!(x -> exp(x * τ), expA, A)
```

src/interface/exponential.jl
```diff
@@ -0,0 +1,39 @@
...
+
+"""
+    exponential(A; kwargs...) -> expA
+    exponential(A, alg::AbstractAlgorithm) -> expA
+    exponential!(A, [expA]; kwargs...) -> expA
+    exponential!(A, [expA], alg::AbstractAlgorithm) -> expA
+    exponential((τ,A); kwargs...) -> expτA
+    exponential((τ,A), alg::AbstractAlgorithm) -> expτA
+    exponential!((τ,A), [expA]; kwargs...) -> expτA
+    exponential!((τ,A), [expA], alg::AbstractAlgorithm) -> expτA
```
lkdvos:
```suggestion
    exponential((τ, A); kwargs...) -> expτA
    exponential((τ, A), alg::AbstractAlgorithm) -> expτA
    exponential!((τ, A), [expA]; kwargs...) -> expτA
    exponential!((τ, A), [expA], alg::AbstractAlgorithm) -> expτA
```

src/interface/exponential.jl
```diff
@@ -0,0 +1,39 @@
...
+
+Compute the exponential of the square matrix `A` or `τ*A`,
```
lkdvos:
```suggestion
Compute the exponential of the square matrix `A` or `τ * A`,
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,110 @@
...
+end
+
+# Diagonal logic
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::DiagonalAlgorithm)
+    check_input(exponential!, A, expA, alg)
+    return map_diagonal!(exp, expA, A)
+end
+
+function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::DiagonalAlgorithm)
```
lkdvos:
```suggestion
function exponential!((τ, A)::Tuple{Number, Any}, expA, alg::DiagonalAlgorithm)
```

src/implementations/exponential.jl
```diff
@@ -0,0 +1,110 @@
...
+        return expA .= real.(expAc)
+    else
+        expA .= V .* transpose(diagview(exponential!((τ, D), D)))
+        return rdiv!(expA, LinearAlgebra.lu!(V))
+    end
+end
+
+# Diagonal logic
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::DiagonalAlgorithm)
```
lkdvos:
```suggestion
function exponential!(A, expA, alg::DiagonalAlgorithm)
```

src/implementations/exponential.jl
sanderdemeyer: Without the specification of AbstractMatrix for expA, this is ambiguous with 
```julia
function exponential!((τ, A)::Tuple{Number, AbstractMatrix}, expA::AbstractMatrix, alg::AbstractAlgorithm)
    expA .= A .* τ
    return exponential!(expA, expA, alg)
end
```
unless you also want to remove the type restriction there?

lkdvos: Oh, you are completely right, apologies that's my bad...
Maybe it is not worth it to bother with this for now and just deal with the extra boilerplate definitions in TensorKit.
Sorry for delaying this a bit more...

sanderdemeyer: I reverted this change. The tests should turn green once again.

lkdvos: great, will merge once everything passes!

lkdvos (approved):

src/implementations/exponential.jl
sanderdemeyer: The error that we get now is that CUDA failed to precompile. I guess this is due to a separate issue, and not related to what we're doing here?

lkdvos: yes, don't worry about this, I'll figure it out!

src/implementations/exponential.jl
```diff
@@ -0,0 +1,109 @@
...
+# --------------
+function exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaLA)
+    check_input(exponential!, A, expA, alg)
+    A = LinearAlgebra.exp!(A)
+    A === expA || copy!(expA, A)
+    return expA
+end
+
+exponential!(A, expA, alg::MatrixFunctionViaEigh) = exponential!((one(eltype(A)), A), expA, alg)
+exponential!(A::AbstractMatrix, expA::AbstractMatrix, alg::MatrixFunctionViaEig) = exponential!((one(eltype(A)), A), expA, alg)
```
Jutho, line 59: Why does one have an `::AbstractMatrix` restriction on the first two arguments and the other not?

lkdvos: I don't think there's a special reason except for overlooking it, I'll patch this in a bit :)

PR #95, opened 2025-11-12
Make `gaugefix!` optional

lkdvos:
After some discussions on PEPS this morning, it turns out that the specific gauge choice can have some (accidental) nice or less nice properties.
For example, it looks like this specific gauge-fixing we have here, for CTMRG algorithms, breaks the C4v symmetry in a more manifest way than not doing it.
The setup is that if we perform CTMRG iteration for a C4v symmetric system without imposing it, not gauge-fixing leads to real and diagonal corner matrices, while this specific gauge choice seems to instead give imaginary ones.
While it sounds like this is a fluke, and definitely not something we ever want to depend on, it does seem fair to have the option to turn off the gauge fixing, both since it isn't strictly required and since if you would want to impose a different gauge yourself, this is now wasted effort.

Here, I added a `gaugefix::Bool = true` keyword argument to the eigenvalue and SVD algorithms to obtain this behavior, and simultaneously slightly refactored the implementation so everything is a little more consistent between the different methods.

### DIFF
--- a/src/implementations/svd.jl
+++ b/src/implementations/svd.jl
@@ -121,14 +121,14 @@ function svd_full!(A::AbstractMatrix, USVᴴ, alg::LAPACK_SVDAlgorithm)
     end
 
     do_gauge_fix = get(alg.kwargs, :gaugefix, default_gaugefix())::Bool
-    lapack_kwargs = Base.structdiff(alg.kwargs, NamedTuple{(:gaugefix,)})
+    alg_kwargs = Base.structdiff(alg.kwargs, NamedTuple{(:gaugefix,)})
 
     if alg isa LAPACK_QRIteration
-        isempty(lapack_kwargs) ||
+        isempty(alg_kwargs) ||
             throw(ArgumentError("invalid keyword arguments for LAPACK_QRIteration"))
         YALAPACK.gesvd!(A, view(S, 1:minmn, 1), U, Vᴴ)
     elseif alg isa LAPACK_DivideAndConquer
-        isempty(lapack_kwargs) ||
+        isempty(alg_kwargs) ||
             throw(ArgumentError("invalid keyword arguments for LAPACK_DivideAndConquer"))
         YALAPACK.gesdd!(A, view(S, 1:minmn, 1), U, Vᴴ)
     elseif alg isa LAPACK_Bisection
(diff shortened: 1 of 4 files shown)


ext/MatrixAlgebraKitGenericLinearAlgebraExt.jl
```diff
@@ -13,18 +13,26 @@ for f! in (:svd_compact!, :svd_full!, :svd_vals!)
     @eval MatrixAlgebraKit.initialize_output(::typeof($f!), A::AbstractMatrix, ::GLA_QRIteration) = nothing
 end
 
-function MatrixAlgebraKit.svd_compact!(A::AbstractMatrix, USVᴴ, ::GLA_QRIteration)
+function MatrixAlgebraKit.svd_compact!(A::AbstractMatrix, USVᴴ, alg::GLA_QRIteration)
     F = svd!(A)
     U, S, Vᴴ = F.U, Diagonal(F.S), F.Vt
-    return MatrixAlgebraKit.gaugefix!(svd_compact!, U, S, Vᴴ, size(A)...)
+
+    dogaugefix = get(alg.kwargs, :gaugefix, true)::Bool
```
kshyatt: might be easier to read as `do_gauge_fix`?

docs/src/user_interface/decompositions.md
```diff
@@ -371,3 +386,47 @@ norm(A * N1') < 1e-14 && norm(A * N2') < 1e-14 &&
...
+
+For eigenvectors, this means that for each column `v` of `V`, we multiply by `conj(sign(v[i]))` where `i` is the index of the entry with largest absolute value.
+
+For singular vectors, we apply the phase factor to both `u` and `v` based on the entry with largest magnitude in `u`.
+This preserves the decomposition `A = U * Σ * Vᴴ` while fixing the gauge.
+
+### Disabling Gauge Fixing
+
+Gauge fixing is enabled by default for all eigenvalue and singular value decompositions.
+If you prefer to obtain the raw results from the underlying LAPACK routines without gauge fixing, you can disable it using the `gaugefix` keyword argument:
```
Jutho:
```suggestion
If you prefer to obtain the raw results from the underlying computational routines without gauge fixing, you can disable it using the `gaugefix` keyword argument:
```

src/common/gauge.jl
```diff
@@ -1,8 +1,64 @@
...
+        s = conj(sign(_argmaxabs(u)))
+        u .*= s
+        v .*= conj(s)
+    end
+    return (U, Vᴴ)
+end
+
+function gaugefix!(::typeof(svd_trunc!), U, Vᴴ)
+    m, n = size(U, 1), size(Vᴴ, 2)
+    for j in 1:min(m, n)
```
Jutho: At which stage is this called in `svd_trunc!`. If we have a `svd_trunc!` algorithm that does not first compute the full decomposition (randomized SVD?), then number of column of `U` = number of rows of `Vᴴ` will be smaller than `min(m,n)`.

src/implementations/svd.jl
```diff
@@ -365,54 +335,65 @@ function svd_full!(A::AbstractMatrix, USVᴴ, alg::GPU_SVDAlgorithm)
...
+
+    return Utr, Str, Vᴴtr, ϵ
 end
 
 function svd_compact!(A::AbstractMatrix, USVᴴ, alg::GPU_SVDAlgorithm)
     check_input(svd_compact!, A, USVᴴ, alg)
     U, S, Vᴴ = USVᴴ
+
+    do_gauge_fix = get(alg.kwargs, :gaugefix, default_gaugefix())::Bool
+    lapack_kwargs = Base.structdiff(alg.kwargs, NamedTuple{(:gaugefix,)})
```
Jutho: I don't remember if I made this comment earlier, but would it make sense to call all the `lapack_kwargs` something like `alg_kwargs`, `computational_kwargs` or `driver_kwargs`, since this is not even LAPACK anymore at this point?

lkdvos:
We could add something like this to the documentation if you like, perhaps in a footnote, although I'm not sure if this helps that much.
I don't think we are claiming anywhere that our gauge choice is smooth or unique.
In general I do think that the documentation of a software package might not be where people go to learn about this though, so it isn't entirely obvious to me to whom this would be directed 🙃

PR #100, opened 2025-11-17
A few more updates for GPU compatibility for TensorKit

kshyatt:
`Base.require_one_based_indexing` returns a `Bool` on some versions, not `n` (integer).

Converting to the full `CuMatrix`/`ROCMatrix` is inefficient/ugly but dispatching on a `SubArray` of a `Diagonal` of a `CuVector` (or `ROCVector`) is also really ugly. Not sure what the best approach is here, but this works for now (we can revisit if it's really problematic).

### DIFF
--- a/test/amd/projections.jl
+++ b/test/amd/projections.jl
@@ -80,6 +80,7 @@ end
 
         m == n && @testset "DiagonalAlgorithm" begin
             A = Diagonal(ROCArray(randn(rng, T, m)))
+            alg = PolarViaSVD(DiagonalAlgorithm())
             W = project_isometric(A, alg)
             @test isisometric(W)
             W2 = project_isometric(W, alg)


lkdvos (commented):
Is it only the `Diagonal` giving issues?
We could also just add a specialization for `Diagonal` in general,  which we should be able to handle in a GPU friendly way:

```julia
function project_hermitian_native!(A::Diagonal, B::Diagonal, ::Val{anti}) where {anti}
   if anti
       diagview(A) .= imag.(diagview(B)) .* im
   else
       diagview(A) .= real.(diagview(B))
   end
end
```

[edit] I think we have a utility function somewhere for the imaginary part, I can't remember the name though

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -128,17 +128,23 @@ function MatrixAlgebraKit._project_hermitian_diag!(A::StridedROCMatrix, B::Strid
...
     all(A .== -adjoint(A))
-MatrixAlgebraKit.isantihermitian_exact(A::Diagonal{T, <:StridedROCVector{T}}) where {T} =
-    all(A.diag .== -adjoint(A.diag))
+MatrixAlgebraKit.isantihermitian_exact(A::Diagonal{T, <:StridedROCVector{T}}) where {T <: Real} = all(iszero(A.diag))
+MatrixAlgebraKit.isantihermitian_exact(A::Diagonal{T, <:StridedROCVector{T}}) where {T <: Complex} =
+    all(iszero.(real.(A.diag)))
 MatrixAlgebraKit.isantihermitian_approx(A::StridedROCMatrix; kwargs...) =
     @invoke MatrixAlgebraKit.isantihermitian_approx(A::Any; kwargs...)
+function MatrixAlgebraKit.isantihermitian_approx(A::Diagonal{T, <:StridedROCVector{T}}; atol, rtol, kwargs...) where {T <: Real}
+    return sum(abs2, A.diag) ≤ max(atol, rtol * norm(A))
```
Jutho: This is a strange condition, as the left hand side is `norm(A)^2`, so this is saying `norm(A)^2 < max(atol, rtol * norm(A))`

This amounts to `norm(A)^2 < atol` or `norm(A) < rtol`, both of which look off.

Because a real diagonal matrix equals its hermitian project and has zero non-hermitian projection, I think the condition simply needs to be `norm(A) < atol` to be compatible with how approximate antithermiticity is checked elsewhere.

src/implementations/projections.jl
```diff
@@ -61,6 +63,15 @@ function project_isometric!(A::AbstractMatrix, W, alg::AbstractAlgorithm)
     return W
 end
 
+function project_hermitian_native!(A::Diagonal, B::Diagonal, ::Val{anti}; kwargs...) where {anti}
+    if anti
+        diagview(A) .= imag.(diagview(B)) .* im
```
Jutho:
```suggestion
        diagview(A) .= _imimag.(diagview(B))
```

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -128,17 +128,23 @@ function MatrixAlgebraKit._project_hermitian_diag!(A::StridedROCMatrix, B::Strid
...
 
 MatrixAlgebraKit.ishermitian_exact(A::StridedROCMatrix) = all(A .== adjoint(A))
-MatrixAlgebraKit.ishermitian_exact(A::Diagonal{T, <:StridedROCVector{T}}) where {T} =
-    all(A.diag .== adjoint(A.diag))
+MatrixAlgebraKit.ishermitian_exact(A::Diagonal{T, <:StridedROCVector{T}}) where {T <: Real} = true
+function MatrixAlgebraKit.ishermitian_exact(A::Diagonal{T, <:StridedROCVector{T}}) where {T <: Complex}
+    return all(isreal.(A.diag))
+end
 MatrixAlgebraKit.ishermitian_approx(A::StridedROCMatrix; kwargs...) =
     @invoke MatrixAlgebraKit.ishermitian_approx(A::Any; kwargs...)
```
Jutho: Is this kind of definition useful? What happens if it is removed?

src/implementations/svd.jl
```diff
@@ -356,6 +356,7 @@ function svd_full!(A::AbstractMatrix, USVᴴ, alg::GPU_SVDAlgorithm)
 
     return USVᴴ
 end
+svd_full!(A::Diagonal, USVᴴ, alg::GPU_SVDAlgorithm) = svd_full!(diagm(A.diag), USVᴴ, alg)
```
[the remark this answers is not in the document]
kshyatt: We currently don't have a `DiagonalGPUAlgorithm`, which is why this happens, I suppose

PR #106, opened 2025-12-01
Implement `DiagonalAlgorithm`s for the GPU

kshyatt:
Decided to try doing these as separate algos for each backend, as I think AMDGPU.jl does not support `sortperm` yet (but CUDA.jl does).


lkdvos (commented):
I might be missing something obvious, but there doesn't seem to be any specific specializations for the different diagonal algorithm types. Does that mean we can simply take the other changes as small fixes and use `DiagonalAlgorithm` for all of them?

PR #107, opened 2025-12-02
[AD] add chainrules support for `svd_vals` and `eig(h)_vals` and add `diagonal`

### LINKED ISSUE #88: [Autodiff] `svd_vals` is not differentiable
We still need to add a rule for `svd_vals`, and presumably also `eig_vals` and `eigh_vals`,

MWE:

```julia
julia> using Zygote, MatrixAlgebraKit

julia> f(A) = sum(svd_vals(A))

julia> gradient(f, A)
ERROR: Can't differentiate foreigncall expression $(Expr(:foreigncall, :(Core.tuple(:dgesdd_64_, MatrixAlgebraKit.YALAPACK.libblastrampoline)), Nothing, svec(Ref{UInt8}, Ref{Int64}, Ref{Int64}, Ptr{Float64}, Ref{Int64}, Ptr{Float64}, Ptr{Float64}, Ref{Int64}, Ptr{Float64}, Ref{Int64}, Ptr{Float64}, Ref{Int64}, Ptr{Int64}, Ptr{Int64}, Int64), 0, :(:ccall), %237, %239, %241, %243, %245, %247, %249, %251, %253, %255, %257, %259, %261, %263, %264, %235, %234, %232, %230, %228, %226, %224, %222, %220, %218, %216, %214, %212, %210, %208)).
You might want to check the Zygote limitations documentation.
https://fluxml.ai/Zygote.jl/latest/limitations
```

Brought up by @leburgel

lkdvos:
Here I write chainrules for the different `f_vals` functions.
To make these implementations sufficiently generic, I added `f_vals_pullback` functions that may be overloaded, but usually don't have to be since they simply fall back to `f_pullback` with the correct zerotangents filled in.

A small hiccup here is that in order to do this, I effectively require the inverse operation of `diagview`, which I introduced here as `diagonal`. I didn't want to simply use `Diagonal`, since that has the restriction that it has to return a `Diagonal` type, and MatrixAlgebraKit feels like the correct place to introduce such a function.

Finally, I also updated the mooncake rules to make use of the updated pullback functions. As I'm still getting the hang of that, it might be nice to double-check that I did that correctly.

This PR fixes #88.

### DIFF
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -63,8 +63,9 @@ export notrunc, truncrank, trunctol, truncerror, truncfilter
         Expr(
             :public, :left_polar_pullback!, :right_polar_pullback!,
             :qr_pullback!, :qr_null_pullback!, :lq_pullback!, :lq_null_pullback!,
-            :eig_pullback!, :eig_trunc_pullback!, :eigh_pullback!, :eigh_trunc_pullback!,
-            :svd_pullback!, :svd_trunc_pullback!
+            :eig_pullback!, :eig_trunc_pullback!, :eig_vals_pullback!,
+            :eigh_pullback!, :eigh_trunc_pullback!, :eigh_vals_pullback!,
+            :svd_pullback!, :svd_trunc_pullback!, :svd_vals_pullback!
         )
     )
     eval(Expr(:public, :is_left_isometric, :is_right_isometric))


lkdvos:
Apologies for the fast merge, I thought this was a pretty small change somehow.
Given that nothing is released, we can definitely still make changes here if you like, although I do think that specifically for the comments above the main point is to leave a hook to overload which does not require overloading the rrules of either ChainRules or Mooncake.

PR #108, opened 2025-12-02
Support Subarray{<:Adjoint{<:GPUMatrix}}

kshyatt:
This is needed to support the blocks of an `AdjointTensorMap` on GPU.

### DIFF
--- a/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
+++ b/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
@@ -129,16 +129,16 @@ function MatrixAlgebraKit._project_hermitian_diag!(A::SupportedROCMatrix, B::Sup
     return nothing
 end
 
-# avoids calling the `SupportedMatrix` specialization to avoid scalar indexing,
+# avoids calling the `StridedMatrix` specialization to avoid scalar indexing,
 # use (allocating) fallback instead until we write a dedicated kernel
-MatrixAlgebraKit.ishermitian_exact(A::SupportedROCMatrix) = A == A'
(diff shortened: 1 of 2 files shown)


ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -112,31 +112,33 @@ function _project_hermitian_diag_kernel(A, B, ::Val{false})
...
-function MatrixAlgebraKit._project_hermitian_diag!(A::StridedROCMatrix, B::StridedROCMatrix, ::Val{anti}) where {anti}
+function MatrixAlgebraKit._project_hermitian_diag!(A::SupportedROCMatrix, B::SupportedROCMatrix, ::Val{anti}) where {anti}
     thread_dim = 512
     block_dim = cld(size(A, 1), thread_dim)
     @roc groupsize = thread_dim gridsize = block_dim _project_hermitian_diag_kernel(A, B, Val(anti))
     return nothing
 end
 
-# avoids calling the `StridedMatrix` specialization to avoid scalar indexing,
+# avoids calling the `SupportedMatrix` specialization to avoid scalar indexing,
```
lkdvos: I think this really has to be `StridedMatrix`, the specialization is only defined for subtypes of `StridedMatrix`, so I think we only have to opt-out for the intersection of `SupportedROCMatrix` and `StridedMatrix`?

kshyatt:
> Looks good to me, although I would have guessed that the `ishermitian` fallbacks automatically catch the non-`<:StridedMatrix` types?

Unfortunately not, at least not from my experiments :(

kshyatt:
Sorry, it's for the projection methods that we need this. I agree the fallbacks catch it for the `is_` checkers.

kshyatt:
Yes, we could save ourselves some hurt by testing fewer algorithms and only
testing Float32 and ComplexF64 ?

PR #116, opened 2025-12-12
Split svd_trunc

kshyatt:
Now we have a `svd_trunc_with_err` that returns epsilon for those who wish, and `svd_trunc` returns only the truncated USVh

### DIFF
--- a/src/interface/svd.jl
+++ b/src/interface/svd.jl
@@ -86,7 +86,7 @@ truncation strategy is already embedded in the algorithm.
     possibly destroys the input matrix `A`. Always use the return value of the function
     as it may not always be possible to use the provided `USVᴴ` as output.
 
-See also [`svd_trunc_no_error(!)`](@ref svd_trunc), [`svd_full(!)`](@ref svd_full),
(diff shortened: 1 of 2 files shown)


src/implementations/svd.jl
```diff
@@ -287,6 +288,22 @@ function check_input(
     return nothing
 end
 
+function check_input(
```
Jutho: Would it be useful to combine this definition with the previous one in the form
```julia
function check_input(::Union{typeof(svd_trunc!),typeof(svd_trunc_with_error!)}, A::AbstractMatrix, USVᴴ, alg::CUSOLVER_Randomized)
```

src/implementations/svd.jl
```diff
@@ -298,6 +315,17 @@ function initialize_output(
     return (U, S, Vᴴ)
 end
 
+function initialize_output(
+        ::typeof(svd_trunc_with_err!), A::AbstractMatrix, alg::TruncatedAlgorithm{<:CUSOLVER_Randomized}
```
Jutho: Same here

PR #119, opened 2025-12-17
Use TestSuite for QR and LQ

kshyatt:
Goal here is to break up https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/93 into more manageable diffs, starting with LQ and QR

### DIFF
--- a/test/qr.jl
+++ b/test/qr.jl
@@ -15,7 +15,7 @@ is_buildkite = get(ENV, "BUILDKITE", "false") == "true"
 m = 54
 for T in (BLASFloats..., GenericFloats...), n in (37, m, 63)
     TestSuite.seed_rng!(123)
-    if is_buildkite && T ∈ BLASFloats
+    if T ∈ BLASFloats
         if CUDA.functional()
             CUDA_QR_ALGS = (CUSOLVER_HouseholderQR(; positive = false), CUSOLVER_HouseholderQR(; positive = true))
             TestSuite.test_qr(CuMatrix{T}, (m, n); test_pivoted = false, test_blocksize = false)
@@ -34,7 +34,7 @@ for T in (BLASFloats..., GenericFloats...), n in (37, m, 63)
                 TestSuite.test_qr_algs(Diagonal{T, ROCVector{T}}, m, (DiagonalAlgorithm(),))
             end
         end
-    else
+    elseif !is_buildkite
         if T ∈ BLASFloats
             TestSuite.test_qr(T, (m, n))
             LAPACK_QR_ALGS = (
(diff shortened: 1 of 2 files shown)


test/testsuite/lq.jl
```diff
@@ -0,0 +1,162 @@
...
+        @test isrightnull(Nᴴ2, A; atol, rtol)
+        @test isisometric(Nᴴ2; side = :right, atol, rtol)
+
+        # do we support `pivoted = true`?
+        if test_pivoted
+            Nᴴpiv = @testinferred lq_null(A; pivoted = true)
+            @test isrightnull(Nᴴpiv, A; atol, rtol)
+            @test isisometric(Nᴴpiv; side = :right, atol, rtol)
+            #else # DISABLE for now as lq_null does support pivoting...
+            #    @test_throws Exception lq_null(A; pivoted = true)
```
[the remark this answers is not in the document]
kshyatt: Apparently not? Tried locally and `lq_null` doesn't throw...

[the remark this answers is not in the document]
kshyatt: Fixed now :)

lkdvos (commented):
Looks great to me, I think only the remaining comment about testing the positive-ness of the alg implementation if need be is what I would still add, otherwise this is good to go for me.

Thanks for all the work on this, I really think this is a massive improvement!

kshyatt:
OK, now the LQ algo tests are in!

test/qr.jl
```diff
@@ -1,223 +1,64 @@
...
-        @test Qp * Rp ≈ A
-        @test isunitary(Qp)
-        @test all(≥(zero(real(T))), real(diag(Rp))) &&
-            all(≈(zero(real(T)); atol), imag(diag(Rp)))
-
-        # null
-        N = @constinferred qr_null(A)
-        @test N isa AbstractMatrix{T} && size(N) == (m, 0)
+if !is_buildkite
+    for T in (BLASFloats..., GenericFloats...)
```
Jutho: Could the block below just be below the `else # if isbuildkite` on line 37, before the `if T ∈ BLASFloats`?
(Same applies for lq.jl)

test/qr.jl
```diff
@@ -1,223 +1,62 @@
...
-            @test_throws ArgumentError qr_compact!(copy!(Q2, A), (Q2, noR); blocksize = 8)
+@isdefined(TestSuite) || include("testsuite/TestSuite.jl")
+using .TestSuite
+
+is_buildkite = get(ENV, "BUILDKITE", "false") == "true"
+
+m = 54
+for T in (BLASFloats..., GenericFloats...), n in (37, m, 63)
+    TestSuite.seed_rng!(123)
+    if is_buildkite && T ∈ BLASFloats
```
lkdvos: Can we slightly alter the logic for this, such that if we would locally trigger the tests, the GPU tests also run? I.e. probably its better to exclude things if `!is_buildkite`

kshyatt: OK, done, I think

PR #122, opened 2025-12-19
stabilize eig(h) tests

Jutho:
This tries to remove all (near) degeneracies in matrices that we will use for the AD of the eigenvalue decomposition, thereby hopefully stabilizing the tests.

I also made some progress with the case of degeneracies, for which I will make a separate PR with some hand written tests.


kshyatt (approved):
LGTM, just left some nitpicky comments

lkdvos (approved):

PR #126, opened 2025-12-23
Use Testsuite for AD tests

kshyatt:
This lets us reuse a lot of the setup infrastructure for ChainRules, Mooncake, and (soon) Enzyme. Also starts testing the AD rules on GPU.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -25,7 +25,12 @@ if filter_tests!(testsuite, args)
         delete!(testsuite, "codequality")
     else
         is_apple_ci = Sys.isapple() && get(ENV, "CI", "false") == "true"
-        (Sys.iswindows() || is_apple_ci) && delete!(testsuite, "enzyme")
+        if is_apple_ci
+            delete!(testsuite, "enzyme")
+            delete!(testsuite, "mooncake")
+            delete!(testsuite, "chainrules")
+        end
+        Sys.iswindows() && delete!(testsuite, "enzyme")
     end
 end
 


ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -183,4 +183,23 @@ function MatrixAlgebraKit._avgdiff!(A::StridedCuMatrix, B::StridedCuMatrix)
...
+
+function LinearAlgebra.sylvester(A::AnyCuMatrix, B::AnyCuMatrix, C::AnyCuMatrix)
+    #=m = size(A, 1)
+    n = size(B, 2)
+    I_n = fill!(similar(A, n), one(eltype(A)))
+    I_m = fill!(similar(B, m), one(eltype(B)))
+    L   = kron(diagm(I_n), A) + kron(adjoint(B), diagm(I_m))
+    x_vec = L \ -vec(C)
+    X = CuMatrix(reshape(x_vec, m, n))=#
+    hX = sylvester(collect(A), collect(B), collect(C))
```
kshyatt: This is very awful but I wasn't able to find a correct way to do it in five minutes so there you go

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -195,4 +195,23 @@ end
 MatrixAlgebraKit._ind_intersect(A::CuVector{Int}, B::CuVector{Int}) =
     MatrixAlgebraKit._ind_intersect(collect(A), collect(B))
 
+MatrixAlgebraKit.default_pullback_rank_atol(A::AnyCuArray) = eps(norm(CuArray(A), Inf))^(3 / 4)
+MatrixAlgebraKit.default_pullback_gauge_atol(A::AnyCuArray) = MatrixAlgebraKit.iszerotangent(A) ? 0 : eps(norm(CuArray(A), Inf))^(3 / 4)
+function MatrixAlgebraKit.default_pullback_gauge_atol(A::AnyCuArray, As...)
+    As′ = filter(!MatrixAlgebraKit.iszerotangent, (A, As...))
+    return isempty(As′) ? 0 : eps(norm(CuArray.(As′), Inf))^(3 / 4)
```
lkdvos: Why was this needed? what breaks if we don't do `CuArray.(As')`?

src/pullbacks/eig.jl
```diff
@@ -39,12 +39,16 @@ function eig_pullback!(
         VᴴΔV = fill!(similar(V), 0)
         indV = axes(V, 2)[ind]
         length(indV) == pV || throw(DimensionMismatch())
-        mul!(view(VᴴΔV, :, indV), V', ΔV)
+        VᴴΔV[:, indV] .= V' * ΔV
+        #mul!(view(VᴴΔV, :, indV), V', ΔV)
```
lkdvos: I think this change was due to the `Diagonal` thingies, should no longer be needed now (unless it is a GPU thing?)
```suggestion
        mul!(view(VᴴΔV, :, indV), V', ΔV)
```

src/pullbacks/eig.jl
```diff
@@ -39,12 +39,16 @@ function eig_pullback!(
...
 
         mask = abs.(transpose(D) .- D) .< degeneracy_atol
```
lkdvos: The mask probably also belongs inside of the `if`

src/pullbacks/lq.jl
```diff
@@ -36,28 +36,30 @@ function lq_pullback!(
...
-            # change upon small variations, and all of the remaining
-            # columns of ΔQ should be zero for a gauge-invariant
-            # cost function
-            ΔQ2 = view(ΔQ, (p + 1):size(Q, 1), :)
-            Δgauge = max(Δgauge, norm(ΔQ2, Inf))
-        end
-        if !iszerotangent(ΔL)
-            ΔL22 = view(ΔL, (p + 1):m, (p + 1):minmn)
-            Δgauge = max(Δgauge, norm(ΔL22, Inf))
+    if isa(ΔA, Array) # not GPU friendly
```
lkdvos: same comment here about refactoring into a separate function

src/pullbacks/qr.jl
```diff
@@ -69,9 +71,11 @@ function qr_pullback!(
             # how the full Q2 will change, but this we omit for now, and we consider
             # Q2' * ΔQ2 as a gauge dependent quantity.
             Q1dΔQ2 = Q1' * ΔQ2
-            Δgauge = norm(mul!(copy(ΔQ2), Q1, Q1dΔQ2, -1, 1), Inf)
-            Δgauge ≤ gauge_atol ||
-                @warn "`qr` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
+            if isa(ΔA, Array) # not GPU friendly
```
lkdvos: same refactor into a function comment here

src/pullbacks/qr.jl
```diff
@@ -87,16 +91,18 @@ function qr_pullback!(
     M = zero!(similar(R, (p, p)))
     if !iszerotangent(ΔR)
         ΔR11 = view(ΔR, 1:p, 1:p)
-        M = mul!(M, ΔR11, R11', 1, 1)
+        M += ΔR11 * R11'
```
lkdvos: Same `_mul!!` comment

test/testsuite/ad_utils.jl
```diff
@@ -0,0 +1,418 @@
...
+    T = eltype(A)
+    r = minmn - 5
+    Ard = randn!(similar(A, T, m, r)) * randn!(similar(A, T, r, n))
+    Q, R = qr_compact(Ard)
+    QR = (Q, R)
+    ΔQ = randn!(similar(A, T, m, minmn))
+    Q1 = view(Q, 1:m, 1:r)
+    Q2 = view(Q, 1:m, (r + 1):minmn)
+    ΔQ2 = view(ΔQ, 1:m, (r + 1):minmn)
+    ΔQ2 .= 0
```
lkdvos:
```suggestion
    zero!(ΔQ2)
```

test/testsuite/chainrules.jl
```diff
@@ -0,0 +1,612 @@
...
+
+for f in
+    (
+        :qr_compact, :qr_full, :qr_null, :lq_compact, :lq_full, :lq_null,
+        :eig_full, :eig_trunc, :eig_vals, :eigh_full, :eigh_trunc, :eigh_vals,
+        :eig_trunc_no_error, :eigh_trunc_no_error,
+        :svd_compact, :svd_trunc, :svd_trunc_no_error, :svd_vals,
+        :left_polar, :right_polar,
+    )
+    copy_f = Symbol(:cr_copy_, f)
```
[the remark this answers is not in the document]
lkdvos: I think I just got the difference with the `mc_` ones, I thought this should all be run in separate modules though so there shouldn't necessarily be any nameclashing here?
I might be missing something though

lkdvos (commented):
I left a bunch of smaller comments through the code, and only quickly went through the actual test suites. Am I correct that these haven't changed much?

I think in general it would be nice to try and link some of the GPU issues we are facing with actual issue numbers and resolve them through dispatch alternatives, rather than having to alter the code for regular arrays as well.
In particular, `isa(A, Array)` seems a bit to strict since we also support views etc, which would then also take the slow path.

PR #128, opened 2025-12-23
Use TestSuite for orthnull

kshyatt:
This one could probably use some work on the tests themselves (trunc not really playing well with the GPU) and more algorithm tests

### DIFF
--- a/src/interface/orthnull.jl
+++ b/src/interface/orthnull.jl
@@ -443,7 +443,6 @@ left_orth_alg(alg::LeftOrthAlgorithm) = alg
 left_orth_alg(alg::QRAlgorithms) = LeftOrthViaQR(alg)
 left_orth_alg(alg::PolarAlgorithms) = LeftOrthViaPolar(alg)
 left_orth_alg(alg::SVDAlgorithms) = LeftOrthViaSVD(alg)
-left_orth_alg(alg::DiagonalAlgorithm) = LeftOrthViaSVD(alg)
 left_orth_alg(alg::TruncatedAlgorithm{<:SVDAlgorithms}) = LeftOrthViaSVD(alg)
 
 """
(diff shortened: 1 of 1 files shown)


test/testsuite/orthnull.jl
```diff
@@ -0,0 +1,276 @@
...
+        m, n = size(A)
+        minmn = min(m, n)
+        @test V isa typeof(A) && size(V) == (m, minmn)
+        @test C isa typeof(A) && size(C) == (minmn, n)
+        @test eltype(N) == eltype(A) && size(N) == (m, m - minmn)
+        @test V * C ≈ A
+        @test isisometric(V)
+        @test LinearAlgebra.norm(A' * N) ≈ 0 atol = MatrixAlgebraKit.defaulttol(T)
+        @test isisometric(N)
+        @test collect(V) * collect(V)' + collect(N) * collect(N)' ≈ I
```
lkdvos: It might make sense to separate this out into a dedicated function in the testsuite, so we can specialize the GPU implementations to add the `collect`?
I also think it just reads quite nice to have something like for example `@test iscomplete(V, N)` or `resolves_identity(V, N)` or `iscompletespan(V, N)`

test/testsuite/orthnull.jl
```diff
@@ -0,0 +1,276 @@
...
+
+        # passing an algorithm
+        if !isa(A, Diagonal)
```
lkdvos: Why do we need this restriction here?

test/testsuite/orthnull.jl
```diff
@@ -0,0 +1,276 @@
...
+    end
+end
+
+function test_left_orthnull(
+        T::Type, sz;
+        atol::Real = 0, rtol::Real = precision(T),
+        kwargs...
+    )
+    summary_str = testargs_summary(T, sz)
+    return @testset "left_orth! and left_null! $summary_str" begin
```
lkdvos: This seems to me like a rather large testset (just based on feeling, definitely feel free to disagree), where it might make sense to try and separate out some of the components. This might also allow us to use dispatch rather than `if` statements for the different handling of the various input types, such that we could possibly also reuse this a bit more in downstream packages?

lkdvos (approved):

PR #130, opened 2025-12-23
Use TestSuite for eig

kshyatt:
Made some changes to the `GenericSchur` wrappers so that input `DV` to `eig_full!` actually does get modified, for example.

### DIFF
--- a/ext/MatrixAlgebraKitGenericSchurExt.jl
+++ b/ext/MatrixAlgebraKitGenericSchurExt.jl
@@ -13,18 +13,7 @@ MatrixAlgebraKit.initialize_output(::typeof(eig_full!), A::AbstractMatrix, ::GS_
 MatrixAlgebraKit.initialize_output(::typeof(eig_vals!), A::AbstractMatrix, ::GS_QRIteration) = nothing
 
 function MatrixAlgebraKit.eig_full!(A::AbstractMatrix, DV, ::GS_QRIteration)
-    D_, V_ = GenericSchur.eigen!(A)
-    D, V = DV
-    if !isnothing(D)
-        copyto!(D, Diagonal(D_))
-    else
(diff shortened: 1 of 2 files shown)


lkdvos (commented):
Except for the comment about the `copy` and the `===` tests I think this is also good for me

kshyatt:
Previously all our tests were checking that that D, V you get out of the !
methods is the same D, V that went in, in terms of memory location. That
might be worth holding on to in case people are eg passing in a SubArray
for D and V?

Jutho:
While I agree that we do not commit to actually using the given output arguments in our current interface specification, I am starting to wonder if that was a mistake. It becomes very hard to draw a line. There are various `...ViaTransposed...` methods, where the final step also involves something like `adjoint!(given_output_argument, newly_allocated_output_of_transposed_problem)`. This is of the same form as a `copy!`, so why are we doing this and not simply return `adjoint(newly_allocated_output_of_transposed_problem)`, thereby discarding `given_output_argument`.

I do think type stability is a good reason. In that case, it would be fairly obvious that that the two approaches yield different output types. But indeed, as pointed out by @kshyatt , also in the more subtle cases that do not involve an `adjoint!` but only a `copy!`, there is a difference if the given output matrix is not a simple `Matrix` but some `SubArray` or ... . So now we have output types that depend on the chosen algorithm? Is that desirable?

I don't think including non-mutable types in that discussion is really valid; if something like `StaticArray`s are ever supported, they should not go via the bang methods altogether.

kshyatt:
I also think if we want to modify these guarantees, it's a bit of a separate issue from this PR (e.g. we can always modify things later to enforce the guarantee and update the tests if/when we do so)

kshyatt:
(Can this be merged)

PR #131, opened 2025-12-23
Use TestSuite for eigh

kshyatt:


test/testsuite/eigh.jl
```diff
@@ -0,0 +1,190 @@
...
+        @testset "specify truncation algorithm" begin
+            atol = sqrt(eps(real(eltype(T))))
+            m4 = 4
+            smallA = randn!(similar(A, (m4, m4)))
+            V = T <: Diagonal ? I : qr_compact(smallA)[1]
+            Ddiag = similar(A, real(eltype(T)), m4)
+            copyto!(Ddiag, real(eltype(T))[0.9, 0.3, 0.1, 0.01])
+            D = Diagonal(Ddiag)
+            A = V * D * V'
+            A = (A + A') / 2
```
lkdvos: Same here

lkdvos (approved):
Left a few small comments about maybe using project_hermitian in some of the tests, doesn't make too much of a difference but if nothing else it covers the package functions slightly better. Otherwise good to go for me

Issue #129, opened 2025-12-23
[Feature request] `lq_null` and `qr_null` for `GenericFloat`s


kshyatt:
We support `qr_full`, `qr_compact`, etc. But without the null space methods we can't support the full `orthnull` suite which is very sad, and it shouldn't be much of a lift to add support for them.

PR #132, opened 2025-12-23
add gla_householder_qr_null

Jutho:
As requested by @kshyatt in #129 .


kshyatt:
Can we turn on the tests for this by getting rid of the `test_null` `kwarg` in the QR and LQ testsuites?

Jutho:
So it seems there are still issues with the eigenvalue AD. I'll investigate this elsewhere.

lkdvos (approved):
Looks good to me!

The only thing I was wondering is whether or not it would be easy/straightforward to share some of the code with the regular `qr_null` implementations, but I had a quick look and that doesn't really seem to be the case, so this is definitely good to go for me.

Left a small style comment

PR #135, opened 2025-12-28
some yalapack fixes and test tryouts

Jutho:


lkdvos (approved):
Out of curiosity, did you find these YALAPACK bugs through the new JET report? Otherwise we might have to also hunt for some extra test coverage at some point

PR #145, opened 2026-01-06
Implement `SafeSVD` algorithm

VictorVanthilt:
When performing a singular value decomposition it can be beneficial to try to use `LAPACK_DivideAndConquer` for its speed but if it fails, use the more stable but slower `LAPACK_QRIteration` algorithm.

Thereto I implemented a very minimal `SafeSVD` algorithm that makes an extra copy of the input data, tries `LAPACK_DivideAndConquer` on the copy, and if it fails performs `LAPACK_QRIteration` on the original data.

Feel free to comment on what other magic functions I need to add to make things like Mooncake happy.
I haven't added any tests yet, does anyone know a pathological matrix that will fail with `YALAPACK.gesdd!` but succeed with `YALAPACK.gesvd!`?

### DIFF
--- a/src/implementations/svd.jl
+++ b/src/implementations/svd.jl
@@ -134,7 +134,7 @@ function svd_full!(A::AbstractMatrix, USVᴴ, alg::LAPACK_SVDAlgorithm)
     elseif alg isa SafeSVD
         isempty(alg_kwargs) ||
             throw(ArgumentError("invalid keyword arguments for SafeSVD"))
-        # extra copy to avoid modifying input if it goes wrong
+        # extra copy to avoid modifying input if sdd goes wrong
         A′ = copy(A)
         try
             YALAPACK.gesdd!(A′, view(S, 1:minmn, 1), U, Vᴴ)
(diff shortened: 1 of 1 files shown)


kshyatt:
Oh I got it mixed up, neither gpu library has `gesdd` yet. So no worries!

It might nevertheless be nice to do this in a way that's easily extensible to non LAPACK backends for the future, in case either does add `gesdd`

lkdvos (commented):
Given the remark about the GPU implementations, it might really make sense to just spend the time and make it a bit more generic:

```julia
struct FallbackAlgorithm{A <: AbstractAlgorithm, B <: AbstractAlgorithm} <: AbstractAlgorithm
    alg1::A
    alg2::B
end

for f in (:svd_compact, :svd_full)
    f! = Symbol(f, :!)

    # hijack non-mutating function to avoid double-copy
    @eval function $f(A, alg::FallbackAlgorithm)
        Ac = copy_input($f, A)
        out = initialize_output($f!, A, alg.alg1)
        return try
            $f!(Ac, out, alg.alg1)
        catch
            copy!(Ac, A)
            $f!(Ac, out, alg.alg2)
        end
    end

    # implement mutating function to make copy
    @eval function $f!(A, out, alg::FallbackAlgorithm)
        Ac = copy_input($f, A)
        return try
            $f!(Ac, out, alg.alg1)
        catch
            $f!(A, out, alg.alg2)
        end
    end
end
```

I think our implementations of the other functions might even be generic enough to make this "just work".

This also needs some additional tests, and possibly some `select_algorithm` overloads or additional constructors to make this a bit more convenient

Jutho:
While I am typically among the first to consider possible generalisations, in this case I am not sure if this is really the right approach. In fact, I would go further and say that I think this should probably be called something like `LAPACK_SafeSVD` (but could become the default for BLAS matrices), and would like to write a specialised YALAPACK call for this, that doesn't really throw the error but simply does the LAPACK style `info` checking, and possibly reuse some of the work space between the `gesvd!` and `gesdd!` calls.

PR #148, opened 2026-01-08
Try to do SVD truncation on GPU with _ind_intersect

kshyatt:
I *think* this should be ok because the svd values should always be sorted (and the tests passed) but I'm not sure if that continues to be true for all the block-sparse tensors that may rely on this.


lkdvos (approved):
I would say that if the tests pass here, we can revert the GPUArrays source thing and then just force-merge, since these changes have nothing to do with this PR to begin with (but it would be nice to verify that this implementation actually works).

I have no strong opinions on merging this before or after a tag.

PR #149, opened 2026-01-08
Bump version

kshyatt:
Can go in before or after https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/148, but not having a new version is holding back the CUDA factorizations at TensorKit


kshyatt:
Sure will do.

lkdvos:
Since this will fail due to the 1.12 GPU breaking, I will merge this as-is

PR #151, opened 2026-01-12
Sort diagonal `eig` values and standardize output types

lkdvos:
This was already changed for `eigh`, and I now made this consistent.

However, while playing around with this I have to say that I don't think we are sorting the LAPACK factorizations either, so this seems inconsistent with that now.
I then browsed through what LinearAlgebra does, and they very explicitly mention that for Diagonal inputs `sortby` might not be supported, and the sorting of the eigenvalues is not consistent between `D` and `Matrix(D) + eps` (note that the eps is necessary since LinearAlgebra does a runtime check for diagonal, therefore leading to the incredibly stupid difference between a diagonal eigendecomposition and a matrix epsilon away from that)

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -30,6 +30,8 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Fixed
 
+- Eigenvalue decompositions of diagonal inputs are sorted and have the same type as non-diagonal inputs ([#151](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/151)
+
 ## [0.6.2](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/compare/v0.6.1...v0.6.2) - 2026-01-08
 
 ### Added


src/implementations/eig.jl
```diff
@@ -30,23 +30,21 @@ end
 
 function check_input(::typeof(eig_full!), A::AbstractMatrix, DV, ::DiagonalAlgorithm)
     m, n = size(A)
-    @assert m == n && isdiag(A)
+    ((m == n) && isdiag(A)) || throw(DimensionMismatch("diagonal input matrix expected"))
```
kshyatt: Can we print the actual dimensions here with a lazy string?

src/implementations/eig.jl
```diff
@@ -30,23 +30,21 @@ end
...
-    # Diagonal doesn't need to promote to complex scalartype since we know it is diagonalizable
-    @check_scalar(D, A)
-    @check_scalar(V, A)
+    @check_scalar(V, A, complex)
     return nothing
 end
 function check_input(::typeof(eig_vals!), A::AbstractMatrix, D, ::DiagonalAlgorithm)
     m, n = size(A)
-    @assert m == n && isdiag(A)
+    ((m == n) && isdiag(A)) || throw(DimensionMismatch("diagonal input matrix expected"))
```
kshyatt: Same as above

PR #156, opened 2026-01-16
Pullbacks for `Diagonal` inputs

lkdvos:
Ran into some issues in https://github.com/QuantumKitHub/TensorKit.jl/pull/350, we are currently not testing pullbacks for `Diagonal` inputs, and apparently this is not working right now.
The issue is that we want to in-place add to $\Delta A$, but we need an additional projection on the diagonal subspace.

I'm guessing similar issues would appear for other restricted input vector spaces, for now I'd be okay with this simple fix only for diagonal.

I'll try add a similar thing for the other pullbacks as well. In principle we can probably improve the performance a bit more but I'm not sure if this is worth the effort.

### DIFF
--- a/src/pullbacks/svd.jl
+++ b/src/pullbacks/svd.jl
@@ -99,6 +99,17 @@ function svd_pullback!(
     end
     return ΔA
 end
+function svd_pullback!(
+        ΔA::Diagonal, A, USVᴴ, ΔUSVᴴ, ind = Colon();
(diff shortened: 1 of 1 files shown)


kshyatt:
FWIW there was some work done for this in the `testsuite-ad` PR

kshyatt:
So you might be able to cannibalize part of that

PR #159, opened 2026-01-23
Add some tests for randomized SVD

kshyatt:
Definitely open to feedback on this one but thought we should start testing this. Also simplified the implementation a bit to reduce repeated code.

### DIFF
--- a/test/testsuite/svd.jl
+++ b/test/testsuite/svd.jl
@@ -322,6 +322,6 @@ function test_randomized_svd(T::Type, sz, algs; kwargs...)
         minmn = min(m, n)
         S₀ = collect(svd_vals(A))
         U1, S1, V1ᴴ, ϵ1 = @testinferred svd_trunc(A; alg)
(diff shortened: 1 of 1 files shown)


src/implementations/svd.jl
```diff
@@ -398,29 +398,20 @@ function svd_trunc_no_error!(A::AbstractMatrix, USVᴴ, alg::TruncatedAlgorithm{
     (Utr, Str, Vᴴtr), _ = truncate(svd_trunc!, (U, S, Vᴴ), alg.trunc)
 
     do_gauge_fix = get(alg.alg.kwargs, :fixgauge, default_fixgauge())::Bool
+    # the output matrices here are the same size as for svd_full!
```
Jutho: I am confused about the conclusion of the resolved comments, and just about the validity of this comment in the code, as well as why it is here in between those two code lines. The arguments to `gaugefix!` on the line below are already truncated, right?

PR #172, opened 2026-02-04
Default tolerances for SVD-based nullspaces

### LINKED ISSUE #166: `right_null` for a simple matrix
MWE (using the [latest](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/tree/8de569fd8710039728f70ded17137a0abf685bf2) main branch):

```julia
using MatrixAlgebraKit, LinearAlgebra
# a very simple rank-1 2x2 matrix
A = [1. 0.; 1. 0.]
# it has a 1D (right) nullspace, spanned by [0, 1]
N = nullspace(A)
#= output
2×1 Matrix{Float64}:
 0.0
 1.0
=#

# MAK fails to find it, with either alg = :svd or :qr
right_null(A)
#= output
0×2 Matrix{Float64}
=#
```

For this simple case, the behavior does not match the docstring:

```
Compute an orthonormal basis N = adjoint(Nᴴ) for the kernel of the matrix A, 
i.e. the nullspace of A, such that A * Nᴴ' ≈ 0 and Nᴴ * Nᴴ' ≈ I.
```

lkdvos:
This PR changes the default `notrunc` to `atol = defaulttol()` for SVD-based nullspace computations.

Fixes #166.

### DIFF
--- a/src/MatrixAlgebraKit.jl
+++ b/src/MatrixAlgebraKit.jl
@@ -34,7 +34,7 @@ export left_orth!, right_orth!, left_null!, right_null!
 export Householder, Native_HouseholderQR, Native_HouseholderLQ
 export LAPACK_HouseholderQR, LAPACK_HouseholderLQ, LAPACK_Simple, LAPACK_Expert,
     LAPACK_QRIteration, LAPACK_Bisection, LAPACK_MultipleRelativelyRobustRepresentations,
-    LAPACK_DivideAndConquer, LAPACK_Jacobi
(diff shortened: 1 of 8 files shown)


lkdvos:
Because I don't necessarily have access to the runtime information here, the algorithm selection procedure is running on types, so the whole scaling with the size of the input is a bit annoying. I can set both an absolute and relative tolerance if you like, but honestly I am not that convinced by this, it all feels incredibly arbitrary. For the truncated svd we also don't really provide a default truncation scheme, and I think here a similar thing is happening, by default the QR is chosen, and otherwise you just specify your own tolerances?

PR #174, opened 2026-02-17
AD rules for (anti-) hermitian projection

### LINKED ISSUE #99: [AD] rrules for `project_(anti)hermitian`

lkdvos:
Fixes #99. 

Requires #175. 

I played around a bit with also implementing the `project_isometric`, but since that was taking more significant changes (possibly writing an actual `sylvester` equation solver), I will leave that for a follow-up.
Also, I'm not actually smart enough for Enzyme rules, so will also ignore that for now. (Additionally I'm not sure we need them in that case, so don't want to pollute this PR)

### DIFF
--- a/test/testsuite/projections.jl
+++ b/test/testsuite/projections.jl
@@ -40,6 +40,14 @@ function test_project_antihermitian(
             @test Ba === Ac
             @test isantihermitian(Ba)
             @test Ba ≈ Aa
+
+            # can we supply a destination
+            Ac = deepcopy(A)
+            Ba = instantiate_matrix(T, sz)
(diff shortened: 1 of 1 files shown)


kshyatt:
AD rules yet no Enzyme 🥺 ? Or would you like me to add those?

lkdvos:
The thing is that for Enzyme I would actually be quite interested to see if they can generate better rules than we can manually write, since this is somewhat of the promise of that system (maybe?), and also I really just was to lazy to actually go and investigate, since I basically mostly needed the chainrules ones since that simply errors without.

kshyatt:
Looks good to me, feel free to just say "no" to my comment

PR #197, opened 2026-03-23
Try Enzyme tests with CUDA

kshyatt:
Since https://github.com/EnzymeAD/Enzyme.jl/pull/2692 this should in theory work, let's give it a spin


test/enzyme/projections.jl
```diff
@@ -18,4 +18,8 @@ for T in (BLASFloats..., GenericFloats...)
         TestSuite.test_enzyme_projections(T, (m, m); atol, rtol)
         TestSuite.test_enzyme_projections(Diagonal{T, Vector{T}}, (m, m); atol, rtol)
     end
+    if CUDA.functional()
+        TestSuite.test_enzyme_projections(CuMatrix{T}, (m, n); atol, rtol)
```
Jutho: `n` is not defined here

```suggestion
        TestSuite.test_enzyme_projections(CuMatrix{T}, (m, m); atol, rtol)
```

PR #208, opened 2026-04-17
pullback reorganization

### LINKED ISSUE #96: `svd_pullback` doesn't handle output from `svd_full`
See https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/85#discussion_r2519991306

### LINKED ISSUE #150: SVD pullback of rank-deficient matrix
> I encountered a DimensionMismatch error during the backward pass of an optimization loop involving SVD truncation on $U(1)$ symmetric tensors. I think this issue occurs when the distribution of bond dimensions across different charge sectors changes between the forward pass and the backward pullback. As I don't want to fixed the sectors, I really need this function maybe.
> 
> ```
> ERROR: LoadError: DimensionMismatch: 
> Stacktrace:
>   [1] svd_pullback!(ΔA::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, A::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, USVᴴ::Tuple{Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}, Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}}, ΔUSVᴴ::Vector{Any}, ind::BitVector; tol::Float64, rank_atol::Float64, degeneracy_atol::Float64, gauge_atol::Float64)
>     @ MatrixAlgebraKit ~/.julia/packages/MatrixAlgebraKit/xv5dv/src/pullbacks/svd.jl:51
>   [2] svd_pullback!(ΔA::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, A::Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, USVᴴ::Tuple{Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}, Diagonal{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}}, Base.ReshapedArray{Float64, 2, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{}}}, ΔUSVᴴ::Vector{Any}, ind::BitVector)
>     @ MatrixAlgebraKit ~/.julia/packages/MatrixAlgebraKit/xv5dv/src/pullbacks/svd.jl:24
> ``` 

 _Originally posted by @Ryo-wtnb11 in [#101](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/101#issuecomment-3718602402)_

I investigated this issue in a bit more detail and can distill the problem to the following MWE:

```julia
using Zygote, MatrixAlgebraKit

# generate a rank-deficient matrix
A = zeros(3, 3)
rand!(@view(A[1:2, 1:2]))

# dummy function to derive
function f(A)
    U, S, Vh = svd_compact(A)
    return tr(U * S * Vh)
end

Zygote.gradient(f, A)
ERROR: DimensionMismatch: 
Stacktrace:
  [1] svd_pullback!(ΔA::Matrix{…}, A::Matrix{…}, USVᴴ::Tuple{…}, ΔUSVᴴ::Vector{…}, ind::Colon; tol::Float64, rank_atol::Float64, degeneracy_atol::Float64, gauge_atol::Float64)
    @ MatrixAlgebraKit ~/.julia/packages/MatrixAlgebraKit/xv5dv/src/pullbacks/svd.jl:51
  [2] svd_pullback!
    @ ~/.julia/packages/MatrixAlgebraKit/xv5dv/src/pullbacks/svd.jl:24 [inlined]
  [3] svd_pullback!(ΔA::Matrix{…}, A::Matrix{…}, USVᴴ::Tuple{…}, ΔUSVᴴ::Vector{…})
    @ MatrixAlgebraKit ~/.julia/packages/MatrixAlgebraKit/xv5dv/src/pullbacks/svd.jl:24
```

---

I think I have an idea of what is going on, I am pretty sure I can bypass the error but I'm not entirely sure everything will still be correct.
The main point is that we are currently checking that we are not keeping more singular values than the numerical rank of the matri.
This happens when `indU` or `indV` has more entries than the rank, throwing a `DimensionMismatch` for that (which is clearly not the most informative message either...).

In principle I might be able to alter the implementation to ignore that, but I'm not entirely sure if that is mathematically allowed, since this is basically one large degenerate subspace (nullspace) and I don't know if this would be correctly taking that into account.

@Jutho, I don't know if you have more insight into this?

Jutho:
This PR does a number of pullback-related things:

1. It moves the `remove_x_gauge_dependence!` functions from the tests to the main repository, so that they can be reused by higher level packages (TensorKit), and that it is easier to keep them in sync with the functions that test the gauge dependence.

2. Because testing the gauge dependence in the adjoints requires some of the same intermediate calculations as the actual computation of the pullback, and so some computations were performed twice before, there has been some restructuring resulting in a rename `check_x_cotangents` to `check_and_prepare_x_cotangents`.

3. The `svd_pullback` implementation has been fixed to accept `svd_full` pullbacks, and should also be more robust for pullbacks resulting from an `svd_trunc` with arbitrary `ind`.

4. The implementation of the `svd_trunc_pullback` that computes the pullback without depending on the full SVD of `A` has been changed, because the old one was reportedly (thanks @pbrehmer) very slow for larger matrices, and the Sylvester solver is not available on GPU hardware. Instead, the Sylvester equation is now explicitly solved as a geometric series. This requires another parameter, called `maxiter` (number of terms being kept in the geometric series). It also requires that it are the largest singular values that are being kept, as the convergence rate is essentially determined by the ratio "largest truncated singular value" / "smallest kept singular value". This implementation is only tested in the chainrules tests, where I already noticed that there were random cases where I needed quite a large number of iterations for convergence (100 was not enough, I bumped it to 1000 which does seem enough for the tests). This could be useful in combination with randomized svd, where because of oversampling you probably do have an explicit value for "largest truncated singular value", such that you can estimate the required number of iterations beforehand. 

Things to do:

* Actually test new `svd_trunc_pullback` on GPU in combination with randomized SVD

* The following suggestion could probably dramatically speed up convergence, as it changes from linear convergence to quadratic convergence (which is the typical rate at which dense eigenvalue decomposition / svd converges, basically the error is squared in every iteration and reaches sub-machine precision in a handful of iterations). I will try to implement this instead asap:

Screenshot 2026-04-18 at 00 59 26


Jutho:
Ok, I've now implemented the quadratic approach, and it is indeed much faster, converging in typically less than 10 iterations for double precision. It is called `svd_trunc_pullback2!`, but should replace `svd_trunc_pullback!`. I'd be interested in hearing about some real case timings from @pbrehmer 😄 .

pbrehmer:
Thanks for the upgraded pullbacks! I have some real case timings but using my own implementation where the Sylvester equation is solved using KrylovKit's `linsolve`. It still might be interesting to see since the scaling should be comparable, I think. My real test case is the fixed-point gradient linear problem of asymmetric CTMRG (without spatial symmetries) for the Fermi-Hubbard model. The benchmarks are run on multiple intermediately-optimized PEPS and then averaged. I plot the timings on a log-log scale as a function of $D^2\chi$ (the matrix that is decomposed is $D^2\chi \times D^2\chi$-dimensional) and normalize them by the CTMRG contraction times (CTMRG runs use a hot start):

image

It's pretty clear that `linsolve` outperforms the LAPACK `sylvester` solver already at moderate dimensions and that the scaling is much better. And it's also interesting to see that indeed the `sylvester` solver approximately has the same scaling as the CTMRG contraction step, i.e. the full SVD.

Jutho:
Which pullback is this now using? Is that the previous `svd_trunc_pullback!` (current main branch or latest release)? Or one of the two pullbacks `svd_trunc_pullback!` or `svd_trunc_pullback2!` in this branch?

Jutho:
Also, thanks for the very quick weekend response.

Jutho:
This PR seems to have broken the GPU extensions (both CUDA and AMD). Not sure what is causing this.

src/pullbacks/lq.jl
```diff
@@ -134,3 +133,51 @@ function lq_null_pullback!(
...
+`Q`. For the full LQ decomposition, the extra rows of `Q` beyond the rank `r` are not uniquely
+determined by `A`, so the corresponding part of `ΔQ` is projected to remove this ambiguity.
+Additionally, columns of `ΔL` beyond the rank are zeroed out.
+"""
+function remove_lq_gauge_dependence!(ΔL, ΔQ, A, L, Q; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(L))
+    r = MatrixAlgebraKit.lq_rank(L; rank_atol)
+    minmn = min(size(A)...)
+    Q₁ = view(Q, 1:r, :)
+    ΔQ₂ = view(ΔQ, (r + 1):minmn, :)
+    ΔQ₂ .= 0
```
lkdvos:
```suggestion
    zero!(ΔQ₂)
```

src/pullbacks/lq.jl
```diff
@@ -134,3 +133,51 @@ function lq_null_pullback!(
...
+    ΔQ₃ = view(ΔQ, (minmn + 1):size(ΔQ, 1), :) # extra rows in the case of lq_full
+    if r == minmn
+        ΔQ₃Q₁ᴴ = ΔQ₃ * Q₁'
+        mul!(ΔQ₃, ΔQ₃Q₁ᴴ, Q₁)
+    else # rank-deficient case, no gauge-invariant information
+        ΔQ₃ .= 0
```
lkdvos:
```suggestion
        zero!(ΔQ₃)
```

src/pullbacks/lq.jl
```diff
@@ -134,3 +133,51 @@ function lq_null_pullback!(
...
+    end
+    ΔL₂₂ = view(ΔL, (r + 1):size(ΔL, 1), (r + 1):minmn)
+    diagview(ΔL₂₂) .= 0
```
lkdvos:
```suggestion
    zero!(diagview(ΔL₂₂))
```

src/pullbacks/qr.jl
```diff
@@ -137,3 +137,50 @@ function qr_null_pullback!(
...
+`R`. For the full QR decomposition, the extra columns of `Q` beyond the rank `r` are not
+uniquely determined by `A`, so the corresponding part of `ΔQ` is projected to remove this
+ambiguity. Additionally, rows of `ΔR` beyond the rank are zeroed out.
+"""
+function remove_qr_gauge_dependence!(ΔQ, ΔR, A, Q, R; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(R))
+    r = MatrixAlgebraKit.qr_rank(R; rank_atol)
+    minmn = min(size(A)...)
+    Q₁ = view(Q, :, 1:r)
+    ΔQ₂ = view(ΔQ, :, (r + 1):minmn)
+    ΔQ₂ .= 0
```
lkdvos:
```suggestion
    zero!(ΔQ₂)
```

src/pullbacks/qr.jl
```diff
@@ -137,3 +137,50 @@ function qr_null_pullback!(
...
+    ΔQ₃ = view(ΔQ, :, (minmn + 1):size(ΔQ, 2)) # extra columns in the case of qr_full
+    if r == minmn # full rank case, ΔQ₃ contains gauge-invariant information along Q₁
+        Q₁ᴴΔQ₃ = Q₁' * ΔQ₃
+        mul!(ΔQ₃, Q₁, Q₁ᴴΔQ₃)
+    else # rank-deficient case, no gauge-invariant information
+        ΔQ₃ .= 0
```
lkdvos:
```suggestion
        zero!(ΔQ₃)
```

src/pullbacks/qr.jl
```diff
@@ -137,3 +137,50 @@ function qr_null_pullback!(
...
+    end
+    ΔR₂₂ = view(ΔR, (r + 1):minmn, (r + 1):size(R, 2))
+    diagview(ΔR₂₂) .= 0
```
lkdvos:
```suggestion
    zero!(diagview(ΔR₂₂))
```

src/pullbacks/qr.jl
```diff
@@ -137,3 +137,50 @@ function qr_null_pullback!(
...
+    view(ΔR₂₂, uppertriangularind(ΔR₂₂)) .= 0
```
lkdvos:
```suggestion
    zero!(view(ΔR₂₂, uppertriangularind(ΔR₂₂)))
```

src/pullbacks/svd.jl
```diff
@@ -34,83 +130,46 @@ function svd_pullback!(
...
-    ΔA = mul!(ΔA, Ur, UdΔAV * Vᴴr, 1, 1) # add the contribution to ΔA
+    ΔA = mul!(ΔA, U₁, UdΔAV * V₁ᴴ, 1, 1) # add the contribution to ΔA
 
     # Add the remaining contributions
-    if m > r && !iszerotangent(ΔU) # remaining ΔU is already orthogonal to Ur
-        Sp = view(S, indU)
-        Vᴴp = view(Vᴴ, indU, :)
-        ΔA = mul!(ΔA, ΔU ./ Sp', Vᴴp, 1, 1)
+    if m > r && !iszerotangent(ΔU₁) # ΔU₁ is already orthogonal to U₁
+        ΔA = mul!(ΔA, ΔU₁ ./ S₁', V₁ᴴ, 1, 1)
```
```suggestion
        ΔU₁ ./= S₁'
        ΔA = mul!(ΔA, ΔU₁, V₁ᴴ, 1, 1)
```
I think here you created a copy for `ΔU₁` so this can be safely overwritten

src/pullbacks/svd.jl
```diff
@@ -34,83 +130,46 @@ function svd_pullback!(
...
     end
-    if n > r && !iszerotangent(ΔVᴴ) # remaining ΔV is already orthogonal to Vᴴr
-        Sp = view(S, indV)
-        Up = view(U, :, indV)
-        ΔA = mul!(ΔA, Up, Sp .\ ΔVᴴ, 1, 1)
+    if n > r && !iszerotangent(ΔV₁ᴴ) # ΔV₁ᴴ is already orthogonal to V₁ᴴ
+        ΔA = mul!(ΔA, U₁, S₁ .\ ΔV₁ᴴ, 1, 1)
```
```suggestion
        ΔV₁ᴴ .= S₁ .\ ΔV₁ᴴ
        ΔA = mul!(ΔA, U₁, ΔV₁ᴴ, 1, 1)
```
Same comment here (I think)

src/pullbacks/svd.jl
```diff
@@ -158,64 +218,120 @@ function svd_trunc_pullback!(
...
+        diagview(UdΔAV) .+= real.(ΔS)
     end
-    if !iszerotangent(ΔVᴴ)
-        (p, n) == size(ΔVᴴ) || throw(DimensionMismatch())
-        mul!(VΔV, Vᴴ, ΔVᴴ')
-        # ΔVᴴ -= VΔVp' * Vᴴr but one less allocation without overwriting ΔVᴴ
-        ΔVᴴ = mul!(copy(ΔVᴴ), VΔV', Vᴴ, -1, 1)
+    ΔA = mul!(ΔA, U, UdΔAV * Vᴴ, 1, 1) # add the contribution to ΔA
+
+    # The contribtutions from the orthogonal complement need to be treated differently
```
```suggestion
    # The contributions from the orthogonal complement need to be treated differently
```

pbrehmer:
> Which pullback is this now using? Is that the previous `svd_trunc_pullback!` (current main branch or latest release)? Or one of the two pullbacks `svd_trunc_pullback!` or `svd_trunc_pullback2!` in this branch?

It was using the `svd_trunc_pullback!` from the latest release (`v0.6.5`). What I should mention again is that `linsolve` will only converge this fast if it is preconditioned properly with the inverse singular values. Here are the lines of code which I changed in `svd_trunc_pullback!` that implement this preconditioning:

```julia
# replace XY = _sylvester(ÃÃ, -Smat, rhs) with linsolve
Smat⁻¹ = diagm(inv_safe.(S, degeneracy_atol))
f(xy) = ÃÃ * xy * Smat⁻¹ - xy
XY₀ = zeros(scalartype(ÃÃ), size(ÃÃ, 2), size(Smat⁻¹, 1))
XY, info = linsolve(f, -rhs * Smat⁻¹, XY₀, solver_alg)
```

src/pullbacks/svd.jl
```diff
@@ -22,9 +118,9 @@ Adds the pullback from the SVD of `A` to `ΔA` given the output `USVᴴ` of `svd
 In particular, it is assumed that `A ≈ U * S * Vᴴ`, or thus, that no singular values with
 magnitude less than `rank_atol` are missing from `S`.  For the cotangents, an arbitrary
 number of singular vectors or singular values can be missing, i.e. for a matrix `A` with
-size `(m, n)`, `ΔU` and `ΔVᴴ` can have sizes `(m, pU)` and `(pV, n)` respectively, whereas
-`diagview(ΔS)` can have length `pS`. In those cases, additionally `ind` is required to
-specify which singular vectors and values are present in `ΔU`, `ΔS` and `ΔVᴴ`.
+size `(m, n)`, `ΔU` and `ΔVᴴ` can have sizes `(m, p)` and `(p, n)` respectively, whereas
+`diagview(ΔS)` can have length `p`. In those cases, an additional list `ind` of length `p`
```
kshyatt: Could we provide an example?

Jutho:
> It was using the `svd_trunc_pullback!` from the latest release (`v0.6.5`). What I should mention again is that `linsolve` will only converge this fast if it is preconditioned properly with the inverse singular values. Here are the lines of code which I changed in `svd_trunc_pullback!` that implement this preconditioning:

@pbrehmer , would it be much work for you to run the same benchmark with `svd_trunc_pullback!` and `svd_trunc_pullback2!` from this branch (especially the latter) and add the result to that plot?

For very large matrix size, I agree that Krylov will still be the preferred option, but also the `svd_trunc` itself should be done with Krylov methods then. It is also a fact that if you did compute the full svd, then simply `svd_pullback!` will still be the fastest. So I think these `svd_trunc_pullback(2)!` methods really only make sense in combination with something like randomized svd, but still it would be interesting to see the relative performance of each of them.

src/pullbacks/lq.jl
```diff
@@ -134,3 +133,51 @@ function lq_null_pullback!(
...
+    view(ΔL₂₂, lowertriangularind(ΔL₂₂)) .= 0
```
Jutho:
```suggestion
    zero!(view(ΔL₂₂, lowertriangularind(ΔL₂₂)))
```

pbrehmer:
> @pbrehmer , would it be much work for you to run the same benchmark with `svd_trunc_pullback!` and `svd_trunc_pullback2!` from this branch (especially the latter) and add the result to that plot?

I can do that but currently my computational resources on the cluster are blocked by some other simulations so it will take a bit!

> For very large matrix size, I agree that Krylov will still be the preferred option, but also the `svd_trunc` itself should be done with Krylov methods then. It is also a fact that if you did compute the full svd, then simply `svd_pullback!` will still be the fastest. So I think these `svd_trunc_pullback(2)!` methods really only make sense in combination with something like randomized svd, but still it would be interesting to see the relative performance of each of them.

Yes indeed. For these benchmarks I explicitly wanted to check the performance of the `trunc` pullbacks but it was still more convenient to just compute the forward using a full SVD.

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
+    ΔU, ΔS, ΔVᴴ, aUᴴΔU, aVᴴΔV = check_and_prepare_svd_cotangents(
+        U, S, Vᴴ, ΔU, ΔSmat, ΔVᴴ, p; degeneracy_atol, gauge_atol
+    )
+
+    # This part is the same as in `svd_pullback!`
+    UdΔAV = (aUᴴΔU .+ aVᴴΔV) .* inv_safe.(S' .- S, degeneracy_atol) .+
+        (aUᴴΔU .- aVᴴΔV) .* inv_safe.(S' .+ S, degeneracy_atol)
+    if !iszerotangent(ΔS)
+        diagview(UdΔAV) .+= real.(ΔS)
     end
```
lkdvos: Given that it is the same, can we refactor this out into a separate function? I'm not sure if it is best/easiest/most convenient to only do these couple lines, or the entire non-orthogonal complement part

Also, am I correct that we don't actually need the `aUᴴΔU` and/or `aVᴴΔV` anymore, so we could possibly reuse that memory? Not sure if this really ends up mattering, but in any case it's convenient to have this in a separate function to see the different parts show up in profilers.

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
-        (p, n) == size(ΔVᴴ) || throw(DimensionMismatch())
-        mul!(VΔV, Vᴴ, ΔVᴴ')
-        # ΔVᴴ -= VΔVp' * Vᴴr but one less allocation without overwriting ΔVᴴ
-        ΔVᴴ = mul!(copy(ΔVᴴ), VΔV', Vᴴ, -1, 1)
+    ΔA = mul!(ΔA, U, UdΔAV * Vᴴ, 1, 1) # add the contribution to ΔA
+
+    # The contributions from the orthogonal complement need to be treated differently
+    # ΔU and ΔVᴴ are already orthogonal to U and Vᴴ
+    if !(iszerotangent(ΔU) && iszerotangent(ΔVᴴ))
+        Aperp = A - U * Smat * Vᴴ
```
```suggestion
        Aperp = mul!(copy(A), U, Smat * Vᴴ, -1, 1)
```

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
+        x₀ = iszerotangent(ΔU) ? zero(U) : rdiv!(ΔU, Diagonal(S))
+        y₀ᴴ = iszerotangent(ΔVᴴ) ? zero(Vᴴ) : ldiv!(Diagonal(S), ΔVᴴ)
```
lkdvos: I guess here we are overwriting `ΔU` and `ΔVᴴ`, which are input tangents, and I think we decided on not risking that we are allowed to alter them?

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
-    end
-    if !iszerotangent(ΔVᴴ)
-        mul!(view(rhs, m̃ .+ (1:ñ), :), Ṽᴴ, ΔVᴴ')
-    else
-        fill!(view(rhs, m̃ .+ (1:ñ), :), 0)
+    # The contribtutions from the orthogonal complement need to be treated differently
+    # ΔU and ΔVᴴ are already orthogonal to U and Vᴴ
+    if !(iszerotangent(ΔU) && iszerotangent(ΔVᴴ))
+        X₀ = iszerotangent(ΔU) ? zero(U) : rdiv!(ΔU, Diagonal(S))
+        Y₀ᴴ = iszerotangent(ΔVᴴ) ? zero(Vᴴ) : ldiv!(Diagonal(S), ΔVᴴ)
```
lkdvos: Same comment about overwriting input tangents

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
+        AP = A - U * Smat * Vᴴ
```
```suggestion
        AP = mul!(copy(A), U, Smat * Vᴴ, -1, 1)
```

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
+        AP ./= S[1]
+        S = S ./ S[1]
```
```suggestion
        S ./= S[1]
```

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
+        X₁ = X₀ + rdiv!(AP * Y₀ᴴ', Diagonal(S))
```
```suggestion
        X₁ = rdiv!(AP * Y₀ᴴ', Diagonal(S))
        X₁ .+= X₀
```

src/pullbacks/svd.jl
```diff
@@ -158,64 +220,120 @@ function svd_trunc_pullback!(
...
+        Y₁ᴴ = Y₀ᴴ + ldiv!(Diagonal(S), X₀' * AP)
```
```suggestion
        Y₁ᴴ = ldiv!(Diagonal(S), X₀' * AP)
        Y₁ᴴ .+= Y₀ᴴ
```

src/pullbacks/svd.jl
```diff
@@ -253,3 +371,52 @@ function svd_vals_pullback!(
...
+    ΔU₁ = view(ΔU, :, 1:r)
+    ΔVᴴ₁ = view(ΔVᴴ, 1:r, :)
+    Sdiag = diagview(S)
+    gaugepart = mul!(U₁' * ΔU₁, Vᴴ₁, ΔVᴴ₁', true, true)
+    gaugepart = project_antihermitian!(gaugepart)
+    gaugepart[abs.(transpose(Sdiag) .- Sdiag) .>= degeneracy_atol] .= 0
+    mul!(ΔU₁, U₁, gaugepart, -1, 1)
+    if size(ΔU, 2) > r
+        if r < length(Sdiag) # rank-deficient case, no stable information can be extracted from extra columns of U
+            ΔU[:, (r + 1):end] .= 0
```
```suggestion
            zero!(ΔU[:, (r + 1):end])
```

src/pullbacks/svd.jl
```diff
@@ -253,3 +371,52 @@ function svd_vals_pullback!(
...
+        else # the component of ΔU₂ along U₁ contains gauge-invariant information
+            p = size(ΔU, 2)
+            ΔU₂ = view(ΔU, :, (r + 1):p)
+            U₁ᴴΔU₂ = U₁' * ΔU₂
+            mul!(ΔU₂, U₁, U₁ᴴΔU₂)
+        end
+    end
+    if size(ΔVᴴ, 1) > r
+        if r < length(Sdiag) # rank-deficient case, no stable information can be extracted from extra rows of Vᴴ
+            ΔVᴴ[(r + 1):end, :] .= 0
```
```suggestion
            zero!(ΔVᴴ[(r + 1):end, :])
```

lkdvos (commented):
It is probably also worth it to add the `remove_f_gauge_dependence` functions to the `public` list, given that we explicitly intend them to be used by TensorKit.

Overall looks like a great PR though, this should hopefully stabilize some of the issues we've been having 🥳 

Some remaining question that definitely don't have to be addressed here but is worth bringing up:
- Is it worth it to refactor the `sylvester` solver into its own function, or is this one too hand-crafted for this specific purpose? This might make it more convenient to swap out the Krylov-based solver, but I don't mean to add too much burden to this PR/this package for that either.

src/pullbacks/svd.jl
Jutho: You mean like a code example in the doc string? I always have to look up the syntax for that 😄 .

src/pullbacks/svd.jl
Jutho: I assume we can move this part also into the `check_and_prepare_svd_cotangents` function, and then only return `UdΔAV` instead of `aUᴴΔU` and `aVᴴΔV` separately (I don't think their memory can be reused later on).

src/pullbacks/svd.jl
Jutho: These are the same variables as `ΔU₁` in the `svd_pullback!` (without trunc), i.e. they are already copies of the original cotangents.

src/pullbacks/svd.jl
kshyatt: Yes, please 🥺

src/pullbacks/svd.jl
Jutho: Here, that would actually overwrite `S`, which is potentially not safe (although pullbacks should be calculated after the original function output has been used, so probably it is harmless). Nonetheless, so far we have been consistent in not modifying `A` or any of its factors in the pullback functions.

Jutho:
> * Is it worth it to refactor the `sylvester` solver into its own function, or is this one too hand-crafted for this specific purpose? This might make it more convenient to swap out the Krylov-based solver, but I don't mean to add too much burden to this PR/this package for that either.

I think it is pretty specific; where else do you think we might use this?

lkdvos:
I really didnt look at this carefully, as in, is this just a regular Sylvester silver? In that case it shows up in the Clebsch-Gordan equation for finite groups

src/pullbacks/svd.jl
```diff
@@ -1,19 +1,115 @@
...
+        VᴴΔV₁ = V₁ᴴ * ΔV₁ᴴ'
+        ΔV₁ᴴ = mul!(ΔV₁ᴴ, VᴴΔV₁', V₁ᴴ, -1, 1)
+        aVᴴΔV₁ = project_antihermitian!(VᴴΔV₁)
+        Δgauge = max(Δgauge, ΔgaugeV)
+    else
+        ΔV₁ᴴ = nothing
+        aVᴴΔV₁ = zero!(similar(V₁ᴴ, (r, r)))
+    end
+    mask = abs.(S₁' .- S₁) .< degeneracy_atol
+    Δgauge = max(Δgauge, norm(view(aUᴴΔU₁, mask) + view(aVᴴΔV₁, mask), Inf))
```
```suggestion
    bc = Base.broadcasted(S₁', S₁, aUᴴΔU₁, aVᴴΔV₁) do s1, s2, u, v
        return abs(s1 - s2) < degeneracy_atol ? zero(u) + zero(v) : u + v
    end
    Δgauge = max(Δgauge, norm(bc, Inf))
```
Fun trick I learned about a while ago, honestly at this point I'm not sure if this is really worth it, but it does remove two more allocations (1 for the mask and 1 for the sum)

src/pullbacks/svd.jl
lkdvos: You are right, I missed this. I agree that it's better to keep this separate.

Jutho:
> I really didnt look at this carefully, as in, is this just a regular Sylvester silver? In that case it shows up in the Clebsch-Gordan equation for finite groups

No not at all; it uses that one of the two matrices is diagonal (or more generally, easily invertible), and furthermore assumes that the smallest singular value of that matrix is larger than the largest singular value of the other matrix (or more generally: `opnorm(inv(B)) * opnorm(A) < 1`)

Jutho:
I am happy to discuss the CG of finite group case though, maybe there are some other tricks that can be used.

Jutho:
Enzyme tests seem to be taken an extreme amount of time / GC / allocations. Not sure if this is a consequence of the changes here, I will have to check.

kshyatt:
> Enzyme tests seem to be taken an extreme amount of time / GC / allocations. Not sure if this is a consequence of the changes here, I will have to check.

I think it's a consequence of how the Enzyme tests work. Basically, they use `autodiff_thunk` to generate the functions for forward and reverse pass, then carefully check all the memory properties and correctness. This is great and very thorough but also expensive, so there's a large compilation burden associated with each test.

kshyatt:
GPU CI now passing thanks to a Buildkite fix.

Jutho:
Great. Not sure why it is failing on Ubuntu latest; probably timeout? I will already swap `svd_trunc_pullback!` with `svd_trunc_pullback2!`, as the latter should be much faster (even though these are only tested in the chainrules tests if I am correct).

src/pullbacks/svd.jl
Jutho: Should be resolved now.

src/pullbacks/svd.jl
Jutho: We don't really have tradition of code examples in MatrixAlgebraKit, so it feels a bit strange to add this only here. Is there any part of this docstring that could be improved for clarity?

kshyatt:
TBH I think the speed of the pullback itself won't change things much, it's entirely compilation time in `autodiff_thunk` in `Enzyme.Compiler` and in the GC

src/pullbacks/svd.jl
Jutho: Oh I had forgotten to include this one.

src/pullbacks/svd.jl
kshyatt: I'd also say that "`diagview(ΔS)` can have length `p`" might be more natural as "`diagview(ΔS)` must have length `p`". "In those cases" is also a bit confusing -- in which cases? The cases when `p =/= min(m, n)`?

Sometimes I find an example is clearest but maybe saying something like

"an additional list `ind` of length `p` is required to indicate the singular values (`diagview(ΔS)[ind]`) and vectors (`ΔU[:, ind]`, `ΔVᴴ[ind, :`]`) which are present"?

kshyatt:
Just as a note: this should close https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/150, right?

src/pullbacks/svd.jl
lkdvos: To be fair, that's also slightly confusing, since the `ΔS` etc here already refer to the selected ones. I think that if we really want to make this clear, we should try and do a fuller explanation in the docs somewhere, but I'd prefer not to hold up this PR for that

kshyatt: This is why I thought a doctest would be helpful 😹 . But it can wait for a different PR.

lkdvos (approved):
@Jutho I rebased on top of the latest main, and added the gauge dependence removal to the `public` list. I don't think the performance of the new SVD pullback is required to already merge this, mostly to have the fix of the full SVD out as soon as possible?

Jutho:
Ok. It wasn't really my intention to keep `svd_trunc_pullback2!` around after some real-world benchmarks confirmed that it is indeed much slower. So the question is whether we merge as is and remove it later? As long as it is not marked public, that should be fine without causing breaking changes, right.

src/pullbacks/svd.jl
```diff
@@ -22,9 +127,9 @@ Adds the pullback from the SVD of `A` to `ΔA` given the output `USVᴴ` of `svd
 In particular, it is assumed that `A ≈ U * S * Vᴴ`, or thus, that no singular values with
 magnitude less than `rank_atol` are missing from `S`.  For the cotangents, an arbitrary
 number of singular vectors or singular values can be missing, i.e. for a matrix `A` with
-size `(m, n)`, `ΔU` and `ΔVᴴ` can have sizes `(m, pU)` and `(pV, n)` respectively, whereas
-`diagview(ΔS)` can have length `pS`. In those cases, additionally `ind` is required to
-specify which singular vectors and values are present in `ΔU`, `ΔS` and `ΔVᴴ`.
+size `(m, n)`, `ΔU` and `ΔVᴴ` can have sizes `(m, p)` and `(p, n)` respectively, whereas
+`diagview(ΔS)` can have length `p`. In those cases, an additional list `ind` of length `p`
+is required to specify which singular vectors and values are present in `ΔU`, `ΔS` and `ΔVᴴ`.
```
Jutho: Let me know if this is more clear, and if so, feel free to commit this suggestion.

```suggestion
size `(m, n)`, `ΔU`, `ΔS` and `ΔVᴴ` can have sizes `(m, p)`, `(p, p)` and `(p, n)` respectively
and the argument `ind` is a list of length `p` indicating that these are cotangents corresponding to `U[:, ind]`, `S[ind, ind]` and `Vᴴ[ind, :]`,
whereas cotangents with respect to the other rows and columns are zero.
If `ind` is not present, `ΔU`, `ΔS` and `ΔVᴴ` are assumed to have the same size as `U`, `S` and `Vᴴ` respectively.
```

Jutho:
One final change could be to replace `Sinv` with `S⁻¹` in `svd_trunc_pullback!`.

PR #211, opened 2026-04-21
Add `==` for `TruncationStrategy`s

sanderdemeyer:
This addresses QuantumKitHub/TensorKit.jl#407 and adds an equality check to `TruncationStrategy`s.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -541,5 +541,5 @@ macro check_size(x, sz, size = :size)
 end
 
 # Check equality of two `TruncationStrategy`s
-isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
+isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(T))
 isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -539,3 +539,6 @@ macro check_size(x, sz, size = :size)
         end
     )
 end
+
+# Check equality of two `TruncationStrategy`s
+isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = (T1 == T2) && all([getfield(t1, f) == getfield(t2, f) for f in fieldnames(typeof(t1))])
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -541,5 +541,5 @@ macro check_size(x, sz, size = :size)
 end
 
 # Check equality of two `TruncationStrategy`s
-isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(getfield(t1, f) == getfield(t2, f) for f in fieldnames(typeof(t1)))
+isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
 isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -541,4 +541,5 @@ macro check_size(x, sz, size = :size)
 end
 
 # Check equality of two `TruncationStrategy`s
-isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = (T1 == T2) && all([getfield(t1, f) == getfield(t2, f) for f in fieldnames(typeof(t1))])
+isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(getfield(t1, f) == getfield(t2, f) for f in fieldnames(typeof(t1)))
+isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -541,5 +541,5 @@ macro check_size(x, sz, size = :size)
 end
 
 # Check equality of two `TruncationStrategy`s
-isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(T))
-isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
+Base.:(==)(t1::T, t2::T) where {T <: TruncationStrategy} = all(getfield(t1, f) == getfield(t2, f) for f in fieldnames(T))
+Base.:(==)(t1::T1, t2::T2) where {T1 <: TruncationStrategy, T2 <: TruncationStrategy} = false
--- a/test/algorithms.jl
+++ b/test/algorithms.jl
@@ -60,3 +60,22 @@ end
         @test @constinferred(select_algorithm(svd_compact!, A, $alg)) === QRIteration()
     end
 end
+
+@testset "Truncation equivalencies" begin
+    truncs = [notrunc(),
+    truncrank(4),
+    truncrank(5),
+    truncfilter(x -> x > 0),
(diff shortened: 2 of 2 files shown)


src/algorithms.jl
```diff
@@ -539,3 +539,6 @@ macro check_size(x, sz, size = :size)
         end
     )
 end
+
+# Check equality of two `TruncationStrategy`s
+isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = (T1 == T2) && all([getfield(t1, f) == getfield(t2, f) for f in fieldnames(typeof(t1))])
```
kshyatt: There might be a way to do this without allocating a `Vector`? Surely most of these should fit in a stack-allocated `Tuple`, no?

kshyatt: For the `T1 == T2` check, why not do

```julia
isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all([getfield(t1, f) == getfield(t2, f) for f in fieldnames(T)])
isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
```

sanderdemeyer:
There is another issue with this code, namely
```julia
t1 = truncfilter(x -> x > 0)
t2 = truncfilter(x -> x > 0)
t1 == t1 # true
t1 == t2 # false
```
which is due to the presence of an anonymous function. I don't know whether there is a way around this.

Jutho:
> There is another issue with this code, namely
> 
> ```julia
> t1 = truncfilter(x -> x > 0)
> t2 = truncfilter(x -> x > 0)
> t1 == t1 # true
> t1 == t2 # false
> ```
> 
> which is due to the presence of an anonymous function. I don't know whether there is a way around this.

I don't think that will have an easy fix.

src/algorithms.jl
```diff
@@ -539,3 +539,7 @@ macro check_size(x, sz, size = :size)
         end
     )
 end
+
+# Check equality of two `TruncationStrategy`s
+isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(getfield(t1, f) == getfield(t2, f) for f in fieldnames(typeof(t1)))
```
Jutho:
```suggestion
isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
```

src/algorithms.jl
sanderdemeyer: I changed it now to
```julia
isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
```
Or would you prefer to do it subtype per subtype, and do something like this?
```julia
isequal(t1::TruncationByOrder, t2::TruncationByOrder) = (t1.howmany == t2.howmany) && (howmany.by == howmany.by) && (t1.rev == t2.rev)
```

src/algorithms.jl
```diff
@@ -539,3 +539,7 @@ macro check_size(x, sz, size = :size)
         end
     )
 end
+
+# Check equality of two `TruncationStrategy`s
+isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
+isequal(t1::T1, t2::T2) where {T1, T2 <: TruncationStrategy} = false
```
lkdvos: I'm not sure if this really matters, but there are some weird edge-cases you can consider equal even if they have different types, e.g.
```julia
julia> 0.5 == Float32(0.5)
true

julia> isequal(0.5, Float32(0.5))
true
```
This makes it such that tolerances that could compare as equal now don't:

```julia
julia> t1 = trunctol(; atol = 0.5)
MatrixAlgebraKit.TruncationByValue{Float64, Int64, typeof(abs)}(0.5, 0.0, 2, abs, false)

julia> t2 = trunctol(; atol = Float32(0.5))
MatrixAlgebraKit.TruncationByValue{Float32, Int64, typeof(abs)}(0.5f0, 0.0f0, 2, abs, false)

julia> all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(MatrixAlgebraKit.TruncationByValue))
true
```


I'm not sure there is a great solution to this that does not consist of manually defining `isequal` for the various strategies, and I'm also not sure if we should care?

src/algorithms.jl
```diff
@@ -539,3 +539,7 @@ macro check_size(x, sz, size = :size)
         end
     )
 end
+
+# Check equality of two `TruncationStrategy`s
+isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
```
lkdvos:
```suggestion
isequal(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(T))
```

This doesn't matter but looks a little less asymmetric to me :)

lkdvos:
Something else to investigate is that by choosing `isequal`, I think you must also guarantee that `isequal(x, y)` implies `isequal(hash(x), hash(y))`, since that is the invariant needed to make dictionary-like objects work. I'm not sure if this implementation actually has that, so it might be better to use `==`?

src/algorithms.jl
sanderdemeyer: It's already a bit weird that you would compare `TruncationStrategy`s with different scalartypes, but since these will be used in inequalities, it seems good to me that `isequal` on two `TruncationStrategy`s with the same physical effect yields `true`.

sanderdemeyer:
> Something else to investigate is that by choosing `isequal`, I think you must also guarantee that `isequal(x, y)` implies `isequal(hash(x), hash(y))`, since that is the invariant needed to make dictionary-like objects work. I'm not sure if this implementation actually has that, so it might be better to use `==`?

I don't know whether there was a specific reason to change `==` to `isequal` in the suggestion of @Jutho?

lkdvos:
> I don't know whether there was a specific reason to change == to isequal in the suggestion of @Jutho?

I think you started with defining `Base.isequal`, and then @Jutho suggested to also use that in the implementation?

sanderdemeyer:
Oh okay, now I get what you meant. I first wanted to implement it like
```julia
Base.:(==)(t1::T, t2::T) where {T <: TruncationStrategy} = all(isequal(getfield(t1, f), getfield(t2, f)) for f in fieldnames(typeof(t1)))
```
but then I got an error in the code quality, saying
```bash
Output generated during execution of 'codequality':
┌ 2 ambiguities found. To get a list, set `broken = false`.
│ Ambiguity #1
│ ==(t1::T1, t2::T2) where {T1, T2<:TruncationStrategy} @ MatrixAlgebraKit ~/Documents/Doctoraat/MatrixAlgebraKit.jl/src/algorithms.jl:545
│ ==(::Missing, ::Any) @ Base missing.jl:75
│ 
│ Possible fix, define
│   ==(::Missing, ::T2) where T2<:MatrixAlgebraKit.TruncationStrategy
│ 
│ Ambiguity #2
│ ==(t1::T1, t2::T2) where {T1, T2<:TruncationStrategy} @ MatrixAlgebraKit ~/Documents/Doctoraat/MatrixAlgebraKit.jl/src/algorithms.jl:545
│ ==(w::WeakRef, v) @ Base gcutils.jl:35
│ 
│ Possible fix, define
│   ==(::WeakRef, ::T2) where T2<:MatrixAlgebraKit.TruncationStrategy
└ 
```
which was not present when I changed it to `isequal`.

Jutho:
That error comes from the fact that `{T1, T2<:TruncationStrategy}` does not mean that `T1<:TruncationStrategy`, so effectively it is `T1<:Any`. You probably intended to have:

`{T1<:TruncationStrategy, T2<:TruncationStrategy}`

And then indeed let's use `Base.:(==)`.

sanderdemeyer:
Oh yes, completely right. I changed this, and moved it back to `Base.:(==)`.

lkdvos (approved):

PR #215, opened 2026-04-22
Use a test setup matrix like TensorKit

kshyatt:
Should give us some benefit from draft PRs and splitting the tests into smaller chunks


kshyatt:
Probably need a way to turn off the Mac + Enzyme combo which is known to be bad due to underpowered mac mini runners

.github/workflows/Tests.yml
```diff
@@ -9,33 +9,77 @@ on:
...
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review, converted_to_draft]
   workflow_dispatch:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   # Cancel intermediate builds: only if it is a pull request build.
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
-
+
```
lkdvos:
```suggestion

```

.github/workflows/Tests.yml
```diff
@@ -9,33 +9,77 @@ on:
...
+            version: 1
+            group: enzyme 
+          - os: macos-latest
+            version: "lts" 
+            group: enzyme 
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 120
+    steps:
+      - uses: actions/checkout@v6
+      - uses: julia-actions/setup-julia@v2
```
lkdvos:
```suggestion
      - uses: julia-actions/setup-julia@v3
```

.github/workflows/Tests.yml
```diff
@@ -9,33 +9,77 @@ on:
...
+    outputs:
+      groups: ${{ steps.mk.outputs.groups }}
+      version: ${{ steps.mk.outputs.version }}
+      os: ${{ steps.mk.outputs.os }}
+    steps:
+      - uses: actions/checkout@v6
+      - id: mk
+        shell: bash
+        run: |
+          # Auto-discover test groups from all subdirectory names.
```
lkdvos, line 31: Do you know if we will bump into issues with the `testsuite` subdirectory?

lkdvos (commented):
I definitely agree that we should start using similar setups in many places, but it could be nice if we could uniformize this and use a shared action/test setup, to keep this maintainable in the future?

kshyatt:
>  it could be nice if we could uniformize this and use a shared action/test setup, to keep this maintainable in the future?

Sure. I set this up here because the Enzyme tests are impacting other people. If someone wants to set up the shared workflow now, I don't see a problem, otherwise we can do this now and then remove it in another PR later once the shared one is set up?

.github/workflows/Tests.yml
kshyatt: It seems what happens is it loads all the packages, notices there are no tests to run, and marks everything good. I can filter out the `testsuite` ones anyway

lkdvos: It's probably better to filter it out here directly, otherwise this will show up as a separate step on every PR run I think?

kshyatt: Yeah I agree re filtering, I'll try to modify the regex

lkdvos: Sorry to do this to you

kshyatt: 1) I'm being punished for my many sins
2) regex is one situation i'll never feel bad about asking my robot friend 😉

kshyatt: I *think* this got resolved?

test/common/utilities.jl
lkdvos, line 1: I think this one now has to be deleted from the `ParallelTestRunner` suite as well, since the path has changed.

test/common/codequality.jl
lkdvos, line 1: It might not be worth it to spin up an entire new set of CI actions for just this test, so I would say to either put this in the `common` group, or put all of the `common` stuff in the `misc` group?

.github/workflows/Tests.yml
```diff
@@ -9,33 +9,76 @@ on:
...
+            | jq -R -s -c '[split("\n")[] | select(length > 0) | select(.!= "testsuite")]')
+
+          echo "groups=${groups}" >> "$GITHUB_OUTPUT"
+
+          # Draft PR: only run ubuntu-latest + version=1
+          if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.event.pull_request.draft }}" == "true" ]]; then
+            echo 'version=["1"]' >> "$GITHUB_OUTPUT"
+            echo 'os=["ubuntu-latest"]' >> "$GITHUB_OUTPUT"
+          else
+            echo 'version=["lts","1"]' >> "$GITHUB_OUTPUT"
```
lkdvos:
```suggestion
            echo 'version=["min","1"]' >> "$GITHUB_OUTPUT"
```

.github/workflows/Tests.yml
```diff
@@ -9,33 +9,76 @@ on:
...
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+        version: ${{ fromJSON(needs.setup-matrix.outputs.version) }}
+        os: ${{ fromJSON(needs.setup-matrix.outputs.os) }}
+        group: ${{ fromJSON(needs.setup-matrix.outputs.groups) }}
+        exclude:
+          - os: macos-latest
+            version: 1
+            group: enzyme 
+          - os: macos-latest
+            version: "lts"
```
lkdvos:
```suggestion
            version: "min" 
```

kshyatt:
Is this good to go if tests pass?

lkdvos (approved):
Good for me!

(Re tests and rebasing this, lets first merge #217 once the tests pass of that one, then rebase this on top. I've cancelled the tests for now since we are oversubscribing our CI workers, and this can then rerun on top of that?)

kshyatt:
Sounds good, sorry for rebasing so much stuff here and at TK 😅

kshyatt:
Once this is in I can try https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/179 on top of it, it should allow the AD tests to actually finish...

PR #217, opened 2026-04-24
Add precompile workload

lkdvos:
I'm not sure how we feel about this, and how much of this is really useful for downstream packages, but this seems like a reasonable workload for things to precompile that does not take too long (~24seconds on my mac).


Jutho (approved):

lkdvos:
Failure is unrelated, will merge here.

PR #221, opened 2026-04-28
make `ind` functions consistent in gauge removal

lkdvos:
Bumped into this while trying to port this over to TensorKit:
Other parts of the code use `ind = Colon()` as default, and then first take `indV = axes(V, 2)[ind]` before doing the check of the length (`length(ind)` fails if `ind = Colon()`).
Here I just did the same.

I checked for the SVD gauge removal and was surprised that there is no `ind` argument there, @Jutho is this meant to be like this?

### DIFF
--- a/src/pullbacks/eig.jl
+++ b/src/pullbacks/eig.jl
@@ -211,11 +211,11 @@ across eigenvectors associated with degenerate eigenvalues), so the correspondin
 `ΔV` are projected out.
 """
 function remove_eig_gauge_dependence!(
-        ΔV, D, V, ind = axes(ΔV, 2);
+        ΔV, D, V, ind = Colon();
         degeneracy_atol = MatrixAlgebraKit.default_pullback_gauge_atol(D)
     )
-    length(ind) == size(ΔV, 2) || throw(DimensionMismatch())
     indV = axes(V, 2)[ind]
+    length(indV) == size(ΔV, 2) || throw(DimensionMismatch("Incompatible size of selected `ind` and `ΔV`"))
     Vp = view(V, :, indV)
     Ddiag = view(diagview(D), indV)
     gaugepart = Vp' * ΔV
--- a/src/pullbacks/eig.jl
+++ b/src/pullbacks/eig.jl
@@ -1,13 +1,53 @@
-function check_eig_cotangents(
-        D, VᴴΔV;
-        degeneracy_atol::Real = default_pullback_rank_atol(D),
-        gauge_atol::Real = default_pullback_gauge_atol(VᴴΔV)
+function check_and_prepare_eig_cotangents(
+        D, V, ViG, ΔDmat, ΔV, ind = Colon();
+        degeneracy_atol::Real = default_pullback_rank_atol(S),
+        gauge_atol::Real = default_pullback_gauge_atol(ΔDmat, ΔV)
     )
-    mask = abs.(transpose(D) .- D) .< degeneracy_atol
-    Δgauge = norm(view(VᴴΔV, mask))
+
+    n, p = size(V)
+    indD = axes(D, 1)[ind]
+    indV = axes(V, 2)[ind]
+    if !iszerotangent(ΔV)
+        n == size(ΔV, 1) || throw(DimensionMismatch())
+        length(indV) == size(ΔV, 2) || throw(DimensionMismatch())
+        if indV == 1:p
+            ΔV₁ = copy(ΔV)
+        else
+            ΔV₁ = zero(V)
+            for (j, i) in enumerate(indV)
+                ΔV₁[:, i] .= view(ΔV, :, j)
+            end
+        end
+        VᴴΔV₁ = V' * ΔV₁
+        if p == n
+            ΔV₊ = zero!(ΔV₁)
+        else
+            ΔV₊ = mul!(ΔV₁, ViG, VᴴΔV₁, -1, 1)
+        end
+    else
+        ΔV₊ = nothing
+        VᴴΔV₁ = zero!(similar(V, (p, p)))
+    end
+    bc = Base.broadcasted(transpose(D), D, VᴴΔV₁) do d₁, d₂, v
+        return abs(d₁ - d₂) < degeneracy_atol ? v : zero(v)
+    end
+    Δgauge = norm(bc, Inf)
+
     Δgauge ≤ gauge_atol ||
         @warn "`eig` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
-    return
+
+    VᴴΔV₁ .*= conj.(inv_safe.(transpose(D) .- D, degeneracy_atol))
+    VᴴAΔV = VᴴΔV₁
+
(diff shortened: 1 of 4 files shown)


Jutho:
Let me continue on this PR, there are some more inconsistencies in how `ind` is treated also in the `eig` and `eigh` pullbacks. This should perhaps have been part of the previous PR, but it got merged quickly to fix the `svd_full` case. I will also switch over the `eig` and `eigh` pullbacks to this `prepare_and_check_cotangents` format, and at an `ind` argument in the `remove_svd_gauge_dependence` .

Jutho:
Upon some consideration, I think it is actually easier to remove the `ind` argument in the `remove_x_gauge_dependence!` functions. I think in these functions it is fine to assume that the `V` and `ΔV` passed in have the same number of columns. The `ind` is mostly an internal implementation detail that needs to be intercepted for `TruncatedAlgorithm` decompositions.

lkdvos:
That's definitely also okay for me

Jutho:
Ok, this is a first set of changes. It makes the `eigh` pullback related functionality in line with the `svd` one, but also fixes some things in the `svd` pullbacks that went unnoticed. In particular, the gauge check rewritten in terms of a broadcast operation to avoid allocations had the arguments in the ternary operator `? : ` reversed, so that it was plain wrong and a lot of gauge warnings were printed, even after applying `remove_gauge_dependence`. This should now be fixed.

Jutho:
Remains to be done: the `eig` case, which I will tackle next.

Jutho:
Ok, I finally found the time to complete this. I hope the tests pass. Then this is ready for a review @lkdvos .

src/pullbacks/eig.jl
```diff
@@ -1,13 +1,53 @@
...
+        n == size(ΔV, 1) || throw(DimensionMismatch())
+        length(indV) == size(ΔV, 2) || throw(DimensionMismatch())
+        if indV == 1:p
+            ΔV₁ = copy(ΔV)
+        else
+            ΔV₁ = zero(V)
+            for (j, i) in enumerate(indV)
+                ΔV₁[:, i] .= view(ΔV, :, j)
+            end
+        end
```
lkdvos:
```suggestion
        ΔV₁ = similar(V)
        ΔV₁[:, indV] = ΔV
        zero!(view(ΔV₁, :, (length(indV) + 1):p))
```
I think this ends up doing the same, I just had a hard time parsing what was going on before

src/pullbacks/eigh.jl
```diff
@@ -1,13 +1,54 @@
...
+        n == size(ΔV, 1) || throw(DimensionMismatch())
+        length(indV) == size(ΔV, 2) || throw(DimensionMismatch())
+        if indV == 1:p
+            ΔV₁ = copy(ΔV)
+        else
+            ΔV₁ = zero(V)
+            for (j, i) in enumerate(indV)
+                ΔV₁[:, i] .= view(ΔV, :, j)
+            end
+        end
```
lkdvos:
```suggestion
        ΔV₁ = similar(V)
        ΔV₁[:, indV] = ΔV
        zero!(view(ΔV₁, :, (length(indV) + 1):p))
```
same comment here

src/pullbacks/eigh.jl
```diff
@@ -113,47 +129,58 @@ not small compared to `gauge_atol`.
...
-        ΔA = mul!(ΔA, Z, V', 1 // 2, 1)
-        ΔA = mul!(ΔA, V, Z', 1 // 2, 1)
-    elseif !iszerotangent(ΔDmat)
-        ΔDvec = diagview(ΔDmat)
-        p == length(ΔDvec) || throw(DimensionMismatch())
-        ΔA = mul!(ΔA, V * Diagonal(real(ΔDvec)), V', 1, 1)
+        Z .+= Xₖ
+        # we cannot directly multiply Z * V' into ΔA, because we have to
+        # take the Hermitian part, and cannot apply project_hermitian! to
+        # the current contents of ΔA
```
lkdvos, line 172: Not really relevant for this PR, but I think this has come up somewhere before as well. Do you think we could benefit from a `project_hermitian!(C, A, alpha, beta)` approach, or is that really overkill?

lkdvos (commented):
I think I only have a small comment about some aesthetics, otherwise I think this looks good, thanks!

src/pullbacks/eig.jl
Jutho: That is indeed much nicer.

src/pullbacks/eigh.jl
Jutho: Maybe in a TODO, so as not to postpone this PR.

Jutho, line 172:
```suggestion
        # the current contents of ΔA
        # TODO: add an `add_project_hermitian!`
```

lkdvos:
(I can't approve this because I started the PR, so will need approval from someone else)

Jutho (approved):

Jutho:
Tests seem to be failing so don't merge; I'll have to look at it.

lkdvos:
I might have found the fix :)

lkdvos:
Tests failures on AMD seem unrelated, so will just merge as-is.

src/pullbacks/eig.jl
```diff
@@ -1,13 +1,47 @@
...
-    Δgauge = norm(view(VᴴΔV, mask))
+
+    n, p = size(V)
+    indD = axes(D, 1)[ind]
+    indV = axes(V, 2)[ind]
+    if !iszerotangent(ΔV)
+        n == size(ΔV, 1) || throw(DimensionMismatch())
+        length(indV) == size(ΔV, 2) || throw(DimensionMismatch())
+        ΔV₁ = zero(V)
+        ΔV₁[:, indV] = ΔV
```
Jutho, line 14: I would still like to keep some of the old behaviour, so maybe this is a compromise

```suggestion
        if indV == 1:p
            ΔV₁ = copy(ΔV)
        else
            ΔV₁ = zero(V)
            ΔV₁[:, indV] .= ΔV
        end
```

src/pullbacks/eigh.jl
```diff
@@ -1,13 +1,48 @@
...
-    Δgauge = norm(view(aVᴴΔV, mask))
+
+    n, p = size(V)
+    indD = axes(D, 1)[ind]
+    indV = axes(V, 2)[ind]
+    if !iszerotangent(ΔV)
+        n == size(ΔV, 1) || throw(DimensionMismatch())
+        length(indV) == size(ΔV, 2) || throw(DimensionMismatch())
+        ΔV₁ = zero(V)
+        ΔV₁[:, indV] = ΔV
```
Jutho, line 14: Same:
```suggestion
        if indV == 1:p
            ΔV₁ = copy(ΔV)
        else
            ΔV₁ = zero(V)
            ΔV₁[:, indV] .= ΔV
        end
```

Jutho: Oh, I only now noticed this is merged already.

PR #225, opened 2026-05-08
feat: Gaussian sketching + sketched SVD

lkdvos:
This PR implements infrastructure for working with matrix sketches, in particular through Gaussian sampling.
The main additions are `left_sketch!` and `right_sketch!`, which could be thought of as `left_orth` with a sketching truncation strategy (although not implemented that way).

Additionally, this brings the `CUSOLVER_Randomized` algorithm implementation on equal footing by using a `Driver` to select between the `CUSOLVER` implementation and the `Native` one.

### Sketching API

- `left_sketch[!](A; howmany, kwargs...) -> (Q, B)` with `Q` an `m×k` isometry and `B = Q'·A`.
- `right_sketch[!](A; howmany, kwargs...) -> (B, Pᴴ)` with `Pᴴ` a `k×n` co-isometry and `B = A·Pᴴ'`.
- `abstract type SketchingStrategy <: AbstractAlgorithm` — supertype for sketching primitives.
- `GaussianSketching(howmany; numiter=2, rng=default_rng())` — the one concrete strategy currently shipped.

### Sketched SVD

- `SketchedAlgorithm{A,S,T,D}` — `@kwdef` wrapper bundling `alg` (small dense factorization), `sketch` (`SketchingStrategy`), `trunc` (`TruncationStrategy`), and `driver` (`Driver`), to denote decomposing via sketch + decompose + truncate. Mimics `TruncatedAlgorithm`.
- `svd_trunc[!]` and `svd_trunc_no_error[!]` now dispatch on `SketchedAlgorithm`.
  They allocate `(U, S, Vᴴ)` sized for the sketch (`k = sketch.howmany`), check shapes, then call `gesvdr!(driver, A, S, U, Vᴴ; sketch, alg, trunc)`.
- Truncation error: `ε = sqrt((‖A‖+‖S‖)(‖A‖−‖S‖))`, identical to the old `CUSOLVER_Randomized` path.
- The CUDA extension now implements `gesvdr!(::CUSOLVER, A, S, U, Vᴴ; sketch, trunc, alg=nothing)` directly. It currently only accepts `sketch::GaussianSketching` and `trunc::TruncationByOrder` (i.e. `truncrank`) and ignores the inner `alg`. Requires dedicated overloads for the allocation of the output because of CUSOLVER conventions.
- `CUSOLVER_Randomized(; k, p, niters)` reimplemented as `SketchedAlgorithm(; sketch=GaussianSketching(k+p; numiter=niters+1), truncrank(k); driver=CUSOLVER())`. Non-breaking deprecation

### 1. `left_sketch!` vs `left_orth!`

I had a hard time deciding between overloading `left_orth!` directly or introducing a dedicated `left_sketch` function. For simplicity I kept it separate for now, mostly since this is easier to reason through while fleshing out the design, but there might be a reasonable point to be made to merge the two.

### 2. Sketching as an algorithm vs Sketching as a truncation strategy

In the `svd_trunc` implementation, I chose to have **both** `trunc` and `sketch` as a keyword argument, rather than trying to fit this through the same keyword. Again, this is mostly for convenience while I was playing around with this, but we could conceive of a way to merge the two concepts, especially if we decide `left_orth!(A; trunc=gaussiansketch(...))` is a reasonable approach.

### 3. Allocating outputs

One of the things I struggled the most with is the issue of deciding what `initialize_output(svd_trunc!, A, ::SketchedAlgorithm)` should return.
The main issue is that we already are abusing this a bit for the `::TruncatedAlgorithm` codepath, where by convention the `USVh` we pass in is the one that will be used for the `svd_compact!` inside, and is not equivalent to the one that is outputted.
For `::SketchedAlgorithm`, I therefore chose to pass in the sketched sizes, since that is some workspace that can be reused, and that has similar semantics, but it is definitely true that this is a bit unintuitive.

However, the biggest struggle in changing this is that our automatic differentiation implementations hinge on the fact that everything passes through `svd_trunc!(A, out, alg)` in the end, so simply not supporting passing in an `out` argument has quite some implications for the remainder of the code.

This is probably a design choice that we could consider revisiting for a v0.7 version of MatrixAlgebraKit, since it might just be that for these cases where the output size is hard to determine beforehand, it doesn't necessarily make too much sense to allow passing it in as inputs.
The alternative would be to start playing around with "reallocation" strategies, where we shrink/expand provided inputs if they don't have the correct sizes. This does however feel like it might just be a bit too convoluted for what it buys us.

### 4. Truncation error formula

```julia
ε = sqrt((‖A‖ + ‖S‖) * (‖A‖ − ‖S‖))
```

This is the `‖A‖_F^2 − ‖S‖_F^2` identity, written for numerical stability.
However, I had to significantly lower the tolerances for any such tests, since the numerical accuracy of the sketching and the floating point errors are quite finicky for this.
This is of course the same discussion we had before, where it is not clear to me that `svd_trunc` really should have the responsibility to provide truncation errors, and `svd_trunc_no_error` is a way more sensible default mode.
Presumably this is fine as long as it is clearly documented?

### 5. Gauge fixing depends on the exact truncation path

In the current implementation, the gauge-fixing happens at the moment of decomposing the inner small core tensor, which will then still get unprojected onto a larger space through the sketch.
While it is somewhat straightforward to change this, it might be confusing that there are gaugefix keywords at multiple places, which don't have the same effect.
I'm not sure what the best way forwards is for this, but this might require some more discussion.

## TODO

- [ ] Documentation page
- [ ] Resolve discussions above
- [ ] Testing in the context of AD
- [ ] Testing in the context of GPU

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -4,11 +4,12 @@ using MatrixAlgebraKit
 using MatrixAlgebraKit: @algdef, Algorithm, check_input
 using MatrixAlgebraKit: one!, zero!, uppertriangular!, lowertriangular!
 using MatrixAlgebraKit: diagview, sign_safe
-using MatrixAlgebraKit: CUSOLVER, LQViaTransposedQR, TruncationByValue, AbstractAlgorithm
+using MatrixAlgebraKit: CUSOLVER, LQViaTransposedQR, TruncationByValue, TruncationByOrder, AbstractAlgorithm
+using MatrixAlgebraKit: GaussianSketching, SketchingStrategy, SketchedAlgorithm
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eig_algorithm, default_eigh_algorithm
 import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdp!, gesvdr!, gesvdj!
 import MatrixAlgebraKit: heevj!, heevd!, geev!
-import MatrixAlgebraKit: _gpu_Xgesvdr!, _sylvester, svd_rank
+import MatrixAlgebraKit: _sylvester, svd_rank
 using CUDA, CUDA.cuBLAS
 using CUDA: i32
 using LinearAlgebra
@@ -17,6 +18,7 @@ using LinearAlgebra: BlasFloat
 include("yacusolver.jl")
 
 MatrixAlgebraKit.default_driver(::Type{TA}) where {TA <: StridedCuVecOrMat{<:BlasFloat}} = CUSOLVER()
+MatrixAlgebraKit.default_driver(::Type{<:SketchedAlgorithm}, ::Type{TA}) where {TA <: StridedCuVecOrMat{<:BlasFloat}} = CUSOLVER()
 
 function MatrixAlgebraKit.default_svd_algorithm(::Type{T}; kwargs...) where {T <: StridedCuVecOrMat{<:BlasFloat}}
     return QRIteration(; kwargs...)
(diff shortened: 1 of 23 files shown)


ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -50,8 +52,31 @@ end
...
+# Sketched SVD via cuSOLVER's gesvdr kernel
+function gesvdr!(
+        ::CUSOLVER, A::StridedCuMatrix, S, U::StridedCuMatrix, Vᴴ::StridedCuMatrix;
+        sketch::GaussianSketching, trunc::TruncationByOrder, alg::AbstractAlgorithm = DefaultAlgorithm()
+    )
+    isempty(A) && return U, S, Vᴴ
+    m, n = size(A); minmn = min(m, n)
+    k = trunc.howmany
+    1 ≤ k ≤ minmn ||
+        throw(ArgumentError("trunc.howmany=$k must satisfy 1 ≤ k ≤ min(size(A))=$minmn"))
```
Jutho: Do we generically throw an error if truncrank is bigger than maximal size? Would it not be more useful if we consider `truncrank` as an upper bound?

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -50,8 +52,31 @@ end
...
+        )
+    )
+    p = min(p, minmn - k - 1)
+    niters = sketch.numiter - 1
+
+    Uk = view(U, :, 1:k)
+    Vᴴk = view(Vᴴ, 1:k, :)
+    Sk = view(diagview(S), 1:k)
+    YACUSOLVER.gesvdr!(A, Sk, Uk, Vᴴk; k, p, niters)
+    return Uk, S, Vᴴk
```
Jutho: Is this the correct return? This does not seem type stable. Above you have `isempty(A) && return U, S, Vᴴ`.

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
lkdvos: You are definitely right, I think I was still dealing with the fact that CUSOLVER doesn't accept that by default.
There is an additional question about if we should catch that and then not do any sketching at all, since in that case we might as well just do the compact decomposition, but I'm fine with leaving that for a future optimization.

This does raise the question if we should error for the case where the `trunc.howmany` is larger than the `sketch.howmany`.
The cuSOLVER interface does not have this problem, since they use `k = trunc.howmany` and `p = sketch.howmany - trunc.howmany` and require both to be positive, and it does seem like it is sensible to not allow people to specify an algorithm where the truncation rank is higher than the sketching rank?

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
lkdvos: Absolutely not, I apparently still had some things to work on for the CUDA case 🙃

Jutho (approved):

kshyatt:
Is there more to be done here? It looks pretty good to me but I'm less familiar with the sketching stuff

lkdvos:
Well, I think the primary questions were about the interface, and how that would interact with TensorKit and higher-level packages. I've been doing some research and think I'll be able to come up with some actual dynamic schemes to have tolerances etc, but this seems like a good start?

kshyatt:
I'm definitely not opposed to just giving it a try and modifying things as needed as we go :)

kshyatt:
Reping on this. Are we good to just rebase and merge if CI passes?

PR #226, opened 2026-05-10
fix: LQ pullback modifies input cotangents

lkdvos:
Bumped into this to resolve https://github.com/QuantumKitHub/TensorKit.jl/pull/419.

TLDR: the QR pullback correctly did not modify the input cotangents, using a `getindex`, while the LQ pullback uses a `view`.

### DIFF
--- a/src/pullbacks/lq.jl
+++ b/src/pullbacks/lq.jl
@@ -13,7 +13,7 @@ function check_and_prepare_lq_cotangents(
         size(ΔQ) == size(Q) || throw(DimensionMismatch("ΔQ must have the same size as Q"))
         ΔQ₁ .= view(ΔQ, 1:p, 1:n)
         if p == minmn # full rank case, ΔQ₃ contains gauge-invariant information along Q₁
-            ΔQ₃ = ΔQ[(minmn + 1):size(Q, 1), :] # extra columns in the case of qr_full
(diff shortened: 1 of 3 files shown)


kshyatt:
Could we add an explicit test to make sure the cotangents are not touched?

Jutho:
Thanks this looks good.

This makes me however wonder to what extent we can rely on `getindex` returning a copy. If we ever switch over the TensorKit `block`s to return a `StridedView` instead of a `reshape(view(...))`, we might run into problem, because `getindex` producing a view on `StridedView` objects has been a (possibly misinformed) design principle from early on. Maybe we should be more explicit with `copy(view(...))` ?

Jutho (approved):

lkdvos:
@Jutho I have switched everything to `copy(view(...))`, but I think this is more a case of Strided performing piracy on a Base method by altering its semantics than MatrixAlgebraKit not being able to rely on this. It is a good comment though, this would be an annoying thing to debug in the future...

@kshyatt Good point, should be done now :).
Interestingly, my first idea was that this should be caught by the AD testing frameworks, but that just isn't the case because of different requirements: Mooncake actually requires us to zero out these output tangents, so that would never catch this (or be affected by this), and adding this to ChainRulesTestUtils seems like it will be hard to find someone to merge this...
This also brings up the point that if we end up dropping chainrules in the end, it would actually be more efficient to assume we can modify these inputs, but I guess we can keep that for the future 😉

Jutho (approved):

PR #232, opened 2026-05-18
Make `svd_pullback!` GPU-compatible

### LINKED ISSUE #228: SVD pullback rules aren't GPU-friendly
As noted in https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/commit/54f602172716372f84424b97fe585222e887e0ca#r185141141 and also notable in https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/227, the SVD pullback triggers scalar indexing due to the computation of `ΔS₁` [here](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/main/src/pullbacks/svd.jl#L93).

kshyatt:
Should fix https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/228

We need to `collect` the `ind`s for the truncated tests because otherwise we trigger scalar indexing :(

The solution for the pullback itself is a bit rough so if someone has a nicer list comprehension, feel free to suggest!

### DIFF
--- a/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
+++ b/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
@@ -191,7 +191,7 @@ function svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback
 end
 
 function svd_pullback!(ΔA::AnyROCMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyROCVector; kwargs...)
-    svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
+    return svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
 end
 
 end
--- a/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
+++ b/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
@@ -186,8 +186,7 @@ function _sylvester(A::AnyROCMatrix, B::AnyROCMatrix, C::AnyROCMatrix)
 end
 
 function svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
-    r = findlast(s -> s ≥ rank_atol, S)
-    return isnothing(r) ? length(S) : r
+    return something(findlast(≥(rank_atol), S), 0)
 end
 
 function svd_pullback!(ΔA::AnyROCMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyROCVector; kwargs...)
--- a/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
+++ b/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
@@ -8,7 +8,7 @@ using MatrixAlgebraKit: ROCSOLVER, LQViaTransposedQR, TruncationStrategy, NoTrun
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eigh_algorithm
 import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdj!
 import MatrixAlgebraKit: heevj!, heevd!, heev!, heevx!
-import MatrixAlgebraKit: _sylvester, svd_rank
+import MatrixAlgebraKit: _sylvester, svd_rank, svd_pullback!
 using AMDGPU
 using LinearAlgebra
 using LinearAlgebra: BlasFloat
@@ -185,6 +185,13 @@ function _sylvester(A::AnyROCMatrix, B::AnyROCMatrix, C::AnyROCMatrix)
     return ROCArray(hX)
 end
 
-svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S)) = findlast(s -> s ≥ rank_atol, S)
+function svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
+    r = findlast(s -> s ≥ rank_atol, S)
+    return isnothing(r) ? length(S) : r
+end
+
+function svd_pullback!(ΔA::AnyROCMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyROCVector; kwargs...)
+    svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
+end
 
 end
--- a/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
+++ b/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
@@ -185,6 +185,6 @@ function _sylvester(A::AnyROCMatrix, B::AnyROCMatrix, C::AnyROCMatrix)
     return ROCArray(hX)
 end
 
-svd_rank(S::AnyROCVector, rank_atol) = findlast(s -> s ≥ rank_atol, S)
+svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S)) = findlast(s -> s ≥ rank_atol, S)
 
 end
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -8,7 +8,7 @@ using MatrixAlgebraKit: CUSOLVER, LQViaTransposedQR, TruncationByValue, Abstract
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eig_algorithm, default_eigh_algorithm
 import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdp!, gesvdr!, gesvdj!
 import MatrixAlgebraKit: heevj!, heevd!, geev!
-import MatrixAlgebraKit: _gpu_Xgesvdr!, _sylvester, svd_rank
+import MatrixAlgebraKit: _gpu_Xgesvdr!, _sylvester, svd_rank, svd_pullback!
 using CUDA, CUDA.cuBLAS
 using CUDA: i32
 using LinearAlgebra
@@ -197,6 +197,13 @@ function _sylvester(A::AnyCuMatrix, B::AnyCuMatrix, C::AnyCuMatrix)
     return CuArray(hX)
 end
 
-svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S)) = findlast(s -> s ≥ rank_atol, S)
+function svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
+    r = findlast(s -> s ≥ rank_atol, S)
+    return isnothing(r) ? length(S) : r
+end
+
+function svd_pullback!(ΔA::AnyCuMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyCuVector; kwargs...)
+    svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
+end
 
 end
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -203,7 +203,7 @@ function svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_
 end
 
 function svd_pullback!(ΔA::AnyCuMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyCuVector; kwargs...)
-    svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
+    return svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
 end
 
 end
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -198,8 +198,7 @@ function _sylvester(A::AnyCuMatrix, B::AnyCuMatrix, C::AnyCuMatrix)
 end
 
 function svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
-    r = findlast(s -> s ≥ rank_atol, S)
-    return isnothing(r) ? length(S) : r
+    return something(findlast(≥(rank_atol), S), 0)
 end
 
 function svd_pullback!(ΔA::AnyCuMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyCuVector; kwargs...)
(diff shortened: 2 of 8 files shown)


test/testsuite/mooncake/svd.jl
```diff
@@ -141,7 +141,7 @@ function test_mooncake_svd_trunc(
 
         @testset "trunctol" begin
             S = svd_vals(A)
-            trunc = trunctol(atol = S[1] / 2)
+            trunc = trunctol(atol = collect(S)[1] / 2)
```
Jutho: Is `first(S)` allowed on the GPU?

kshyatt:
```julia
julia> using CUDA
[ Info: Precompiling CUDA [052768ef-5323-5732-b1bb-66c8b64840ba](cache misses: wrong source (2), mismatched flags (7))
Precompiling CUDA finished.
  15 dependencies successfully precompiled in 165 seconds. 87 already precompiled.

julia> S = CUDA.cuRAND.rand(32);

julia> first(S)
ERROR: Scalar indexing is disallowed.
Invocation of getindex resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.

If you want to allow scalar iteration, use `allowscalar` or `@allowscalar`
to enable scalar iteration globally or for the operations in question.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:44
 [2] errorscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:151
 [3] _assertscalar(op::String, behavior::GPUArraysCore.ScalarIndexing)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:124
 [4] assertscalar(op::String)
   @ GPUArraysCore ~/.julia/packages/GPUArraysCore/aNaXo/src/GPUArraysCore.jl:112
 [5] getindex
   @ ~/.julia/packages/GPUArrays/Lr6Tr/src/host/indexing.jl:50 [inlined]
 [6] first(a::CuArray{Float32, 1, CUDACore.DeviceMemory})
   @ Base ./abstractarray.jl:452
 [7] top-level scope
   @ REPL[3]:1
```

kshyatt: in short, no :(

lkdvos:
```suggestion
            trunc = trunctol(atol = maximum(S) / 2)
```
This should probably work?

src/pullbacks/svd.jl
```diff
@@ -84,19 +84,16 @@ function check_and_prepare_svd_cotangents(
...
-            if i <= r
-                ΔS₁[i] = real(ΔS[j])
-            else
-                Δgauge = max(Δgauge, abs(ΔS[j]))
-            end
-        end
+        good_indS = findall(i -> i <= r, indS)
+        bad_indS = setdiff(1:length(indS), good_indS)
+        ΔS₁[indS[good_indS]] = real.(ΔS[good_indS])
+        Δgauge = max(Δgauge, mapreduce(abs, max, ΔS[bad_indS]))
```
lkdvos: Would it help if we re-use the machinery we have for the `TruncationIntersection`, which I think we already made work, and then something like this? (untested)

```
bad_indS = _ind_intersect(r+1:length(ΔS), indS)
ΔS₁ = real(ΔS)
badΔS₁ = view(ΔS₁, bad_indS)
Δgauge = max(Δgauge, maximum(abs, badΔS₁))
badΔS₁ .= 0
```

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -185,6 +185,6 @@ function _sylvester(A::AnyROCMatrix, B::AnyROCMatrix, C::AnyROCMatrix)
     return ROCArray(hX)
 end
 
-svd_rank(S::AnyROCVector, rank_atol) = findlast(s -> s ≥ rank_atol, S)
+svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S)) = findlast(s -> s ≥ rank_atol, S)
```
lkdvos: Here we probably also should have a fallback for when `all(>=(rank_atol), S)`, since this will now probably return `nothing`.
(Also fair if this is outside of the scope of this PR though!

lkdvos (commented):
It seems a bit much to force a copy of `ind` across the board, even in non-GPU context. Is there any way we can only selectively do this, or find a different approach?
I'm also slightly confused by the `collect`, I thought we did our best to keep everything on GPU precisely to allow the GPU kernels to do their work, and having `ind::Vector{Int}` combined with `values::CuVector` was bad?

kshyatt:
> I'm also slightly confused by the collect, I thought we did our best to keep everything on GPU precisely to allow the GPU kernels to do their work, and having ind::Vector{Int} combined with values::CuVector was bad?

I think in general we want to keep the `inds` on the GPU, but in this case, where we're indexing a CPU array (`axes(A, dim)`), it's a problem. I should be able to make a pass-through in the GPU extensions that doesn't force the CPU ones to copy.

test/testsuite/mooncake/svd.jl
kshyatt: Why is this suggestion showing up twice lol, I swear to god Microsoft is trying to Havana syndrome me

src/pullbacks/svd.jl
kshyatt: I'll test this now

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
kshyatt: Added this, it could impact things downstream for sure...

src/pullbacks/svd.jl
kshyatt: ΔS₁ ends up with the wrong length with this, I'll try to fix...

kshyatt: Not sure this is much nicer but it's passing again at least 😓

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -185,6 +185,13 @@ function _sylvester(A::AnyROCMatrix, B::AnyROCMatrix, C::AnyROCMatrix)
     return ROCArray(hX)
 end
 
-svd_rank(S::AnyROCVector, rank_atol) = findlast(s -> s ≥ rank_atol, S)
+function svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
+    r = findlast(s -> s ≥ rank_atol, S)
+    return isnothing(r) ? length(S) : r
```
lkdvos:
```suggestion
    return something(findlast(≥(rank_atol), S), 0)
```

Slight readability change, but also: shouldn't this be `0`? If none of the entries are larger than the tolerance, the rank is 0 I think?

src/pullbacks/svd.jl
```diff
@@ -84,19 +84,18 @@ function check_and_prepare_svd_cotangents(
...
-            if i <= r
-                ΔS₁[i] = real(ΔS[j])
-            else
-                Δgauge = max(Δgauge, abs(ΔS[j]))
-            end
-        end
+        view(ΔS₁, 1:length(good_indS)) .= real.(view(ΔS, good_indS))
+        length(ΔS₁) == length(S₁) || throw(DimensionMismatch(lazy"length of ΔS₁ ($(length(ΔS₁))) does not match length of S₁ ($(length(S₁)))"))
+        badΔS₁ = view(ΔS, bad_indS)
+        Δgauge = max(Δgauge, maximum(abs, badΔS₁; init = 0.0))
```
```suggestion
        Δgauge = max(Δgauge, maximum(abs, badΔS₁; init = abs(zero(eltype(ΔS))))
```

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -197,6 +197,13 @@ function _sylvester(A::AnyCuMatrix, B::AnyCuMatrix, C::AnyCuMatrix)
     return CuArray(hX)
 end
 
-svd_rank(S::AnyCuVector, rank_atol) = findlast(s -> s ≥ rank_atol, S)
+function svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
+    r = findlast(s -> s ≥ rank_atol, S)
+    return isnothing(r) ? length(S) : r
+end
```
lkdvos, line 202:
```suggestion
function svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
    return something(findlast(≥(rank_atol), S), 0)
end
```
Same comment here

kshyatt:
Moved away from `view` as the broadcast from `SubArray` to `SubArray` was causing compilation errors, the current change allows a sample PEPSKit script of @GlebFedorovich to run successfully 🚀

lkdvos (approved):

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -185,6 +185,12 @@ function _sylvester(A::AnyROCMatrix, B::AnyROCMatrix, C::AnyROCMatrix)
     return ROCArray(hX)
 end
 
-svd_rank(S::AnyROCVector, rank_atol) = findlast(s -> s ≥ rank_atol, S)
+function svd_rank(S::AnyROCVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
+    return something(findlast(≥(rank_atol), S), 0)
```
Jutho, line 189: Can we make this type stable?


```suggestion
    return something(findlast(≥(rank_atol), S), zero(eltype(S)))
```

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -197,6 +197,12 @@ function _sylvester(A::AnyCuMatrix, B::AnyCuMatrix, C::AnyCuMatrix)
     return CuArray(hX)
 end
 
-svd_rank(S::AnyCuVector, rank_atol) = findlast(s -> s ≥ rank_atol, S)
+function svd_rank(S::AnyCuVector; rank_atol = MatrixAlgebraKit.default_pullback_rank_atol(S))
+    return something(findlast(≥(rank_atol), S), 0)
```
Jutho, line 201:
```suggestion
    return something(findlast(≥(rank_atol), S), zero(eltype(S)))
```

src/pullbacks/svd.jl
```diff
@@ -84,19 +84,18 @@ function check_and_prepare_svd_cotangents(
...
+        good_indS = _ind_intersect(1:r, indS)
         ΔS₁ = zero(S₁)
-        for (j, i) in enumerate(indS)
-            if i <= r
-                ΔS₁[i] = real(ΔS[j])
-            else
-                Δgauge = max(Δgauge, abs(ΔS[j]))
-            end
-        end
+        ΔS₁[1:length(good_indS)] .= real.(ΔS[good_indS])
```
Jutho, line 96: Does this work on the GPU?
```suggestion
        ΔS₁[1:length(good_indS)] .= real.(view(ΔS, good_indS))
```

src/pullbacks/svd.jl
```diff
@@ -84,19 +84,18 @@ function check_and_prepare_svd_cotangents(
...
+        length(ΔS₁) == length(S₁) || throw(DimensionMismatch(lazy"length of ΔS₁ ($(length(ΔS₁))) does not match length of S₁ ($(length(S₁)))"))
```
Jutho: We have `ΔS₁ = zero(S₁)` two lines above, so how could this not be satisfied?


```suggestion
```

test/testsuite/enzyme/svd.jl
```diff
@@ -69,7 +69,7 @@ function test_enzyme_svd_trunc(
         end
         @testset "trunctol" begin
             S = svd_vals(A, alg)
-            trunc = trunctol(atol = S[1] / 2)
+            trunc = trunctol(atol = maximum(S) / 2)
```
Jutho, line 72: Hooray for `maximum`.

Jutho:
I did have some final suggestions so I disabled auto-merge.

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
lkdvos: Wait, this returns an `Int` to give you the *index*, and not the actual value, so this is already type-stable

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
lkdvos: Same comment here, closing this :)

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
Jutho: Ok, I was not paying attention.

src/pullbacks/svd.jl
kshyatt: I tried this and was getting compiler errors from Gleb's code. Let me give it one more go...

kshyatt:
```julia
ERROR: LoadError: GPU compilation of MethodInstance for (::GPUArrays.var"#gpu_broadcast_kernel_linear#_copyto!##1")(::KernelAbstractions.CompilerMetadata{KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicCheck, Nothing, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, KernelAbstractions.NDIteration.NDRange{1, KernelAbstractions.NDIteration.DynamicSize, KernelAbstractions.NDIteration.DynamicSize, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}, CartesianIndices{1, Tuple{Base.OneTo{Int64}}}}}, ::CuDeviceVector{Float64, 1}, ::Base.Broadcast.Broadcasted{CUDACore.CuArrayStyle{1, CUDACore.DeviceMemory}, Tuple{Base.OneTo{Int64}}, typeof(real), Tuple{Base.Broadcast.Extruded{SubArray{ComplexF64, 1, CuDeviceVector{ComplexF64, 1}, Tuple{Vector{Int64}}, false}, Tuple{Bool}, Tuple{Int64}}}}) failed        
KernelError: passing non-bitstype argument                                                                                                                                                                      
                                                                                                                                                                                                                
Argument 4 to your kernel function is of type Base.Broadcast.Broadcasted{CUDACore.CuArrayStyle{1, CUDACore.DeviceMemory}, Tuple{Base.OneTo{Int64}}, typeof(real), Tuple{Base.Broadcast.Extruded{SubArray{ComplexF64, 1, CUDACore.CuDeviceVector{ComplexF64, 1}, Tuple{Vector{Int64}}, false}, Tuple{Bool}, Tuple{Int64}}}}, which is not a bitstype:                                                                            
  .args is of type Tuple{Base.Broadcast.Extruded{SubArray{ComplexF64, 1, CUDACore.CuDeviceVector{ComplexF64, 1}, Tuple{Vector{Int64}}, false}, Tuple{Bool}, Tuple{Int64}}} which is not isbits.                 
    .1 is of type Base.Broadcast.Extruded{SubArray{ComplexF64, 1, CUDACore.CuDeviceVector{ComplexF64, 1}, Tuple{Vector{Int64}}, false}, Tuple{Bool}, Tuple{Int64}} which is not isbits.                         
      .x is of type SubArray{ComplexF64, 1, CUDACore.CuDeviceVector{ComplexF64, 1}, Tuple{Vector{Int64}}, false} which is not isbits.                                                                           
        .indices is of type Tuple{Vector{Int64}} which is not isbits.                                                                                                                                           
          .1 is of type Vector{Int64} which is not isbits.                                                                                                                                                      
            .ref is of type MemoryRef{Int64} which is not isbits.                                                                                                                                               
              .mem is of type Memory{Int64} which is not isbits.                                                                                                                                                
                                                                                                                                                                                                                
                                                                                                                                                                                                                
Only bitstypes, which are "plain data" types that are immutable                                                                                                                                                 
and contain no references to other values, can be used in GPU kernels.                                                                                                                                          
For more information, see the `Base.isbitstype` function.      
```

From ` MatrixAlgebraKit ~/.julia/dev/MatrixAlgebraKit/src/pullbacks/svd.jl:95  `

Jutho:
Ok, failures on windows-latest seem unrelated, I'll merge.

PR #233, opened 2026-05-18
fix: special case pullbacks for fully truncated decompositions

lkdvos:
Brought up by @leburgel, the `norm(Base.broadcasted(...), Inf)` ends up calling `zero(eltype(bc))` on the broadcasted expression, and when `eltype(bc) = Any` this ends up throwing.
The easiest way around this is to simply catch the "empty" case in the pullbacks themselves, which I think I've now implemented correctly but it might be good to have another look at this.

### DIFF
--- a/src/pullbacks/eig.jl
+++ b/src/pullbacks/eig.jl
@@ -27,14 +27,13 @@ function check_and_prepare_eig_cotangents(
         VᴴΔV₁ = zero!(similar(V, (p, p)))
     end
 
-    if !isempty(D) # norm(bc, Inf) calls eltype on empty inputs
-        bc = Base.broadcasted(transpose(D), D, VᴴΔV₁) do d₁, d₂, v
-            return abs(d₁ - d₂) < degeneracy_atol ? v : zero(v)
-        end
-        Δgauge = norm(bc, Inf)
-        Δgauge ≤ gauge_atol ||
-            @warn "`eig` cotangents sensitive to gauge choice: (|Δgauge| = $Δgauge)"
+    bc = Base.broadcasted(transpose(D), D, VᴴΔV₁) do d₁, d₂, v
+        return abs(d₁ - d₂) < degeneracy_atol ? v : zero(v)
     end
+    Δgauge = norm(bc, Inf)
+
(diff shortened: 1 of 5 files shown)


src/pullbacks/eig.jl
```diff
@@ -81,6 +82,7 @@ function eig_pullback!(
     D = diagview(Dmat)
     n == length(D) || throw(DimensionMismatch())
     (n, n) == size(ΔA) || throw(DimensionMismatch())
+    isempty(D) && return ΔA
```
Jutho: This is a very subjective change, but since we already have the length of `D` available:

```suggestion
    iszero(n) && return ΔA
```

src/pullbacks/eig.jl
```diff
@@ -144,6 +146,7 @@ function eig_trunc_pullback!(
     (n, n) == size(ΔA) || throw(DimensionMismatch())
     D = diagview(Dmat)
     p == length(D) || throw(DimensionMismatch())
+    isempty(D) && return ΔA
```
Jutho:
```suggestion
    iszero(p) && return ΔA
```

src/pullbacks/eigh.jl
```diff
@@ -82,6 +83,7 @@ function eigh_pullback!(
     D = diagview(Dmat)
     n == length(D) || throw(DimensionMismatch())
     (n, n) == size(ΔA) || throw(DimensionMismatch())
+    isempty(D) && return ΔA
```
Jutho:
```suggestion
    iszero(n) && return ΔA
```

src/pullbacks/eigh.jl
```diff
@@ -137,6 +139,7 @@ function eigh_trunc_pullback!(
     D = diagview(Dmat)
     p == length(D) || throw(DimensionMismatch())
     (n, n) == size(ΔA) || throw(DimensionMismatch())
+    isempty(D) && return ΔA
```
Jutho:
```suggestion
    iszero(p) && return ΔA
```

Jutho (approved):

Issue #229, opened 2026-05-14
EIGH pullback rules aren't GPU-friendly


kshyatt:
cf https://buildkite.com/julialang/matrixalgebrakit-dot-jl/builds/2714#019e2099-3d27-4688-9409-6ff615e2f5d6

Culprit for scalar indexing is https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/main/src/pullbacks/eigh.jl#L33

One simple solution for now would be to just compute `maximum(abs, bc)` while I add support for the `norm(bc, Inf)` in GPUArrays

PR #235, opened 2026-05-20
Working pullback and tests for EIGH + CUDA

kshyatt:
https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/229 should be fixed. Again happy to take suggestions. I'll try to run down the `SubArray` sadness in `GPUArrays` as this is hardly ideal

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -8,7 +8,7 @@ using MatrixAlgebraKit: CUSOLVER, LQViaTransposedQR, TruncationByValue, Abstract
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eig_algorithm, default_eigh_algorithm
 import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdp!, gesvdr!, gesvdj!
 import MatrixAlgebraKit: heevj!, heevd!, geev!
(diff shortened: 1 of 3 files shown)


lkdvos (approved):
Looks good to me!

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -205,4 +205,8 @@ function svd_pullback!(ΔA::AnyCuMatrix, A, USVᴴ, ΔUSVᴴ, ind::AnyCuVector;
     return svd_pullback!(ΔA, A, USVᴴ, ΔUSVᴴ, collect(ind); kwargs...)
 end
 
+function eigh_pullback!(ΔA::AnyCuMatrix, A, DV, ΔDV, ind::AnyCuVector; kwargs...)
+    return eigh_pullback!(ΔA, A, DV, ΔDV, collect(ind); kwargs...)
```
Jutho, line 209: Just as a question; how are `ind` of type `AnyCuVector` (I assume that is then `CuVector{Int}`) generated? Wouldn't most index finding operations automatically yield ranges or `Vector{Int}` instances?

Jutho (approved):

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
kshyatt: In this case it's from `_ind_intersect` and `findtruncated` for these cases. My recollection (which might well be wrong) is that we wanted these inds to live on the GPU due to how the pullbacks were previously implemented.

PR #236, opened 2026-05-20
Working EIG + CUDA AD

kshyatt:
Added the `Hermitian` wrapper to avoid `cholesky` yelling that `V' * V` isn't Hermitian.

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -8,7 +8,11 @@ using MatrixAlgebraKit: CUSOLVER, LQViaTransposedQR, TruncationByValue, Abstract
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eig_algorithm, default_eigh_algorithm
 import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdp!, gesvdr!, gesvdj!
 import MatrixAlgebraKit: heevj!, heevd!, geev!
+<<<<<<< HEAD
 import MatrixAlgebraKit: _gpu_Xgesvdr!, _sylvester, svd_rank, svd_pullback!, eigh_pullback!
+=======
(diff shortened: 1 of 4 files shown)


kshyatt:
Sigh... this *was* working locally for me... back to the CUDA mines we go!

kshyatt:
What the helly, it's passing locally for me still...

kshyatt:
OK the 1.10 failure is RNG related

lkdvos:
Was it just accidentally using a pathological case, or is this actually not very stable to begin with?

kshyatt:
I think the former, considering it worked on 1.12...

lkdvos (approved):
If tests pass, happy to merge this!

kshyatt:
> Was it just accidentally using a pathological case, or is this actually not very stable to begin with?

Or maybe not. Something really weird on 1.10 here

Jutho:
There does seem to be something going on, potentially with degenerate eigenvalues. There is a "gauge sensitive" warning. 

```julia
Output generated during execution of 'mooncake/eig':
┌ 7-element Vector{Tuple{Float64, Float64}}:
│  (-0.701143421209963, -1.2117169941375323)
│  (-13314.483677792308, -1.2117169941375323)
│  (-0.6760447251535879, -1.2117169941375323)
│  (-1.5543476938692038, -1.2117169941375323)
│  (3.792817048932193, -1.2117169941375323)
│  (-92.0990632729372, -1.2117169941375323)
│  (0.09235575050115585, -1.2117169941375323)
│ 
│ ┌ Warning: `eig` cotangents sensitive to gauge choice: (|Δgauge| = 1.3540607507349773)
└ └ @ MatrixAlgebraKit ~/work/MatrixAlgebraKit.jl/MatrixAlgebraKit.jl/src/pullbacks/eig.jl:36
```

I would also have to look into it more closely to see what is going on. Not sure what this vector of tuples exactly means.

kshyatt:
@jutho it's annoyingly hard to interpret, basically it's a comparison of the summed (y, x) differences from AD (right column) and finite differences (left column) for various epsilon in FD.

Jutho:
Ok so that is why the right column is constant, there is no epsilon in AD. But the left column then shows no convergence or consistency for different epsilon, so I assume there is some degeneracy messing things up. I have been wondering if it would be useful to have the "gauge sensitivity" warning be more informative, to explain what exactly is contributing to this.

kshyatt:
@Jutho seems to be working now with a different seed?

PR #238, opened 2026-05-26
Improve svd gaugefixing performance

kshyatt:
The existing function is extremely inefficient on GPU as it performs a **separate** `reduce` for every column of `U`. Here, these are combined into one `mapreducedim` call under the hood. This also allows us to avoid a lot of separate kernel launches by updating `U` and `Vh` in one broadcast call each, rather than separate calls for each column/row.

Some testing with `@btime` for `size(U) = (64,32)` and `size(Vh) = (32,128)` showed:

```
Old:  3.093 ms (3976 allocations: 132.08 KiB)
New:  66.295 μs (120 allocations: 4.38 KiB)
```

So a pretty substantial improvement. And 

```julia
U1 = copy(U)                                                                                                                                                                                                    
U2 = copy(U)                                                                                                                                                                                                    
Vᴴ1 = copy(Vᴴ)                                                                                                                                                                                                  
Vᴴ2 = copy(Vᴴ)                                                                                                                                                                                                  
                                                                                                                                                                                                                
@assert collect(gaugefix_old!(U1, Vᴴ1)) ≈ collect(gaugefix_new!(U2, Vᴴ2))    
```
passed! (Although let me check this again with `randn`...)

### DIFF
--- a/src/common/gauge.jl
+++ b/src/common/gauge.jl
@@ -67,9 +67,9 @@ end
 
 function gaugefix!(::Union{typeof(svd_compact!), typeof(svd_trunc!)}, U, Vᴴ)
     @assert axes(U, 2) == axes(Vᴴ, 1)
-    signs = sign.(reduce(_largest, U; dims = 1, init = zero(eltype(U))))
-    signs_t = transpose(signs)
-    @. U = U * conj(signs)
-    @. Vᴴ = signs_t * Vᴴ
+    max_us = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
+    max_us_t = transpose(max_us)
+    @. U = U * (conj ∘ sign)(max_us)
+    @. Vᴴ = sign(max_us_t) * Vᴴ
     return (U, Vᴴ)
 end
--- a/src/common/gauge.jl
+++ b/src/common/gauge.jl
@@ -67,9 +67,10 @@ end
 
 function gaugefix!(::Union{typeof(svd_compact!), typeof(svd_trunc!)}, U, Vᴴ)
     @assert axes(U, 2) == axes(Vᴴ, 1)
-    max_us = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
-    max_us_t = transpose(max_us)
-    @. U = U * (conj ∘ sign)(max_us)
-    @. Vᴴ = sign(max_us_t) * Vᴴ
+    signs = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
+    @. signs = sign(signs)
+    signs_t = transpose(signs)
+    @. U = U * conj(signs)
+    @. Vᴴ = signs_t * Vᴴ
     return (U, Vᴴ)
 end
--- a/src/common/gauge.jl
+++ b/src/common/gauge.jl
@@ -10,7 +10,8 @@ is real and positive.
 
 # Helper functions
 _argmaxabs(x) = reduce(_largest, x; init = zero(eltype(x)))
-_largest(x, y) = abs(x) < abs(y) ? y : x
+_largest(x::Real, y::Real) = abs(x) < abs(y) ? y : x
+_largest(x::Complex, y::Complex) = abs2(x) < abs2(y) ? y : x
 
 function gaugefix!(::typeof(qr_householder!), Q, R, Rd)
     ax = Base.OneTo(length(Rd))
--- a/src/common/gauge.jl
+++ b/src/common/gauge.jl
@@ -10,7 +10,7 @@ is real and positive.
 
 # Helper functions
 _argmaxabs(x) = reduce(_largest, x; init = zero(eltype(x)))
-_largest(x, y) = abs(x) < abs(y) ? y : x
+_largest(x, y) = real(abs(x)) < real(abs(y)) ? y : x
 
 function gaugefix!(::typeof(qr_householder!), Q, R, Rd)
     ax = Base.OneTo(length(Rd))
@@ -67,12 +67,8 @@ end
 
 function gaugefix!(::Union{typeof(svd_compact!), typeof(svd_trunc!)}, U, Vᴴ)
     @assert axes(U, 2) == axes(Vᴴ, 1)
-    for j in axes(U, 2)
-        u = view(U, :, j)
-        v = view(Vᴴ, j, :)
-        s = sign(_argmaxabs(u))
-        u .*= conj(s)
-        v .*= s
-    end
+    signs = sign.(reduce(_largest, U; dims = 1, init = zero(eltype(U))))
+    @. U = U * signs
+    @. transpose(Vᴴ) = signs * transpose(Vᴴ)
     return (U, Vᴴ)
 end
--- a/src/common/gauge.jl
+++ b/src/common/gauge.jl
@@ -10,7 +10,7 @@ is real and positive.
 
 # Helper functions
 _argmaxabs(x) = reduce(_largest, x; init = zero(eltype(x)))
-_largest(x, y) = real(abs(x)) < real(abs(y)) ? y : x
+_largest(x, y) = abs(x) < abs(y) ? y : x
 
 function gaugefix!(::typeof(qr_householder!), Q, R, Rd)
     ax = Base.OneTo(length(Rd))
@@ -68,7 +68,8 @@ end
 function gaugefix!(::Union{typeof(svd_compact!), typeof(svd_trunc!)}, U, Vᴴ)
     @assert axes(U, 2) == axes(Vᴴ, 1)
     signs = sign.(reduce(_largest, U; dims = 1, init = zero(eltype(U))))
-    @. U = U * signs
-    @. transpose(Vᴴ) = signs * transpose(Vᴴ)
+    signs_t = transpose(signs)
+    @. U = U * conj(signs)
+    @. Vᴴ = signs_t * Vᴴ
     return (U, Vᴴ)
 end


src/common/gauge.jl
```diff
@@ -10,7 +10,7 @@ is real and positive.
 
 # Helper functions
 _argmaxabs(x) = reduce(_largest, x; init = zero(eltype(x)))
-_largest(x, y) = abs(x) < abs(y) ? y : x
+_largest(x, y) = real(abs(x)) < real(abs(y)) ? y : x
```
lkdvos: Are there cases where `abs` does not output a `real` ?

src/common/gauge.jl
```diff
@@ -67,12 +67,8 @@ end
...
-    for j in axes(U, 2)
-        u = view(U, :, j)
-        v = view(Vᴴ, j, :)
-        s = sign(_argmaxabs(u))
-        u .*= conj(s)
-        v .*= s
-    end
+    signs = sign.(reduce(_largest, U; dims = 1, init = zero(eltype(U))))
+    @. U = U * signs
+    @. transpose(Vᴴ) = signs * transpose(Vᴴ)
```
lkdvos:
```suggestion
    @. Vᴴ = transpose(signs) * Vᴴ
```
Does this give the same result? I am slightly worried about the `transpose` appearing on the left-hand-side, since this implicitly assumes that this is a view of the original, which I'm not sure is guaranteed in general

src/common/gauge.jl
```diff
@@ -67,12 +67,8 @@ end
...
 function gaugefix!(::Union{typeof(svd_compact!), typeof(svd_trunc!)}, U, Vᴴ)
     @assert axes(U, 2) == axes(Vᴴ, 1)
-    for j in axes(U, 2)
-        u = view(U, :, j)
-        v = view(Vᴴ, j, :)
-        s = sign(_argmaxabs(u))
-        u .*= conj(s)
-        v .*= s
-    end
+    signs = sign.(reduce(_largest, U; dims = 1, init = zero(eltype(U))))
```
lkdvos: Might be worth benchmarking, but in principle you could also just compute the vector of largests, and then recompute the `sign` twice below, in order to avoid having to allocate a (second) new vector here. I.e. something like:

```julia
largests = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
U .= U .* sign.(largests)
...
```

lkdvos: edit: does `mapreduce` support the `dims = 1` keyword? Otherwise this could probably just be `mapreduce(sign, _largest, U; dims = 1, init = sign(zero(eltype(U)))` or something similar?

lkdvos (commented):
Looks like a good alternative!
Do you also have benchmarks for the CPU results before and after?
I don't mind too much if this is a slight regression, just want to have an idea of the ballpark.

kshyatt:
I missed the `conj` for the `Us`, fixes incoming...

src/common/gauge.jl
kshyatt: No, because this will apply `sign` **before** `_largest`

src/common/gauge.jl
kshyatt: I think GPUCompiler was flipping out with some spurious IR generation errors about this, can probably revert

src/common/gauge.jl
kshyatt: Actually we'd better not have any `transpose` in the `@.` at all, fixed now :)

kshyatt:
I'm running these on a machine with a good GPU but potato-tier CPU, just keep in mind:

GPU:                                                                                                                                                                                                            
  Old 2.967 ms (3971 allocations: 131.81 KiB)                                                                                                                                                                       
  New 85.828 μs (161 allocations: 6.11 KiB)                                                                                                                                                                         
CPU:                                                                                                                                                                                                            
  Old 99.024 μs (0 allocations: 0 bytes)                                                                                                                                                                            
  New 102.448 μs (4 allocations: 1.16 KiB)

lkdvos:
I guess the extra allocations in the CPU case are negligible for the SVD, which presumably takes a lot more than the 100 microseconds anyways?

kshyatt:
I would think so, we could specialize my new version only for GPU arrays if you prefer no allocs

src/common/gauge.jl
lkdvos: Oh right, forgot about that. The other suggestion might still be worth checking to avoid one more allocation though

kshyatt: We need `reducemap`

kshyatt: OK, did what I could for both broadcasts!

kshyatt:
With the latest (old above new):

GPU:
  2.955 ms (3990 allocations: 132.11 KiB)
  66.249 μs (111 allocations: 4.27 KiB)
CPU:
  99.666 μs (0 allocations: 0 bytes)
  165.045 μs (2 allocations: 592 bytes)

kshyatt:
You don't have to trust me but `decompositions/svd` did pass locally for me 😅

lkdvos:
Wow, I was not expecting that difference to be so large! Looks like the extra allocation is worth it, I forgot that `sign` on complex numbers can actually be expensive...
So, final attempt from my side:

```julia
signs = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
signs .= sign.(signs)
...
```

lkdvos:
> I would think so, we could specialize my new version only for GPU arrays if you prefer no allocs

I'm trying to figure out if it is even worth it to do that, it seems like it could be fine to just eat the one extra allocation here 🤷

kshyatt:
Just from a naming clarity PoV I may keep the `max_us` for the first line in your last snippet if that's ok

lkdvos:
Yeah no worries, the only thing I wanted to try is the in-place sign call

kshyatt:
Thanks @lkdvos for checking this:

```julia
julia> _largest1(x, y) = abs(x) < abs(y) ? y : x
_largest1 (generic function with 1 method)

julia> _largest2(x, y) = abs2(x) < abs2(y) ? y : x
_largest2 (generic function with 1 method)

julia> T = ComplexF64
ComplexF64 (alias for Complex{Float64})

julia> U = randn(T, 100, 100);

julia> @benchmark reduce(_largest1, $U; dims = 1, init = zero(eltype($U)))
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  208.089 μs … 229.534 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     208.909 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   209.349 μs ±   1.414 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

  ▅▅▃▁   ▇█▆▄▃▃▃▂▂▄▄▂▁ ▂▃▂▂▁▁▁▁▂▂▁▁▁▁▁                          ▂
  ██████▇█████████████████████████████▇▇▇▇▆▆▄▆▅▆▆▅▅▄▄▄▇▇▅▃▃▄▂█▇ █
  208 μs        Histogram: log(frequency) by time        214 μs <

 Memory estimate: 1.64 KiB, allocs estimate: 2.

julia> @benchmark reduce(_largest2, $U; dims = 1, init = zero(eltype($U)))
BenchmarkTools.Trial: 10000 samples with 1 evaluation per sample.
 Range (min … max):  35.305 μs …  53.111 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     35.336 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):   35.999 μs ± 948.177 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █▂ ▁          ▂▁     ▅▆▄▃▂▁ ▁   ▂▁  ▁▁                       ▂
  ██▇█▅▄▅▄▄▄▄▅▁▃██▆▅▅▅▄██████████████████▇▇▇▇▆▅▆▅▆▅▄▅▄▅▅▅▅▄▅▅▅ █
  35.3 μs       Histogram: log(frequency) by time      38.8 μs <

 Memory estimate: 1.64 KiB, allocs estimate: 2.
```

For `Real`, the choice of `abs` vs `abs2` doesn't matter much, but for `Complex`, `abs2` is notably better

lkdvos (approved):

src/common/gauge.jl
```diff
@@ -67,12 +68,10 @@ end
...
-        v = view(Vᴴ, j, :)
-        s = sign(_argmaxabs(u))
-        u .*= conj(s)
-        v .*= s
-    end
+    signs = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
+    @. signs = sign(signs)
+    signs_t = transpose(signs)
+    @. U = U * conj(signs)
+    @. Vᴴ = signs_t * Vᴴ
```
Jutho, line 75: I guess we can do with one less allocation by also doing the `sign` calculation in the broadcast, though then the variable should probably not be called `signs`.

Something like

```suggestion
    largestentries = reduce(_largest, U; dims = 1, init = zero(eltype(U)))
    largestentries_t = transpose(largestentries)
    @. U = U * conj(sign(largestentries))
    @. Vᴴ = sign(largestentries_t) * Vᴴ
```

But maybe the doubled `sign` calculation makes this slower. In principle, I tend to also use `safesign` since `sign(0.) = 0.`, but I guess this is no problem here, as the largest entry of a column of an orthogonal matrix can never be zero (and indeed, the original code was also using `sign`).

kshyatt: I think Lukas and I addressed this in a previously resolved comment

kshyatt: The short story is that doubled computation of `sign` is annoyingly expensive for complex elements (complex division is pricey)

kshyatt:
Enzyme failures look unrelated to me, I'll try to look into this later today

src/common/gauge.jl
lkdvos: Great minds think alike, but apparently without actually finding useful stuff :)

kshyatt: Well, it was worth a try!

kshyatt:
OK, the enzyme failures seemingly have nothing to do with this PR so I will merge it and try to fix separately.

PR #239, opened 2026-05-27
Fix check of whether to copy arg for Enzyme

kshyatt:


kshyatt:
Verified this fixed the enzyme tests locally. I'll try to work out where the problem is for newer versions and fix it but in the meantime this should help CI.

lkdvos (commented):
I understand this fixes the CI, but I think effectively this means that if you for whatever reason have a newer version of Enzyme, it will just revert to an older version of MatrixAlgebraKit that is compatible with that. I'm not sure what the recommended way around this is, does this mean we should just kind of wait for Enzyme to unbreak?

kshyatt:
I also opened an issue over there and am trying to figure out the cause at the moment. I was mostly thinking we could effectively revert this once the problem is fixed but for now it won't spuriously break CI for other people 🤷

kshyatt:
But maybe it's unnecessary as I think I found the problem...

kshyatt:
OK, it seems on more recent versions of Enzyme, this `arg.val === ret` check (which we originally added for the case when `arg = (nothing, nothing)`) varies depending on the Enzyme version. I think it doesn't make sense to check this alongside the one for whether `A` is present in `arg` so I've modified that and it seems to make everyone happy.

lkdvos:
I have very little feeling with what is the correct way of doing this since I don't know the Enzyme details, this definitely looks reasonable. Is there any benefit of factoring that out into a separate function, which is then at least centralized?

kshyatt:
> Is there any benefit of factoring that out into a separate function, which is then at least centralized?

There may be but I think there is a bunch of refactoring of these rules I'd like to do in a separate PR (e.g. handle all the code duplication in the truncated rules)

PR #241, opened 2026-06-01
Test QR rules with CUDA

kshyatt:


kshyatt:
OK, this is now working for QR on full `CuMatrix`, but there are still scalar indexing errors for `Diagonal` because dispatch for `SubArray{T, 2, Diagonal{T, CuVector{T}}` is awful. I'd be ok with merging this as-is if the pullback modifications look ok, and then come back for another swing at the `Diagonal` case.

src/pullbacks/qr.jl
```diff
@@ -31,9 +31,18 @@ function check_and_prepare_qr_cotangents(
...
-        Δgauge_R = norm(view(ΔR₂₂, uppertriangularind(ΔR₂₂)), Inf)
-        Δgauge_R = max(Δgauge_R, norm(view(ΔR₂₂, diagind(ΔR₂₂)), Inf))
-        Δgauge = max(Δgauge, Δgauge_R)
+        if p < minmn # otherwise ΔR₂₂ is empty
+            # uppertriangularind generates linear indices
+            # compute the appropriate offset in ΔR so we aren't
+            # operating on a view-of-view, which doesn't work
+            # for GPU arrays
+            offset = LinearIndices(ΔR)[p + 1, p + 1]
+            upper_inds = uppertriangularind(ΔR₂₂) .+ offset
```
Jutho: I would expect some `-1` to appear here. The current `offset` is not really an offset, but the first index. Put differently, if `p=0`, we should just have `uppertriangularind = uppertriangularind(ΔR₂₂)`. But `offset` would be `1` in this case. 

Actually, looking into this further, this doesn't work at all. You cannot simply take `uppertriangularind(ΔR₂₂)` and compute from this the corresponing indices in `ΔR` by just adding the same offset to all indices. It depends on the columns that they are in. Let me think about this.

kshyatt: What's crazy then is that it works for Mooncake 😬 . The perils of doing this sleep deprived...

kshyatt: The problem is that these pullbacks constantly create views of wrappers of views which just doesn't work on the GPU arrays because we don't have dispatch for it. Either we need to support this in StridedView or just convert these to proper arrays and allocate.

Jutho: No problem; I didn't know views of views don't work on GPU, so that's why I did it like this. `uppertriangularind` is not a `StridedView`, so StridedViews.jl cannot help. I'll also think about other solutions.

kshyatt: Yeah, it's super frustrating! I may see if I can tweak things in GPUArrays to make these views-of-wrapper-of-views work, Tim may be sad about it because of compiler burden but ✨ in theory ✨ it should work

kshyatt: I think "view of view" is fine -- the problem comes when you have something like `SubArray{T, 2, UpperTriangular{T, SubArray{T, 2, CuArray{T, 2}}}}` which is a dispatch nightmare (how do you detect this is "really" a GPU array?)

lkdvos (commented):
Would it make sense to introduce a `project_triu(!)` and/or `norm_triu` function for this, and just overload the GPU ones to allocate?

As a side note, we might at some point try and not generate any of the gauge dependency checks if the tolerance is set to 0, as a way to disable the warnings, or alternatively put the entire thing in a `@warn` block so we can disable that, since these checks are starting to feel a bit expensive if we need to start allocating.

kshyatt:
> Would it make sense to introduce a project_triu(!) and/or norm_triu function for this, and just overload the GPU ones to allocate?

Yeah I think that could be a nice way to handle it, however I'm unsure if it's that helpful if it's only used in these checks.

src/pullbacks/qr.jl
Jutho: It seems it must be view of view, since `ΔR₂₂ = view(ΔR, (p + 1):minmn, (p + 1):n)` is not explicitly wrapped in an `UpperTriangular` array. The original lines directly contained:

```julia
ΔR₂₂ = view(ΔR, (p + 1):minmn, (p + 1):n)
do_something_with( view(ΔR₂₂, uppertriangularind(ΔR₂₂)) )
```

Jutho: Would something like this work:
```julia
I = uppertriangularind(ΔR₂₂)
upper_inds = view(LinearIndices(ΔR), (p + 1):minmn, (p + 1):n)[I]
ΔR₂₂upper = view(ΔR, upper_inds)
```

kshyatt: I'll give it a try, thanks for the suggestion

kshyatt: Seems to have helped here, now we have a problem elsewhere in the pullback!

src/pullbacks/qr.jl
```diff
@@ -160,7 +170,16 @@ function remove_qr_gauge_dependence!(ΔQ, ΔR, A, Q, R; rank_atol = MatrixAlgebr
...
     ΔR₂₂ = view(ΔR, (r + 1):minmn, (r + 1):size(R, 2))
     zero!(diagview(ΔR₂₂))
-    zero!(view(ΔR₂₂, uppertriangularind(ΔR₂₂)))
+    if r < minmn
+        # uppertriangularind generates linear indices
+        # compute the appropriate offset in ΔR so we aren't
+        # operating on a view-of-view, which doesn't work
+        # for GPU arrays
+        offset = LinearIndices(ΔR)[r + 1, r + 1]
+        upper_inds = uppertriangularind(ΔR₂₂) .+ offset
```
Jutho: This is still using the old (faulty) strategy, so it needs to be updated in the same way as in the gauge check above. Maybe that is the origin of the remaining failures.

src/pullbacks/qr.jl
```diff
@@ -75,7 +84,7 @@ function qr_pullback!(
 
 
     Q₁ = view(Q, :, 1:p)
-    R₁₁ = UpperTriangular(view(R, 1:p, 1:p))
+    R₁₁ = UpperTriangular(R[1:p, 1:p])
```
Jutho: This is a subtle and impactful change. The `UpperTriangular` wrapper is really only necessary to enable the `rdiv!` call below. If GPUs cannot deal with `UpperTriangular` of a view of a `GPUArray`, then maybe we need to call the corresponding BLAS/LAPACK methods directly, or have some intermediate wrapper like `rdiv_uppertriangular!`.

kshyatt: > If GPUs cannot deal with UpperTriangular of a view of a GPUArray

Indeed they can't :(

Jutho: I even wonder how `rdiv!(::Matrix, ::UpperTriangular)` is evaluated on the GPU, since you need cuSOLVERDx to access TRSM.

kshyatt: Through https://docs.nvidia.com/cuda/cublas/#cublas-t-trsm I think

kshyatt: BTW `cuSOLVERDx` is only for *device side* code, so it can only be called by running CUDA kernels, not from *host side* code as we are doing here...

kshyatt:
The Enzyme windows fail is happening on `main` and seems unrelated

src/pullbacks/qr.jl
Jutho: Strange that I did not discover cublas trsm when I was looking for it. Anyway, I think it would be useful to keep the view if possible. I don't actually see the code pathway that `rdiv!(CuMatrix, UpperTriangular(CuArray(...)))` tales to reach CUBLAS TRSM, so I don't understand how it works now (it is hard to test this without access to a GPU).

src/pullbacks/qr.jl
```diff
@@ -160,7 +171,20 @@ function remove_qr_gauge_dependence!(ΔQ, ΔR, A, Q, R; rank_atol = MatrixAlgebr
...
-    zero!(view(ΔR₂₂, uppertriangularind(ΔR₂₂)))
+    if r < minmn
+        # uppertriangularind generates linear indices
+        # compute the appropriate offset in ΔR so we aren't
+        # operating on a view-of-view, which doesn't work
+        # for GPU arrays
+        # uppertriangularind generates linear indices
+        # compute the appropriate offset in ΔR so we aren't
+        # operating on a view-of-view, which doesn't work
+        # for GPU arrays
```
Jutho, line 182: Comment seems to be duplicated.

src/pullbacks/qr.jl
kshyatt: https://github.com/JuliaGPU/CUDA.jl/blob/fbb90981cbde21d979087ad518a510f5b38f95b3/lib/cusolver/src/linalg.jl#L43 is this what you're looking for?

Jutho: Not really, this is internally in dividing using `\` by a general matrix, which is then evaluated by computing its QR decomposition and then directly calling `cuBLAS.trsm!` on the triangular factor. Here we already have the triangular factor, so I indeed want to call `trsm!`, but using generic code, which is why I was using `ldiv!`/`rdiv!`. And I don't see where `rdiv!(first_arg, second_arg::UpperTrangiular)` is then actually lowered to `cuBLAS.trsm!` in the CuArray case (and why that only works for a pure CuMatrix and not for a view over it.)

kshyatt: I think I figured it out, it's a combo of:

https://github.com/JuliaLang/LinearAlgebra.jl/blob/1dcf75c70ea5a8d518ce71efe04c6c1e2093628d/src/triangular.jl#L1263

and 

https://github.com/JuliaGPU/CUDA.jl/blob/fbb90981cbde21d979087ad518a510f5b38f95b3/lib/cublas/src/linalg.jl#L443

lkdvos: To unblock this: shall we indeed create a helper `rdiv!_uppertriangular!` that for now just avoids the copy on the CPU and simply copies on the GPU with a `# TODO: dispatch to trsm! directly`

Jutho: But Then I don't understand why it doesn't work. The argument `B::StridedCuMatrix` in https://github.com/JuliaGPU/CUDA.jl/blob/fbb90981cbde21d979087ad518a510f5b38f95b3/lib/cublas/src/linalg.jl#L443 should accept a view of a `CuMatrix`, no? My apologies for being annoying, my lack of access to a GPU to test these things myself makes me ask these questions.

kshyatt: Sorry that I left this by the wayside. @jutho it's **not** annoying at all so please don't hesitate to ask such questions :). [Here](https://github.com/JuliaGPU/CUDA.jl/blob/5d78ab5e456a64999931883222b95e2e7cb31d6e/CUDACore/src/array.jl#L391) is the definition of `StridedCuMatrix`, but I think the problem we have is this doesn't accept `UpperTriangular{T, <:StridedCuMatrix{T}}`

kshyatt: So either one must create an `UpperTriangular` of a straight `CuMatrix` OR create a "pure" view

Jutho: No I am still confused. The `LinearAlgebra._rdiv!` lowers to `generic_mattridiv!`, but also unwraps the `UpperTriangular` argument, so at that point it is just `generic_mattrivdiv!` with the `parent` of the `UpperTriangular`, and there are no more explicit `Triangular` wrappers present anymore. And then the implementation of `generic_mattrivdiv!` in CUDA takes over, and that does accept a `StridedCuMatrix`, no?

kshyatt: Oh I see what's wrong, it's that additionally we use `rdiv!(ΔQ₁, R₁₁') ` so the type of the RHS is 

`LinearAlgebra.LowerTriangular{ComplexF64, LinearAlgebra.Adjoint{ComplexF64, SubArray{ComplexF64, 2, CUDACore.CuArray{ComplexF64, 2, CUDACore.DeviceMemory}`

which is too much unwrapping :(

Jutho: No that cannot be it; the `_rdiv!` is

```
_rdiv!(C::AbstractMatrix, A::AbstractMatrix, B::UpperOrLowerTriangular) =
    generic_mattridiv!(C, uplo_char(B), isunit_char(B), _wrapperop(parent(B)), A, _unwrap_at(parent(B)))
```

and the `_wrapperop(parent(B))` extracts the `adjoint`, whereas `_unwrap_at(parent(B))` returns the actual `view(::CuArray)`.

kshyatt: 🤷 

This is what's being called in the stack trace if we restore what was there previously with the view

```julia
rdiv!(A::CUDACore.CuArray{Float64, 2, CUDACore.DeviceMemory}, B::LinearAlgebra.LowerTriangular{Float64, LinearAlgebra.Adjoint{Float64, SubArray{Float64, 2, CUDACore.CuArray{Float64, 2, CUDACore.DeviceMemory}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}}})                                                                                           
      @ LinearAlgebra ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/LinearAlgebra/src/triangular.jl:1184   
```

kshyatt: I think it's actually `istriu` that's hurting us due to the nested types

kshyatt: OK, testing locally and using internal `_rdiv!`, which bypasses the `istriu` check which doesn't dispatch nicely, everything seems to work well 🎉 with the view 🎉

Jutho: This is ridiculous:
```julia
function rdiv!(A::AbstractMatrix, B::AbstractTriangular)
    if istriu(B)
        _rdiv!(A, A, uppertriangular(B))
    else
        _rdiv!(A, A, lowertriangular(B))
    end
end
```
So `istriu` returns true for `UpperTriangular` and `UnitUpperTriangular`, but for `(Unit)LowerTrianguar`, rather than returning `false`, it goes via the generic `istriu` for `AbstractMatrix` check, checking that all elements are zero. But in this function in particular, you already know that `B` is `AbstractTriangular`, so really there is no point in doing this.

Jutho: This kind of shenanigans is exactly why I don't like LinearAlgebra.jl

kshyatt: Yes, exactly. You can have the stupid and annoying case where the `LowerTriangular` has all zeros except on diagonal 0 so it would in fact be `triu`. But here we have extra info, so we can bypass. The risk of course is the internal API changes...

kshyatt: Shenanigans, or hijinx? 😇

Jutho: This goes beyond my English vocabulary. I will need to consult with the person saying "I know words. I have the best words".

kshyatt: - Shenanigans
- Hijinx
- Tomfoolery

Trying to think of others...

Jutho: In any case, I think this needs to be reported, I think this could lead to silent errors even, if you make a `LowerTriangular` wrapper of a matrix that indeed happens to be zero below the diagonal, but not above the diagonal (so the storage is in fact upper triangular, whereas `LowerTriangular(storage)` is essentially diagonal).

Jutho: I'd be happy to tap into `generic_mattridiv!(C, uplo_char(B), isunit_char(B), _wrapperop(parent(B)), A, _unwrap_at(parent(B)))` for now, even if this is not a public function.

kshyatt: Hm, so do we want to keep `_rdiv!`, or change to `generic_mattridiv!`?

Jutho: I am fine either way. `_rdiv!` feels a bit more internal than `generic_mattridiv!`, but the latter has a more complicated signature, so `_rdiv!` probably is the easier solution. Maybe add a `TODO` comment to change this at some point in the future when LinearAlgebra.jl gets its **** together.

Jutho (approved):

PR #242, opened 2026-06-03
Forward AD rules and tests for polar decompositions

kshyatt:
This seems to just work out of the box without hacking at special tangents or anything, so why not add...

### DIFF
--- a/src/pushforwards/polar.jl
+++ b/src/pushforwards/polar.jl
@@ -1,0 +1,21 @@
+function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
+    W, P = WP
+    ΔW, ΔP = ΔWP
+    aWdA = adjoint(W) * ΔA
+    K̇ = sylvester(P, P, -(aWdA - adjoint(aWdA)))
+    L̇ = (Diagonal(ones(eltype(W), size(W, 1))) - W * adjoint(W)) * ΔA * inv(P)
+    ΔW .= W * K̇ + L̇
+    ΔP .= aWdA - K̇ * P
+    return (ΔW, ΔP)
+end
+
+function right_polar_pushforward!(ΔA, A, PWᴴ, ΔPWᴴ; kwargs...)
+    P, Wᴴ = PWᴴ
+    ΔP, ΔWᴴ = ΔPWᴴ
+    dAW = ΔA * adjoint(Wᴴ)
+    K̇ = sylvester(P, P, -(dAW - adjoint(dAW)))
+    L̇ = inv(P) * ΔA * (Diagonal(ones(eltype(Wᴴ), size(Wᴴ, 2))) - adjoint(Wᴴ) * Wᴴ)
+    ΔWᴴ .= K̇ * Wᴴ + L̇
+    ΔP .= dAW - P * K̇
+    return (ΔWᴴ, ΔP)
+end
--- a/src/pushforwards/polar.jl
+++ b/src/pushforwards/polar.jl
@@ -3,7 +3,9 @@ function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
     ΔW, ΔP = ΔWP
     WᴴdA = adjoint(W) * ΔA
     K̇ = _sylvester(P, P, -(WᴴdA - adjoint(WᴴdA)))
-    L̇ = (LinearAlgebra.UniformScaling(1) - W * adjoint(W)) * ΔA * inv(P)
+    dAiP = ΔA * inv(P)
+    WᴴdAiP = W' * dAiP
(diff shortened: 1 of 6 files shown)


src/pushforwards/polar.jl
```diff
@@ -0,0 +1,21 @@
+function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
+    W, P = WP
+    ΔW, ΔP = ΔWP
+    aWdA = adjoint(W) * ΔA
```
Jutho: Can we call this variable 

```suggestion
     WᴴdA = adjoint(W) * ΔA
```

I think that, whenever I have a prefix `a` in the variables in the pullbacks, it is for antihermitian part of. But you only take the antihermitian part on the next line, and not in place, which seems correct because you still need the full value in the computation of `ΔP` below.

kshyatt:
🎉 CUDA tests are passing 🎉

src/pushforwards/polar.jl
```diff
@@ -1,11 +1,11 @@
 function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
     W, P = WP
     ΔW, ΔP = ΔWP
-    aWdA = adjoint(W) * ΔA
-    K̇ = _sylvester(P, P, -(aWdA - adjoint(aWdA)))
-    L̇ = (Diagonal(ones(eltype(W), size(W, 1))) - W * adjoint(W)) * ΔA * inv(P)
+    WᴴdA = adjoint(W) * ΔA
+    K̇ = _sylvester(P, P, -(WᴴdA - adjoint(WᴴdA)))
+    L̇ = (LinearAlgebra.UniformScaling(1) - W * adjoint(W)) * ΔA * inv(P)
```
Jutho: One slightly longer but more memory efficient way of doing this is:

```suggestion
    dAiP = ΔA * inv(P)
    WᴴdAiP = W' * dAiP
    L̇ = mul!(dAiP, W, WᴴdAiP, -1, +1)
```

kshyatt: Done here and below

src/pushforwards/polar.jl
```diff
@@ -0,0 +1,25 @@
+function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
+    W, P = WP
+    ΔW, ΔP = ΔWP
+    WᴴdA = adjoint(W) * ΔA
+    K̇ = _sylvester(P, P, -(WᴴdA - adjoint(WᴴdA)))
+    dAiP = ΔA * inv(P)
+    WᴴdAiP = W' * dAiP
+    L̇ = mul!(dAiP, W, WᴴdAiP, -1, +1)
+    ΔW .= W * K̇ + L̇
```
lkdvos: I think here we can also save a couple allocations if dAiP shares memory with DeltaW, and then W * Kdot is an inplace adding mul! 

(From my phone, otherwise I'd suggest the full code 😉)

src/pushforwards/polar.jl
```diff
@@ -0,0 +1,25 @@
+function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
+    W, P = WP
+    ΔW, ΔP = ΔWP
+    WᴴdA = adjoint(W) * ΔA
+    K̇ = _sylvester(P, P, -(WᴴdA - adjoint(WᴴdA)))
```
lkdvos:
```suggestion
    K̇ = _sylvester(P, P, adjoint(WᴴdA) - WᴴdA))
```
I think this avoids another allocation.

src/pushforwards/polar.jl
```diff
@@ -0,0 +1,25 @@
+function left_polar_pushforward!(ΔA, A, WP, ΔWP; kwargs...)
+    W, P = WP
+    ΔW, ΔP = ΔWP
+    WᴴdA = adjoint(W) * ΔA
```
lkdvos: I think you can immediately write this into DeltaP, and then in the last line of this function do another in place muladd

lkdvos (commented):
Left some suggestions about performance but otherwise looks great!!

kshyatt:
> Left some suggestions about performance but otherwise looks great!!

I wrote this in a less performant way but clearer to me way at the start just to make sure it was correct, thanks for the suggestions! Think I got everything in both push-forwards.

lkdvos (approved):
It might even be worth it to add a comment to the implementations with the actual math for the formulas, you are right that it is way harder to follow what is going on now

kshyatt:
We don't have that right now for other pullbacks so I am ok with this as-is, would be good to go through all of them at some point and add this though. I think the fail is unrelated, I'm going to merge.

PR #244, opened 2026-06-04
Forward rules and tests for project functions

kshyatt:


lkdvos (commented):
Looks good to me! How would you feel about splitting the testsets into forwards and backwards subsets? I think it might be useful to see which is failing, but this is purely cosmetic

kshyatt:
For enzyme this is already the case, no? For Mooncake I really don’t have
an opinion, happy to do it if you think it would be clearer

lkdvos:
I was thinking about the `@testset` split, but maybe it doesn't matter since indeed the line numbers should tell the diagnostics

lkdvos (approved):

kshyatt:
Enzyme fail unrelated, I think. We really need a way to exclude these AD runs on `macos`

PR #247, opened 2026-06-08
Some basic svd forward rules and tests

kshyatt:
Definitely not optimized...

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -8,7 +8,7 @@ using MatrixAlgebraKit: CUSOLVER, LQViaTransposedQR, TruncationByValue, Abstract
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eig_algorithm, default_eigh_algorithm
 import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdp!, gesvdr!, gesvdj!
 import MatrixAlgebraKit: heevj!, heevd!, geev!
-import MatrixAlgebraKit: _gpu_Xgesvdr!, _sylvester, svd_rank, svd_pullback!, eigh_pullback!, eig_pullback!
+import MatrixAlgebraKit: _gpu_Xgesvdr!, _sylvester, svd_rank, svd_pullback!, eigh_pullback!, eig_pullback!, svd_pushforward!
 using CUDA, CUDA.cuBLAS
 using CUDA: i32
 using LinearAlgebra
@@ -211,6 +211,16 @@ end
 
 function eig_pullback!(ΔA::AnyCuMatrix, A, DV, ΔDV, ind::AnyCuVector; kwargs...)
     return eig_pullback!(ΔA, A, DV, ΔDV, collect(ind); kwargs...)
+end
+
+# have to override this as methods are missing in GPUArrays for the various
+# views of Diagonal of ΔA
+function svd_pushforward!(
+        ΔA::Diagonal{T, <:CuVector{T}}, A, USVᴴ, ΔUSVᴴ, ind = Colon();
+        rank_atol::Real = MatrixAlgebraKit.default_pullback_rank_atol(USVᴴ[2]),
+        degeneracy_atol::Real = MatrixAlgebraKit.default_pullback_rank_atol(USVᴴ[2])
+    ) where {T}
+    return MatrixAlgebraKit.svd_pushforward!(diagm(diagview(ΔA)), A, USVᴴ, ΔUSVᴴ, ind; rank_atol, degeneracy_atol)
 end
 
 end
--- a/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
+++ b/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
@@ -210,11 +210,11 @@ for (f, pb) in (
     end
 end
 
-for f in (:svd_compact!, :svd_full!)
+for f! in (:svd_compact!, :svd_full!)
     @eval begin
         function EnzymeRules.augmented_primal(
                 config::EnzymeRules.RevConfigWidth{1},
-                func::Const{typeof($f)},
+                func::Const{typeof($f!)},
                 ::Type{RT},
                 A::Annotation,
                 USVᴴ::Annotation{Tuple{TU, TS, TVᴴ}},
@@ -232,7 +232,7 @@ for f in (:svd_compact!, :svd_full!)
             dret = if EnzymeRules.needs_shadow(config) && ((TU == TS == TVᴴ == Nothing) || isa(USVᴴ, Const))
                 dU = zero(ret[1])
                 # special casing `Diagonal` seems to be necessary due to Enzyme's type analysis
-                dS = $(f == svd_compact!) ? Diagonal(zero(ret[2].diag)) : zero(ret[2])
+                dS = $(f! == svd_compact!) ? Diagonal(zero(ret[2].diag)) : zero(ret[2])
                 dVᴴ = zero(ret[3])
                 (dU, dS, dVᴴ)
             elseif EnzymeRules.needs_shadow(config)
@@ -245,7 +245,7 @@ for f in (:svd_compact!, :svd_full!)
         end
         function EnzymeRules.reverse(
                 config::EnzymeRules.RevConfigWidth{1},
-                func::Const{typeof($f)},
+                func::Const{typeof($f!)},
                 ::Type{RT},
                 cache,
                 A::Annotation,
@@ -267,21 +267,17 @@ for f in (:svd_compact!, :svd_full!)
         end
         function EnzymeRules.forward(
                 config::EnzymeRules.FwdConfigWidth{1},
-                func::Const{typeof($f)},
+                func::Const{typeof($f!)},
                 ::Type{RT},
                 A::Annotation{TA},
                 USVᴴ::Annotation,
                 alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
             ) where {RT, TA}
-            $f(A.val, USVᴴ.val, alg.val)
+            $f!(A.val, USVᴴ.val, alg.val)
+            A_is_arg = !isa(A, Const) && TA <: Diagonal && diagview(A.dval) === USVᴴ.dval[2]
             if !isa(A, Const)
-                if $(f == svd_compact!)
-                    make_zero!(USVᴴ.dval[2].diag)
-                else
-                    make_zero!(USVᴴ.dval[2])
-                end
                 !isa(USVᴴ, Const) && svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
-                make_zero!(A.dval)
+                !A_is_arg && make_zero!(A.dval)
             end
             if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
                 return USVᴴ
--- a/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
+++ b/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
@@ -275,7 +275,7 @@ for f in (:svd_compact!, :svd_full!)
             ) where {RT, TA}
             $f(A.val, USVᴴ.val, alg.val)
             if !isa(A, Const)
-                if $(f == svd_compact!) 
+                if $(f == svd_compact!)
                     make_zero!(USVᴴ.dval[2].diag)
                 else
                     make_zero!(USVᴴ.dval[2])
--- a/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
+++ b/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
@@ -274,8 +274,10 @@ for f in (:svd_compact!, :svd_full!)
                 alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
             ) where {RT, TA}
             $f(A.val, USVᴴ.val, alg.val)
-            !isa(A, Const) && !isa(USVᴴ, Const) && svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
-            make_zero!(A.dval)
+            if !isa(A, Const)
+                !isa(USVᴴ, Const) && svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
+                make_zero!(A.dval)
+            end
             if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
                 return USVᴴ
             elseif EnzymeRules.needs_primal(config)
--- a/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
+++ b/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
@@ -269,16 +269,13 @@ for f in (:svd_compact!, :svd_full!)
                 config::EnzymeRules.FwdConfigWidth{1},
                 func::Const{typeof($f)},
                 ::Type{RT},
-                A::Annotation,
+                A::Annotation{TA},
                 USVᴴ::Annotation,
                 alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
-            ) where {RT}
+            ) where {RT, TA}
             $f(A.val, USVᴴ.val, alg.val)
-            if !isa(A, Const) && !isa(USVᴴ, Const)
-                make_zero!(USVᴴ.dval)
-                svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
-            end
-            !isa(A, Const) && make_zero!(A.dval)
+            !isa(A, Const) && !isa(USVᴴ, Const) && svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
+            make_zero!(A.dval)
             if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
                 return USVᴴ
             elseif EnzymeRules.needs_primal(config)
@@ -542,7 +539,7 @@ function EnzymeRules.forward(
         svd_vals_pushforward!(A.dval, A.val, (U, Diagonal(diagview(S_)), Vᴴ), ΔS)
         A_is_arg && (S.dval .= ΔS)
     end
-    !isa(A, Const) && !A_is_arg && make_zero!(A.dval)
+    !A_is_arg && make_zero!(A.dval)
     copyto!(S.val, diagview(S_))
     if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
         return S
--- a/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
+++ b/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
@@ -278,7 +278,7 @@ for f in (:svd_compact!, :svd_full!)
                 make_zero!(USVᴴ.dval)
                 svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
             end
-            #!isa(A, Const) && make_zero!(A.dval)
+            !isa(A, Const) && make_zero!(A.dval)
             if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
                 return USVᴴ
             elseif EnzymeRules.needs_primal(config)
@@ -537,13 +537,13 @@ function EnzymeRules.forward(
     ) where {RT, TA}
     A_is_arg = !isa(A, Const) && TA <: Diagonal && diagview(A.dval) === S.dval
     U, S_, Vᴴ = svd_compact!(A.val, alg.val)
-    copyto!(S.val, diagview(S_))
     if !isa(A, Const) && !isa(S, Const)
         ΔS = A_is_arg ? make_zero(S.dval) : S.dval
-        svd_vals_pushforward!(A.dval, A.val, (U, Diagonal(S.val), Vᴴ), ΔS)
+        svd_vals_pushforward!(A.dval, A.val, (U, Diagonal(diagview(S_)), Vᴴ), ΔS)
         A_is_arg && (S.dval .= ΔS)
     end
     !isa(A, Const) && !A_is_arg && make_zero!(A.dval)
+    copyto!(S.val, diagview(S_))
     if EnzymeRules.needs_primal(config) && EnzymeRules.needs_shadow(config)
         return S
     elseif EnzymeRules.needs_primal(config)
--- a/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
+++ b/ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
@@ -275,6 +275,11 @@ for f in (:svd_compact!, :svd_full!)
             ) where {RT, TA}
             $f(A.val, USVᴴ.val, alg.val)
             if !isa(A, Const)
+                if $(f == svd_compact!) 
+                    make_zero!(USVᴴ.dval[2].diag)
+                else
+                    make_zero!(USVᴴ.dval[2])
+                end
                 !isa(USVᴴ, Const) && svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
                 make_zero!(A.dval)
             end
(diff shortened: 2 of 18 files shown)


src/pushforwards/svd.jl
```diff
@@ -0,0 +1,91 @@
...
+
+    vU = view(U, :, 1:r)
+    vS = view(S, 1:r)
+    vSmat = view(Smat, 1:r, 1:r)
+    vVᴴ = view(Vᴴ, 1:r, :)
+
+    # compact region
+    vV = adjoint(vVᴴ)
+    UΔAV = vU' * ΔA * vV
+    copyto!(diagview(vΔS), diag(real.(UΔAV)))
```
Jutho:
```suggestion
    copyto!(diagview(vΔS), real.(diagview(UΔAV)))
```

I think we need a `realcopy!` one of these days 😄 .

kshyatt:
Took another look at this. The Enzyme tests seem to be failing because of finite differences, for `svd_full` for example when `size(A) == (17, 19)`, the finite differences result for `dU` is all over the place compared to that from the rule **for the section of `U` which is only present in the `full` case**. I'll try to think of a nice way to handle this, I think it is not occurring for Mooncake because Mooncake uses a different technique for checking against FD.

src/pushforwards/svd.jl
```diff
@@ -0,0 +1,91 @@
+function svd_pushforward!(ΔA, A, USVᴴ, ΔUSVᴴ, ind = Colon(); rank_atol = default_pullback_rank_atol(A), kwargs...)
+    U, Smat, Vᴴ = USVᴴ
+    m, n = size(U, 1), size(Vᴴ, 2)
+    (m, n) == size(ΔA) || throw(DimensionMismatch("size of ΔA ($(size(ΔA))) does not match size of U*S*Vᴴ ($m, $n)"))
+    minmn = min(m, n)
+    S = diagview(Smat)
+    ΔU, ΔS, ΔVᴴ = ΔUSVᴴ
+    r = svd_rank(S; rank_atol)
+
+    vΔS = view(ΔS, 1:r, 1:r)
```
lkdvos: I think in general we only need the diagonal, so it might make sense to go the other way around: `view(diagview(ΔS), 1:r)`

src/common/initialization.jl
```diff
@@ -15,6 +15,12 @@ function one!(A::AbstractMatrix)
     return A
 end
 
+function one!(A::AbstractVector)
```
Jutho: I don't quite like this one. `one` in Julia is the multiplicative identity. This doesn't make sense on a vector, since you cannot multiply them. Where do you use / need this?

kshyatt: In the construction of `diagm` in the pushforward for SVD to avoid creating a CPU side array

kshyatt: Maybe that's the wrong place to fix this, though, and we could instead make our own `_eye` function or similar? Actually, let me see if `UniformScaling` now works...

kshyatt: (When I first was trying these pushforwards, `UniformScaling` really didn't work with GPUArrays but I think this has now been fixed)

src/pushforwards/svd.jl
```diff
@@ -0,0 +1,91 @@
...
+        view(UÃÃV, size(aUAV, 1) .+ (1:size(aUAV, 2)), 1:size(aUAV, 1)) .= aUAV'
+        rhs = vcat(adjoint(Uperp * ΔA * Vᴴ), Vᴴperp * ΔA' * U)
+        superKM = -_sylvester(UÃÃV, Smat, rhs)
+        K̇perp = view(superKM, 1:size(aUAV, 2))
+        Ṁperp = view(superKM, (size(aUAV, 2) + 1):(size(aUAV, 1) + size(aUAV, 2)))
+        ∂U .+= Uperp * K̇perp
+        ∂V .+= Vᴴperp * Ṁperp
+    else
+        ImUU = (LinearAlgebra.diagm(one!(similar(U, m))) - vU * vU')
+        ImVV = (LinearAlgebra.diagm(one!(similar(Vᴴ, n))) - vV * vVᴴ)
```
Jutho: Why not `LinearAlgebra.I - vU * vU'`. Also, it can help to recycle a few quantities. Above, you already have `UΔAV = vU' * ΔA * vV`. So by storing the temporary `ΔAV = ΔA * vV`, you can for example do `upper = ΔAV - vU * UΔAV` and so forth. Do you mind if I try to push a few changes to this PR?

kshyatt: I already did the first suggestion :).

kshyatt: But go ahead with further changes, like the others, this one wasn't very optimized as I was trying to make sure it was more or less correct first.

Jutho: Ok. I will check out this branch and try to make some changes if you are ok with this (and as long as ICE 🚋  keeps giving me internet 😄 ).

kshyatt: Sounds good to me!

src/pushforwards/svd.jl
```diff
@@ -7,72 +7,58 @@ function svd_pushforward!(ΔA, A, USVᴴ, ΔUSVᴴ, ind = Colon(); rank_atol = d
...
+    UᴴΔAV₁ = U₁' * ΔAV₁
+    if !iszerotangent(ΔS)
+        ΔS₁ = view(diagview(ΔS), 1:r)
+        ΔS₁ .= real.(diagview(UᴴΔAV₁))
     end
-    if !iszerotangent(ΔVᴴ)
-        vΔVᴴ = view(ΔVᴴ, 1:r, :)
-        adjoint!(vΔVᴴ, ∂V)
+    if !iszerotangent(ΔU) || !iszerotangent(ΔVᴴ)
+        hUᴴΔAV₁ = inv_safe.(transpose(S₁) .- S₁) .* project_hermitian(UᴴΔAV₁)
```
kshyatt: We might want to do this on two separate lines and use broadcast fusion?

Jutho: Isn't this fusing now? Or do you mean that we do not call `project_hermitian` to avoid one more allocation, i.e.

`hUᴴΔAV₁ = inv_safe.(transpose(S₁) .- S₁) .* (UᴴΔAV₁ .+ UᴴΔAV₁') ./ 2`

Jutho:
In the gauge fixing parts, we could use more views for the slicing `ΔU₁[I]` etc, but I did not want to push my luck on the GPU with views into views using CartesianIndices.

Jutho:
Anyway, I will leave this aside now for a while and look at some other PRs. In principle, tests can now be expanded to cover the complex case and `svd_compact` of rectangular matrices as well (but as always, I would only consider double precision for finite difference comparisons).

src/pushforwards/svd.jl
kshyatt: Since we went to all the effort of adding the projection functions I guess we should use them, extra alloc or not

src/pushforwards/svd.jl
```diff
@@ -0,0 +1,77 @@
...
+    V₁ = adjoint(V₁ᴴ)
+    ΔAV₁ = ΔA * V₁
+    UᴴΔAV₁ = U₁' * ΔAV₁
+    if !iszerotangent(ΔS)
+        ΔS₁ = view(diagview(ΔS), 1:r)
+        ΔS₁ .= real.(diagview(UᴴΔAV₁))
+    end
+    if !iszerotangent(ΔU) || !iszerotangent(ΔVᴴ)
+        hUᴴΔAV₁ = inv_safe.(transpose(S₁) .- S₁) .* project_hermitian(UᴴΔAV₁)
+        aUᴴΔAV₁ = inv_safe.(transpose(S₁) .+ S₁) .* project_antihermitian(UᴴΔAV₁)
```
lkdvos: I think below only the sum and difference are actually used, we could use a kernel like https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/e271b4aa2151ef3044407b46016c56ffe0606bb2/src/implementations/polar.jl#L209-L220 to avoid the two extra allocations, but I'm also happy to just leave them as-is, it's hard to imagine this really making that huge of a difference

src/pushforwards/svd.jl
```diff
@@ -0,0 +1,77 @@
...
+            infinitesimal_phases = imag.(ΔU₁[I] ./ U₁[I])
+            ΔU₁ .-= im .* U₁ .* infinitesimal_phases
+            ΔV₁ᴴ .+= im .* transpose(infinitesimal_phases) .* V₁ᴴ
+        end
+    end
+    return (ΔU, ΔS, ΔVᴴ)
+end
+
+function svd_trunc_pushforward!(ΔA, A, USVᴴ, ΔUSVᴴ, ind; rank_atol = default_pullback_rank_atol(A), kwargs...)
+    # TODO
```
lkdvos: Probably best to either not define this or explicitly `error()`?

lkdvos (commented):
Left some final small comments, otherwise good to go?

kshyatt:
The GPU tests will probably fail until a new version is tagged at GPUArrays (https://github.com/JuliaGPU/GPUArrays.jl/pull/738)

test/testsuite/enzyme/svd.jl
```diff
@@ -8,48 +8,80 @@ function test_enzyme_svd(T::Type, sz; kwargs...)
...
     )
-    return @testset "svd_compact reverse: RT $RT, TA $TA" for RT in (Duplicated,), TA in (Duplicated,)
+    return @testset "svd_compact: RT $RT, TA $TA" for RT in (Duplicated,), TA in (Duplicated,)
         A = instantiate_matrix(T, sz)
         alg = MatrixAlgebraKit.select_algorithm(svd_compact, A)
         USVᴴ, ΔUSVᴴ = ad_svd_compact_setup(A)
         test_reverse(svd_compact, RT, (A, TA), (alg, Const); atol, rtol, output_tangent = ΔUSVᴴ, fdm)
-        test_reverse(call_and_zero!, RT, (svd_compact!, Const), (A, TA), (alg, Const); atol, rtol, output_tangent = ΔUSVᴴ, fdm)
+        test_reverse(call_and_zero!, RT, (svd_compact!, Const), (copy(A), TA), (alg, Const); atol, rtol, output_tangent = ΔUSVᴴ, fdm)
+        if eltype(T) <: Real
```
Jutho: I think you should now be able to remove this constraint, and even extend to the `m != n` case. Might be worth trying (first locally maybe, I haven't tried these tests myself, just compared against my own very simple finite difference).

test/testsuite/enzyme/svd.jl
```diff
@@ -8,48 +8,80 @@ function test_enzyme_svd(T::Type, sz; kwargs...)
...
     )
-    return @testset "svd_full reverse: RT $RT, TA $TA" for RT in (Duplicated,), TA in (Duplicated,)
+    return @testset "svd_full: RT $RT, TA $TA" for RT in (Duplicated,), TA in (Duplicated,)
         A = instantiate_matrix(T, sz)
         alg = MatrixAlgebraKit.select_algorithm(svd_full, A)
         USVᴴ, ΔUSVᴴ = ad_svd_full_setup(A)
         test_reverse(svd_full, RT, (A, TA), (alg, Const); atol, rtol, output_tangent = ΔUSVᴴ, fdm)
-        test_reverse(call_and_zero!, RT, (svd_full!, Const), (A, TA), (alg, Const); atol, rtol, output_tangent = ΔUSVᴴ, fdm)
+        test_reverse(call_and_zero!, RT, (svd_full!, Const), (copy(A), TA), (alg, Const); atol, rtol, output_tangent = ΔUSVᴴ, fdm)
+        if eltype(T) <: Real && size(A, 1) == size(A, 2) # finite differences check for free component is very finicky
```
Jutho: Here it should also work in the complex case, but you should keep the `size(A,1) == size(A,2)` check.

test/testsuite/mooncake/svd.jl
```diff
@@ -33,16 +33,26 @@ function test_mooncake_svd_compact(
             mode = Mooncake.ReverseMode, output_tangent, atol, rtol
         )
         Mooncake.TestUtils.test_rule(
-            rng, call_and_zero!, svd_compact!, A, alg;
+            rng, call_and_zero!, svd_compact!, copy(A), alg;
             mode = Mooncake.ReverseMode, output_tangent, atol, rtol, is_primitive = false
         )
+        if eltype(T) <: Real # gauge freedom in complex outputs
```
Jutho: Same comments for Mooncake.

test/testsuite/mooncake/svd.jl
```diff
@@ -60,16 +70,26 @@ function test_mooncake_svd_full(
             mode = Mooncake.ReverseMode, output_tangent, atol, rtol
         )
         Mooncake.TestUtils.test_rule(
-            rng, call_and_zero!, svd_full!, A, alg;
+            rng, call_and_zero!, svd_full!, copy(A), alg;
             mode = Mooncake.ReverseMode, output_tangent, atol, rtol, is_primitive = false
         )
+        if eltype(T) <: Real # gauge freedom in complex outputs
```
Jutho: Replace with equal size check.

kshyatt:
Ran it locally and it worked, let's hope!

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -562,7 +563,18 @@ for (f!, f) in (
...
-        @is_primitive Mooncake.DefaultCtx Mooncake.ReverseMode Tuple{typeof($f), Any, MatrixAlgebraKit.AbstractAlgorithm}
+        function Mooncake.frule!!(::Dual{typeof($f!)}, A_dA::Dual, USVᴴ_dUSVᴴ::Dual, alg_dalg::Dual)
+            A, dA = arrayify(A_dA)
+            USVᴴ = Mooncake.primal(USVᴴ_dUSVᴴ)
+            dUSVᴴ = Mooncake.tangent(USVᴴ_dUSVᴴ)
+            U, dU = arrayify(USVᴴ[1], dUSVᴴ[1])
+            S, dS = arrayify(USVᴴ[2], dUSVᴴ[2])
+            Vᴴ, dVᴴ = arrayify(USVᴴ[3], dUSVᴴ[3])
+            $f!(A, USVᴴ, Mooncake.primal(alg_dalg))
+            svd_pushforward!(dA, A, (U, S, Vᴴ), (dU, dS, dVᴴ))
```
Jutho: This again works because `svd_pushforward!` doesn't actually need `A`, since `A` is destroyed at this point. Not sure if it is worth to add a comment.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -604,8 +629,17 @@ function Mooncake.rrule!!(::CoDual{typeof(svd_vals!)}, A_dA::CoDual, S_dS::CoDua
     end
     return S_dS, svd_vals_adjoint
 end
+function Mooncake.frule!!(::Dual{typeof(svd_vals!)}, A_dA::Dual, S_dS::Dual, alg_dalg::Dual)
+    # compute primal
+    A, dA = arrayify(A_dA)
+    S, dS = arrayify(S_dS)
+    USVᴴ = svd_compact(A, Mooncake.primal(alg_dalg))
```
Jutho: Any reason not to use `svd_compact!` here?

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -213,4 +213,14 @@ function eig_pullback!(ΔA::AnyCuMatrix, A, DV, ΔDV, ind::AnyCuVector; kwargs..
     return eig_pullback!(ΔA, A, DV, ΔDV, collect(ind); kwargs...)
 end
 
+# have to override this as methods are missing in GPUArrays for the various
+# views of Diagonal of ΔA
```
Jutho, line 217: This is a bit of a confusing comment: what exactly is missing?
Might be useful to keep track of this, in case it gets fixed.

kshyatt: It's again a situation of `mul!(::Diagonal{T, CuVector{T}}, [horrific view of adjoint of view], CuArray)` which GPUArrays cannot dispatch onto at all.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
Jutho: Would you mind, in this whole block, to replace `f` with `f!`. That makes it a bit clearer that this is about the mutating method, and is in line with Mooncake, where we define rules for both `f` and `f!`.

ext/MatrixAlgebraKitEnzymeExt/MatrixAlgebraKitEnzymeExt.jl
```diff
@@ -264,6 +265,34 @@ for f in (:svd_compact!, :svd_full!)
...
+                alg::Const{<:MatrixAlgebraKit.AbstractAlgorithm},
+            ) where {RT, TA}
+            $f(A.val, USVᴴ.val, alg.val)
+            if !isa(A, Const)
+                if $(f == svd_compact!)
+                    make_zero!(USVᴴ.dval[2].diag)
+                else
+                    make_zero!(USVᴴ.dval[2])
+                end
+                !isa(USVᴴ, Const) && svd_pushforward!(A.dval, A.val, USVᴴ.val, USVᴴ.dval)
```
Jutho: Why is it necessary to make `USVᴴ.dval[2]`, so `dS` zero specifically? And then why not `dU` and `dV`. Is that something we can fix directly in `svd_pushforward!`?

Also, does it make sense to still have a `!isa(USVᴴ, Const)` test afterwards? Are you not already assuming that it is not constant if you try to access `USVᴴ.dval` before? 

Finally, what happens in the case where `A.val::Diagonal` and `S = USVᴴ.val[2]` might potentially shadow `A.val`. Will then `USVᴴ.dval[2]` also be identical to `A.dval`? In that case, this `make_zero!` would really be problematic, as we erase `A.dval`, which we still need.

kshyatt: I think this is an artifact from the previous version of the push forward, let's remove it. That version also didn't test with Diagonal yet.

kshyatt: Also the next `make_zero!(A.dval)` after the pf runs is also problematic since it doesn't check for the `A === S` case, I'll fix that too.

kshyatt: It's tough to reason about all this because of the inconsistent "when is `A` also an `arg`?" stuff we were talking about on Zulip 😢

kshyatt:
Of course when I ran this locally (5 times!!!) it passed every single time.

kshyatt:
Some of these look like tolerance issues to me at least

src/pushforwards/svd.jl
```diff
@@ -55,7 +55,7 @@ function svd_pushforward!(ΔA, A, USVᴴ, ΔUSVᴴ, ind = Colon(); rank_atol = d
         end
         if eltype(U) <: Complex && !iszerotangent(ΔU) && !iszerotangent(ΔVᴴ) # fix gauge for `gaugefix!` compatibility
             _, I = findmax(abs, U₁; dims = 1)
-            infinitesimal_phases = imag.(ΔU₁[I] ./ U₁[I])
+            infinitesimal_phases = imag.(ΔU₁[I] .* inv_safe.(U₁[I]))
```
Jutho, line 63: I would be suprised if that is needed or makes a difference. `U₁[I]` is the maximum element of every column of `U₁`. As `U₁` is an isometric matrix, all of its columns have norm 1, and therefore, the largest element needs to be at least `1/sqrt(m)`, in magnitude, with `m` the number of rows. Typically, it will be larger.

Jutho:
I am also happy to take a look at the test failures.

src/pushforwards/svd.jl
kshyatt: I would be quite surprised too but I'm otherwise very confused where the `NaN` are emerging from (the `./transpose(S1)` line has the same objection). I'll try stepping through the pushfoward to see if I can find the culprit.

kshyatt:
> I am also happy to take a look at the test failures.

If you like! I'm about to get on a plane and unsure if it will have WiFi with which to push if I can find the problem

Jutho:
Ok, I located the problem, but also started changing various other things so this turned out a bit bigger than anticipated. The problem was that the initialization of the tangent in the `frule` is random introducing NaNs, subnormals and other random stuff. But `svd_pushforward` only touches the diagonal of `S`, and did not set the off-diagonal entries (in the case of `svd_full`) to zero.

Jutho:
I also tried to make some steps in the various implementations more consistent, and to simplify them a bit here and there. I introduced a `has_equal_storage` for simplifying and generalizing the detection of the input matrix `A` overlapping with some of the output arguments.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -624,6 +639,16 @@ function Mooncake.rrule!!(::CoDual{typeof(svd_vals)}, A_dA::CoDual, alg_dalg::Co
     end
     return S_codual, svd_vals_adjoint
 end
+function Mooncake.frule!!(::Dual{typeof(svd_vals)}, A_dA::Dual, alg_dalg::Dual)
+    # compute primal
+    A, dA = arrayify(A_dA)
+    USVᴴ = svd_compact(A, Mooncake.primal(alg_dalg))
+    S = diagview(USVᴴ[2])
+    S_dual = Dual(S, Mooncake.zero_tangent(S))
```
lkdvos: Is there a `Mooncake.zero_fdual` similar to the `zero_fcodual`? There might be a couple more places where this could be used to simplify the code

src/common/utility.jl
```diff
@@ -0,0 +1,16 @@
+
+function has_equal_storage(A::Diagonal, B::Diagonal)
+    return diagview(A) === diagview(B)
```
lkdvos, line 2: Does this overlap with `Base.mightalias`?

src/pushforwards/eig.jl
```diff
@@ -12,11 +11,16 @@ function eig_pushforward!(
     end
     if !iszerotangent(ΔV)
         ∂K .*= inv_safe.(transpose(diagview(D)) .- diagview(D), degeneracy_atol)
-        mul!(ΔV, V, ∂K, 1, 0)
+        mul!(ΔV, V, ∂K)
+        if eltype(V) <: Complex # fix gauge for `gaugefix!` compatibility
```
lkdvos, line 15: Are we enforcing that the primal computation is always gauge fixing, and is that relevant for this?

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
kshyatt: https://github.com/chalk-lab/Mooncake.jl/blob/ab5d8023f95aca5ed20100ec72960f22ad33cddf/src/tangents/dual.jl#L28 It's `zero_dual`

src/common/utility.jl
Jutho: This is more strict. The way we currently deal with the cases where `has_equal_storage(input_matrix, some_output_arg)` (mostly in the case of `Diagonal`, but also in some other), would not work if the storage has some general aliasing. (So we don't support general aliasing, either they are distinct or they are equal in the above sense).

src/pushforwards/eig.jl
Jutho: No we are not enforcing this, but it is the default. This is relevant to the extent that if `gaugefix = false`, there is no way to compute the pushforward at all. So we could throw a warning. There is some asymmetry between the pullbacks and the pushforwards in that the latter don't throw warnings for cases where there is gauge freedom.

But the way to deal with this is also very different between the two.

ext/MatrixAlgebraKitMooncakeExt/MatrixAlgebraKitMooncakeExt.jl
```diff
@@ -181,48 +180,55 @@ for (f!, f, pb, adj) in (
...
-                    dA .= A
-                end
-                if A !== D
                     zero!(dD)
                     copy!(D, Dc)
-                else
+                else # A and D have the same storage
+                    ΔA = zero(A)
+                    $pb(ΔA, Ac, DV, dD)
+                    dA .= ΔA
```
lkdvos:
```suggestion
                    copy!(dA, ΔA)
```
This is slightly inconsistent with the line below, not sure if it really matters but might be good to remain consistent?

lkdvos (approved):
Minor detail, but otherwise this looks ready to me?

lkdvos (approved):

src/common/utility.jl
```diff
@@ -13,4 +12,4 @@ function has_equal_storage(A::AbstractVector, B::Diagonal)
     return A === diagview(B)
 end
 has_equal_storage(A::AbstractMatrix, B::AbstractVector) = false
-has_equal_storage(A::AbstractVector, B::AbstractMatrix) = false
\ No newline at end of file
+has_equal_storage(A::AbstractVector, B::AbstractMatrix) = false
```
Jutho, line 15: So this was my formatting issue? I don't actually see any change? Is it some invisible character?

kshyatt: I think it's that we were missing a `\n` after the `false`

PR #251, opened 2026-06-30
Add docs for matrix functions and `exponential`

sanderdemeyer:
Add documentation for general matrix functions, specifically for `exponential`.

### DIFF
--- a/docs/src/user_interface/algorithms.md
+++ b/docs/src/user_interface/algorithms.md
@@ -80,8 +80,10 @@ DefaultAlgorithm
 
 ## Available Algorithm Types
 
-The following high-level algorithm types are available.
+There are several high-level algorithm types available.
 They all accept an optional `driver` keyword to select the computational backend; see [Driver Selection](@ref sec_driverselection) for details.
+
+The following algorithms for matrix decompositions are available.
 
 | Algorithm | Applicable decompositions | Key keyword arguments |
 |:----------|:--------------------------|:----------------------|
@@ -95,6 +97,11 @@ They all accept an optional `driver` keyword to select the computational backend
 | [`SVDViaPolar`](@ref) | SVD | `fixgauge`, `tol` |
 | [`PolarViaSVD`](@ref) | polar | positional `svd_alg` argument |
 | [`PolarNewton`](@ref) | polar | `maxiter`, `tol` |
+
+The following algorithms for matrix functions are available.
+
+| Algorithm | Applicable matrix functions | Key keyword arguments |
+|:----------|:--------------------------|:----------------------|
 | [`MatrixFunctionViaLA`](@ref) | exponential | |
 | [`MatrixFunctionViaEig`](@ref) | exponential | `eig_alg` |
 | [`MatrixFunctionViaEigh`](@ref) | exponential | `eigh_alg` |
--- a/docs/src/user_interface/algorithms.md
+++ b/docs/src/user_interface/algorithms.md
@@ -95,6 +95,9 @@ They all accept an optional `driver` keyword to select the computational backend
 | [`SVDViaPolar`](@ref) | SVD | `fixgauge`, `tol` |
 | [`PolarViaSVD`](@ref) | polar | positional `svd_alg` argument |
 | [`PolarNewton`](@ref) | polar | `maxiter`, `tol` |
+| [`MatrixFunctionViaLA`](@ref) | exponential | |
+| [`MatrixFunctionViaEig`](@ref) | exponential | `eig_alg` |
+| [`MatrixFunctionViaEigh`](@ref) | exponential | `eigh_alg` |
 
 For full docstring details on each algorithm type, see the corresponding section in [Decompositions](@ref).
 
--- a/docs/src/user_interface/matrix_functions.md
+++ b/docs/src/user_interface/matrix_functions.md
@@ -22,7 +22,7 @@ Additionally, the `f!` method typically assumes that it is allowed to destroy th
 
 ## Exponential
 
-The [exponential](https://en.wikipedia.org/wiki/Matrix_exponential) of a rectangular matrix `A` is used in many quantum many-body problems.
+The [exponential](https://en.wikipedia.org/wiki/Matrix_exponential) of a square matrix `A` is used in many scientific applications, as it arises in the solution of an autonomous linear differential equation.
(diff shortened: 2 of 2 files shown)


docs/src/user_interface/matrix_functions.md
```diff
@@ -1,3 +1,35 @@
...
+For a full description of how to select and configure algorithms, see [Algorithm Selection](@ref sec_algorithmselection).
+Importantly, for generic code patterns it is recommended to always use the output `F` explicitly, since some implementations may not be able to reuse the provided memory.
+Additionally, the `f!` method typically assumes that it is allowed to destroy the input `A`, and making use of the contents of `A` afterwards should be deemed as undefined behavior.
+
+## Exponential
+
+The [exponential](https://en.wikipedia.org/wiki/Matrix_exponential) of a rectangular matrix `A` is used in many quantum many-body problems. This is implemented in `LinearAlgebra`, which can be accessed by the algorithm [`MatrixFunctionViaLA`](@ref). For more generic data types, the exponential can be calculated by first calculating the (hermitian) eigenvalue decomposition. This is done by `eig_full` and `eigh_full` via the algorithms [`MatrixFunctionViaEig`](@ref) and [`MatrixFunctionViaEigh`](@ref), respectively. 
+Additionally, in order to calculate `exp(τ*A)`, the function `exponential` can be called with (τ,A), using the same algorithms as before.
```
lkdvos:
```suggestion
The [exponential](https://en.wikipedia.org/wiki/Matrix_exponential) of a rectangular matrix `A` is used in many quantum many-body problems.
This is implemented in `LinearAlgebra`, which can be accessed by the algorithm [`MatrixFunctionViaLA`](@ref).
For more generic data types, the exponential can be calculated by first calculating the (hermitian) eigenvalue decomposition.
This is done by `eig_full` and `eigh_full` via the algorithms [`MatrixFunctionViaEig`](@ref) and [`MatrixFunctionViaEigh`](@ref), respectively.
 Additionally, in order to calculate `exp(τ * A)`, the function `exponential` can be called with `(τ, A)`, using the same algorithms as before.
```

docs/src/user_interface/matrix_functions.md
```diff
@@ -1,3 +1,35 @@
...
+
+```@docs; canonical=false
+exponential
+```
+
+```@autodocs; canonical=false
+Modules = [MatrixAlgebraKit]
+Filter = t -> t isa Type && (t <: MatrixAlgebraKit.MatrixFunctionViaLA || t <: MatrixAlgebraKit.MatrixFunctionViaEig || t <: MatrixAlgebraKit.MatrixFunctionViaEigh)
```
lkdvos: I think here we either want to add a union type for matrixfunction algorithms, or just replace this with a `@docs` block, it is probably overkill to search through all algorithms and then only filter for 3 specific ones, since in that case we can just manually list them :)

docs/src/user_interface/algorithms.md
```diff
@@ -95,6 +95,9 @@ They all accept an optional `driver` keyword to select the computational backend
 | [`SVDViaPolar`](@ref) | SVD | `fixgauge`, `tol` |
 | [`PolarViaSVD`](@ref) | polar | positional `svd_alg` argument |
 | [`PolarNewton`](@ref) | polar | `maxiter`, `tol` |
+| [`MatrixFunctionViaLA`](@ref) | exponential | |
```
lkdvos, line 105: Do we want to specifically separate out the matrix functions from the decompositions? It might be more clear to keep these separate

lkdvos (commented):
Thanks for getting this started! Left some small comments/suggestions, but overall this looks good!

docs/src/user_interface/algorithms.md
sanderdemeyer: That's a good idea. A suggestion on how to do this is in the latest commit.

docs/src/user_interface/matrix_functions.md
```diff
@@ -1,3 +1,36 @@
...
+
+Here, the input matrix is always the first argument, and optionally the output can be provided as well.
+The keywords are algorithm-specific, and can be used to influence the behavior of the algorithms.
+For a full description of how to select and configure algorithms, see [Algorithm Selection](@ref sec_algorithmselection).
+Importantly, for generic code patterns it is recommended to always use the output `F` explicitly, since some implementations may not be able to reuse the provided memory.
+Additionally, the `f!` method typically assumes that it is allowed to destroy the input `A`, and making use of the contents of `A` afterwards should be deemed as undefined behavior.
+
+## Exponential
+
+The [exponential](https://en.wikipedia.org/wiki/Matrix_exponential) of a rectangular matrix `A` is used in many quantum many-body problems.
```
Jutho:
```suggestion
The [exponential](https://en.wikipedia.org/wiki/Matrix_exponential) of a square matrix `A` is used in many scientific applications, as it arises in the solution of an autonomous linear differential equation.
```

docs/src/user_interface/matrix_functions.md
```diff
@@ -1,3 +1,36 @@
...
+This is implemented in `LinearAlgebra`, which can be accessed by the algorithm [`MatrixFunctionViaLA`](@ref).
+For more generic data types, the exponential can be calculated by first calculating the (hermitian) eigenvalue decomposition.
+This is done by `eig_full` and `eigh_full` via the algorithms [`MatrixFunctionViaEig`](@ref) and [`MatrixFunctionViaEigh`](@ref), respectively.
+ Additionally, in order to calculate `exp(τ * A)`, the function `exponential` can be called with `(τ, A)`, using the same algorithms as before.
```
Jutho:
```suggestion
An implementation for the matrix exponential based on a Padé approximation is available in `LinearAlgebra`, and can be accessed by the algorithm [`MatrixFunctionViaLA`](@ref).
For more generic data types, the exponential can be calculated by first calculating the (hermitian) eigenvalue decomposition, and then computing
the scalar exponential of the diagonal elements.
This strategy is implemented via the algorithms [`MatrixFunctionViaEig`](@ref) and [`MatrixFunctionViaEigh`](@ref), and call `eig_full` and `eigh_full`, respectively.
Additionally, in order to calculate `exp(τ * A)`, the function `exponential` can be called with `(τ, A)`, using the same algorithms as before.
```

lkdvos (approved):

PR #253, opened 2026-07-03
fix: `exponential` method ambiguity resolutions

sanderdemeyer:
This is another type restriction (see #252) to make sure the exponential in TensorKit will be implemented correctly (see QuantumKitHub/TensorKit.jl#465). If I don't restrict these types, it will be ambiguous with this function in TensorKit (src/factorizations/matrixalgebrakit.jl).

```julia
for f! in (
        :qr_null!, :lq_null!,
        :svd_vals!, :eig_vals!, :eigh_vals!,
        :project_hermitian!, :project_antihermitian!, :project_isometric!,
        :exponential!,
    )
    @eval function MAK.$f!(t::AbstractTensorMap, N, alg::AbstractAlgorithm)
        $(f! in (:eig_vals!, :eigh_vals!, :project_hermitian!, :project_antihermitian!) && :(LinearAlgebra.checksquare(t)))
        foreachblock(t, N) do _, (tblock, Nblock)
            Nblock′ = $f!(tblock, Nblock, alg)
            # deal with the case where the output is not the same as the input
            Nblock === Nblock′ || copy!(Nblock, Nblock′)
            return nothing
        end
        return N
    end
end
```

### DIFF
--- a/src/implementations/exponential.jl
+++ b/src/implementations/exponential.jl
@@ -44,8 +44,8 @@ end
 # ---------------------------
 exponential!(A::AbstractMatrix, alg::DefaultAlgorithm) = exponential!(A, select_algorithm(exponential!, A, nothing; alg.kwargs...))
 exponential!(A::AbstractMatrix, out, alg::DefaultAlgorithm) = exponential!(A, out, select_algorithm(exponential!, A, nothing; alg.kwargs...))
-exponential!((τ, A)::Tuple{Number, AbstractMatrix}, alg::DefaultAlgorithm) = exponential!(A, select_algorithm(exponential!, A, nothing; alg.kwargs...))
-exponential!((τ, A)::Tuple{Number, AbstractMatrix}, out, alg::DefaultAlgorithm) = exponential!(A, out, select_algorithm(exponential!, A, nothing; alg.kwargs...))
+exponential!(τA::Tuple{Number, AbstractMatrix}, alg::DefaultAlgorithm) = exponential!(τA, select_algorithm(exponential!, τA, nothing; alg.kwargs...))
+exponential!(τA::Tuple{Number, AbstractMatrix}, out, alg::DefaultAlgorithm) = exponential!(τA, out, select_algorithm(exponential!, τA, nothing; alg.kwargs...))
 
 # Outputs
 # -------
--- a/src/implementations/exponential.jl
+++ b/src/implementations/exponential.jl
@@ -8,14 +8,14 @@ copy_input(::typeof(exponential), A::Diagonal) = copy(A)
 copy_input(::typeof(exponential), (τ, A)::Tuple{Number, AbstractMatrix}) = (τ, copy!(similar(A, float(eltype(A))), A))
 copy_input(::typeof(exponential), (τ, A)::Tuple{Number, Diagonal}) = τ, copy(A)
 
-function check_input(::typeof(exponential!), A::AbstractMatrix, expA::AbstractMatrix, alg::AbstractAlgorithm)
+function check_input(::typeof(exponential!), A::AbstractMatrix, expA, alg::AbstractAlgorithm)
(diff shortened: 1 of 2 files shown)


lkdvos (approved):
Did a quick search throughout the codebase, and it seems like except for `svd_compact`, we always combine `DiagonalAlgorithm` with `AbstractMatrix`, so this is consistent with most of it.

src/implementations/exponential.jl
```diff
@@ -40,6 +40,12 @@ function check_input(::typeof(exponential!), (τ, A)::Tuple{Number, AbstractMatr
     return nothing
 end
 
+# DefaultAlgorithm intercepts
+# ---------------------------
+
+exponential!(t::Tuple{E, T}, ::DefaultAlgorithm) where {E <: Number, T <: Diagonal} = MAK.exponential!(t, DiagonalAlgorithm())
+exponential!(t::Tuple{E, T1}, out::T2, ::DefaultAlgorithm) where {E <: Number, T1 <: Diagonal, T2 <: Diagonal} = MAK.exponential!(t, out, DiagonalAlgorithm())
```
lkdvos: I think the idea here is rather to make this go back to the default algorithm selection procedure, see all factorizations for examples: e.g. https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/11096c1667ca0c08cb15eaa7e8388a55a0abe24a/src/implementations/lq.jl#L91-L98

(Also, this should not only be for `Diagonal`, but for `AbstractMatrix` instead.

src/implementations/exponential.jl
```diff
@@ -40,6 +40,13 @@ function check_input(::typeof(exponential!), (τ, A)::Tuple{Number, AbstractMatr
     return nothing
 end
 
+# Algorithm selection
+# ---------------------------
+exponential!(A::AbstractMatrix, alg::DefaultAlgorithm) = exponential!(A, select_algorithm(exponential!, A, nothing; alg.kwargs...))
+exponential!(A::AbstractMatrix, out, alg::DefaultAlgorithm) = exponential!(A, out, select_algorithm(exponential!, A, nothing; alg.kwargs...))
+exponential!((τ, A)::Tuple{Number, AbstractMatrix}, alg::DefaultAlgorithm) = exponential!(A, select_algorithm(exponential!, A, nothing; alg.kwargs...))
+exponential!((τ, A)::Tuple{Number, AbstractMatrix}, out, alg::DefaultAlgorithm) = exponential!(A, out, select_algorithm(exponential!, A, nothing; alg.kwargs...))
```
lkdvos:
```suggestion
exponential!(τA::Tuple{Number, AbstractMatrix}, alg::DefaultAlgorithm) = exponential!(τA, select_algorithm(exponential!, τA, nothing; alg.kwargs...))
exponential!(τA::Tuple{Number, AbstractMatrix}, out, alg::DefaultAlgorithm) = exponential!(τA, out, select_algorithm(exponential!, τA, nothing; alg.kwargs...))
```

Jutho:
There are still ambiguities currently. Also, I do think the current these specializations invalidate some of the discussions we had previously, e.g. regarding `map_diagonal` vs broadcasting.

sanderdemeyer:
These ambiguities should now be removed. It is indeed a bit strange that we're specialising these methods, when we indeed wanted to keep them general and not use broadcasting. I think it was probably still the right choice, even if we restrict some of the methods here.

lkdvos (approved):
I made a slight change to the implementation to remain consistent with the decompositions - we are now never specializing the `out` arguments, and only using `input::AbstractMatrix`. I'm not saying this is really the best solution, and I agree that it is unfortunate we can't (yet?) handle this in a more satisfactory way, but barring the option to actually fully solve this right now, which I think will take a substantial refactor, I would vote to just keep this consistent with the decompositions, where this problem is also present.

Jutho (approved):

PR #255, opened 2026-07-10
fix: Taylor exponentiate shouldn't assume `typeof(A*A) === typeof(A)`

lkdvos:
Bumped into this while trying it out in TensorKit, where `reshape(view(...))` errors.


lkdvos:
CUDA failures look like unrelated truncated eig failures.

Jutho (approved):

PR #256, opened 2026-07-14
Don't use nothing in caches

kshyatt:
This passes a reference to the argument in the cache, rather than `nothing`, which should not affect the results but does help Enzyme with type analysis which will allow us to use these pullbacks directly in TK.


lkdvos (approved):
definitely makes sense to me!

Jutho (approved):
Looks good to me. Are these the only places where we use this pattern? it feels like there are many other places (but I haven't actually looked at any code).

PR #257, opened 2026-07-14
Try using Enzyme with BigFloat

kshyatt:


Jutho:
Does not seem to work unfortunately. The forward rules seem to work though. Not sure what goes wrong with the reverse rules; I don't quite understand where the warning is coming from.

kshyatt:
Very weird indeed. I know this is a rather marginal use case but would still be nice to support.

PR #261, opened 2026-07-27
feat: `squareroot`, `logarithm` and `power`

lkdvos:
Rebased onto `main`, which now contains #262 (the Pkg-workspace test migration). This PR adds the `test/matrixfunctions` group that #262's description deferred here, and updates `test/README.md` accordingly.

Following up on the `exponential` implementation, with the goal of getting https://github.com/QuantumKitHub/TensorKit.jl/pull/493 to work.

I do have the intention of adding better algorithms for the `squareroot` and `logarithm` implementations, but I figured it might be nice to avoid doing that in one huge PR and to keep things moving by defining the interface first.

Still to do:
- [x] Refactor tests into testsuite
- [ ] Decide if we want to restructure the folders to separate decompositions and matrix functions
- [x] See if I can refactor some of these implementations to share a bit more code
- [x] Add tests for GPU support

## Notes for review

A few things worth knowing, mostly about behaviour that was discovered rather than designed:

- **`power(A, 0)` and `power(A, 1)` short-circuit** to `I` and `A` before any decomposition, so they
  now apply to any square matrix whatever the algorithm — including input the algorithm would
  otherwise reject, e.g. a non-hermitian matrix with `MatrixFunctionViaEigh`. `A^0 = I` and
  `A^1 = A` hold regardless of spectrum, and `Base.^` makes no hermiticity demand either.
- **An integral `Float` exponent is still recognised at runtime**, via `isinteger(p)` rather than
  dispatch on `typeof(p)`. This matches `Base`, whose `^(::AbstractMatrix, ::Real)` does
  `isinteger(p) && return integerpow(A, p)`; dispatching on the type instead would make
  `power(A, 3.0)` throw a `DomainError` for a matrix with a negative eigenvalue while `A^3.0`
  succeeds.
- **`exponential` of a complex hermitian matrix is now exactly hermitian**, which it was not before:
  the real branch built a symmetric product but the complex branch used a plain `V exp(D) V'` with
  no closing projection. A complex `τ` is still left unprojected, because `exp(τA)` is hermitian
  only for real `τ`.
- **The domain rules genuinely differ per function**, and the tests encode that rather than
  smoothing it over: a roundoff-negative eigenvalue clamps for `squareroot` and positive fractional
  `power`, but still throws for `logarithm`, where clamping onto zero leaves the matrix singular.
  `MatrixFunctionViaLA` inspects only the realness of its result, so it accepts singular input that
  the spectrum-based algorithms reject.

Two pre-existing issues on `main` were found while testing and are deliberately *not* addressed
here:

- `MatrixFunctionViaEig` is not GPU-clean — its `lu!`-based solve scalar-indexes — and
  `MatrixFunctionViaLA` calls LAPACK on device memory. Device support for `squareroot`, `logarithm`
  and `power` is therefore hermitian-via-`eigh` plus the `Diagonal` fast path; `exponential` escapes
  this through the native `MatrixFunctionViaTaylor`.
- `power(A, -1)` on a numerically singular matrix returns garbage rather than throwing for the
  `eig`-based algorithms, because `any(iszero, λ)` is an exact test against *computed* eigenvalues.
  The `Diagonal` path throws `SingularException` correctly.


docs/src/user_interface/matrix_functions.md
```diff
@@ -37,3 +37,42 @@ MatrixAlgebraKit.MatrixFunctionViaLA
 MatrixAlgebraKit.MatrixFunctionViaEig
 MatrixAlgebraKit.MatrixFunctionViaEigh
 ```
+
+## Domain considerations
+
+The functions below ([`squareroot`](@ref), [`logarithm`](@ref) and [`power`](@ref) with fractional powers) are only defined for matrices whose eigenvalues avoid (part of) the negative real axis, and their principal values are complex whenever eigenvalues on that axis are present.
+In MatrixAlgebraKit, we aim to keep type stability, and thus the scalar type of the output always matches that of the input.
+As such, a real matrix with eigenvalues on the negative real axis leads to a `DomainError`, you should pass a complex matrix instead to obtain the complex principal value.
```
kshyatt: There should be a period after `DomainError`. I think this is [comma splice](https://en.wikipedia.org/wiki/Comma_splice)

src/implementations/logarithm.jl
```diff
@@ -0,0 +1,84 @@
...
+    m = LinearAlgebra.checksquare(A)
+    @check_size(logA, (m, m))
+    @check_scalar(logA, A)
+    return nothing
+end
+
+function check_input(::typeof(logarithm!), A::AbstractMatrix, logA, ::DiagonalAlgorithm)
+    m = LinearAlgebra.checksquare(A)
+    @assert isdiag(A)
+    @assert logA isa Diagonal
```
kshyatt: if `logA` has to be a `Diagonal`, why not enforce that in the function arguments? Or should this be `isdiagonal`?

src/implementations/polar.jl
```diff
@@ -81,19 +81,6 @@ function right_polar!(A::AbstractMatrix, PWᴴ, alg::PolarViaSVD)
...
-function _mul_herm!(C, A)
-    mul!(C, A, A')
-    project_hermitian!(C)
-    return C
-end
-function _mul_herm!(C::YALAPACK.BlasMat{T}, A::YALAPACK.BlasMat{T}) where {T <: YALAPACK.BlasFloat}
-    mul!(C, A, A')
-    return C
-end
-
```
kshyatt, line 96: Is this no longer needed? What happened here?

kshyatt: Oh I see it moved to another file. OK!

src/implementations/power.jl
```diff
@@ -0,0 +1,99 @@
...
+    end
+end
+
+# Diagonal logic
+# --------------
+function power!(A::AbstractMatrix, p::Real, powA, alg::DiagonalAlgorithm)
+    check_input(power!, A, p, powA, alg)
+    λ = diagview(powA)
+    copyto!(λ, diagview(A))
+    if isinteger(p)
```
kshyatt: Should we add a short-circuit here for when `p === 0` and when `p === 1`?

src/implementations/power.jl
```diff
@@ -0,0 +1,99 @@
...
+        mul!(powA, VD, V')
+        return project_hermitian!(powA)
+    else
+        # `A^p = (V * D^(p/2)) * (V * D^(p/2))'` is hermitian by construction
+        Vs = rmul!(V, power!(D, p / 2, D, diag_alg))
+        return _mul_herm!(powA, Vs)
+    end
+end
+
+function power!(A::AbstractMatrix, p::Real, powA, alg::MatrixFunctionViaEig)
```
kshyatt, line 50: similar question to below about `p === 0` and `p === 1`

src/implementations/logarithm.jl
lkdvos: This is consistent with the rest of the package's `DiagonalAlgorithm` implementations, although I'm not sure I still think this is the best approach.
There are some arguments in favor of having a specific error instead of a `MethodError` (although in that case I should probably add proper error messages), and I think it is convenient for method ambiguity reasons to always have the same pattern, i.e. not specializing on the `out` type.
Ultimately, I don't think there is any great reason for this, but since it is consistent with the rest of the package I'd vote to keep it in line for now?

docs/src/user_interface/matrix_functions.md
```diff
@@ -37,3 +37,43 @@ MatrixAlgebraKit.MatrixFunctionViaLA
 MatrixAlgebraKit.MatrixFunctionViaEig
 MatrixAlgebraKit.MatrixFunctionViaEigh
 ```
+
+## Domain considerations
+
+The functions below ([`squareroot`](@ref), [`logarithm`](@ref) and [`power`](@ref) with fractional powers) are only defined for matrices whose eigenvalues avoid (part of) the negative real axis, and their principal values are complex whenever eigenvalues on that axis are present.
+In MatrixAlgebraKit, we aim to keep type stability, and thus the scalar type of the output always matches that of the input.
```
Jutho, line 44: That is not really true, i.e. `eig_full` and `eig_trunc` will always switch to complex.

src/implementations/logarithm.jl
```diff
@@ -0,0 +1,62 @@
...
+# Implementation
+# --------------
+function logarithm!(A::AbstractMatrix, logA, alg::MatrixFunctionViaLA)
+    check_input(logarithm!, A, logA, alg)
+    isempty(alg.kwargs) || throw(ArgumentError("`MatrixFunctionViaLA` does not accept keyword arguments for `logarithm`"))
+    # `LinearAlgebra.log` of a real matrix is real whenever the principal logarithm is,
+    # so a complex result with a real output signals a genuine domain violation
+    logAc = LinearAlgebra.log(A)
+    if eltype(logAc) <: Complex && !(eltype(logA) <: Complex)
+        throw(_realness_domainerror(logarithm!))
```
Jutho: Do we need a tolerance check on the imaginary parts as for the case of `power`, or is the situation different here?

src/implementations/power.jl
```diff
@@ -0,0 +1,81 @@
...
+        # so rounding-level imaginary components do not signal a domain violation.
+        # The tolerance is based on the working precision, which may be lower than
+        # the result eltype suggests (e.g. `Float32` input promotes to `ComplexF64`).
+        atol = defaulttol(powA) * norm(powAc, Inf)
+        all(x -> abs(imag(x)) <= atol, powAc) || throw(_realness_domainerror(power!))
+        powA .= real.(powAc)
+        return powA
+    end
+    copy!(powA, powAc)
+    return powA
```
Jutho: I would prefer the following logic:

```suggestion
        powA .= real.(powAc)
    else
        copy!(powA, powAc)
    end
    return powA
```

docs/src/user_interface/matrix_functions.md
Jutho: Ok, I guess within the section of matrix functions this is true.

src/interface/matrixfunctions.jl
```diff
@@ -29,31 +29,49 @@ As this algorithm requires no LAPACK support, it also applies at arbitrary preci
...
+For matrix functions with a restricted domain (e.g. [`squareroot`](@ref) and [`logarithm`](@ref)),
+`domain_atol` specifies the absolute tolerance below which out-of-domain eigenvalues are treated
+as rounding artifacts and clamped to the domain boundary, with `nothing` denoting the default
+tolerance [`default_domain_atol`](@ref).
 """
-struct MatrixFunctionViaEigh{A <: AbstractAlgorithm} <: AbstractAlgorithm
+struct MatrixFunctionViaEigh{A <: AbstractAlgorithm, T <: Union{Nothing, Real}} <: AbstractAlgorithm
     eigh_alg::A
+    domain_atol::T
+end
```
Jutho: I guess changing this `struct` definition makes this a breaking change?

src/implementations/squareroot.jl
```diff
@@ -0,0 +1,62 @@
...
+    diag_alg = DiagonalAlgorithm(; domain_atol = alg.domain_atol)
+    return _apply_eig!(sqrtA, V, squareroot!(D, D, diag_alg))
+end
+
+# Diagonal logic
+# --------------
+function squareroot!(A::AbstractMatrix, sqrtA, alg::DiagonalAlgorithm)
+    check_input(squareroot!, A, sqrtA, alg)
+    λ = diagview(sqrtA)
+    copyto!(λ, diagview(A))
```
Jutho: Any reason for `copyto!` over `copy!`?

Jutho:
I did not go through all the test changes, also because this contains stuff from the other PR. The changes to `src`, `ext` and `docs` look good to me, though for the latter, it would be useful if we can also have a github action that builds a docs preview and put the links here, like in some of the other repos.

Jutho:
One thing I just noted in the docs is that we have a section on "Matrix Properties" and then one on "Matrix functions", i.e. we have a capitalization inconsistency 😄 .

src/interface/matrixfunctions.jl
lkdvos: I guess you are right... There might be an argument to be made for hardcoding the tolerance to `Float64`, not adding the type parameter, and then this looks non-breaking to me?

In the following round of breaking changes, we could also consider making this actual `@algdef`s, which would have solved this problem...

lkdvos:
Okay, update here: I looked at the whole `domain_tol` situation, and honestly I'm not sure how to best deal with this. Let me at least lay out some of the problems I ran into, and I'd be really happy to get some suggestions and opinions on the matter because I'm actually really not sure how to handle this.

Firstly, there is an annoying distinction between functions that have a domain that includes or excludes the boundary. For example, `squareroot` can just take negative eigenvalues and shift them to `0`, which seems sensible, while `logarithm` cannot do the same since `log(0) = -Inf`. 
As a result, for the former, increasing the tolerance can be used to still continue without erroring, while for the latter this is just not really possible (I think?).

Secondly, not all methods have access to the actual spectrum, the LinearAlgebra wrappers being a good example. As a result, for these functions I can only inspect the output `f(A)`, and try to derive from this what the tolerance was on the input. This means that tolerances have different scales depending on what `f` is, and for `logarithm` this again does not really make sense since any negative value will end up with `log(-x) = log(x) + pi * im`, so the magnitude of the imaginary part doesn't actually tell me anything about "how far from the domain" the input was.

Finally, for negative `power`s, the origin is the excluded region, so there it actually kind of makes sense to have a `tol` that means the **opposite** of what it means for `sqrt`, i.e. increasing the tolerance somehow means rejecting a bigger area, while for `sqrt` increasing the tolerance means accepting a bigger area?

So my current implementation is absolutely not consistent, and I'm not really sure what the best way forwards is to make this consistent, and I also dont know if this even matters all that much, so maybe it might be reasonable to simply have something only for `squareroot` for now?

kshyatt:
NB it would be great to hook up `pinv` to our SVD infrastructure as well

PR #262, opened 2026-07-28
Migrate test infrastructure to Pkg workspaces

lkdvos:
Replaces the legacy `[extras]`/`[targets]` mechanism with the Julia 1.12 [workspace](https://pkgdocs.julialang.org/dev/workspaces/) feature, bringing MatrixAlgebraKit in line with TensorKit.jl, MPSKit.jl, PEPSKit.jl and the rest of the org.

## Latent fixes

Three unrelated issues surfaced while moving things around:

- **The Buildkite skips in `runtests.jl` were no-ops.** `find_tests` keys tests by path relative to `test/`, but the five `delete!` calls used bare names (`"algorithms"` vs. `common/algorithms`), so none of them ever matched. Now path-qualified. Note that `common/exponential` and `common/projections` are deliberately *not* skipped: unlike the other files in `common/`, they carry `CUDA.functional()` / `AMDGPU.functional()` blocks, so skipping them would lose GPU coverage. `chainrules` is skipped as a group — its body was already guarded by `if !is_buildkite`, so this just saves spinning up a worker.
- **`test/exponential.jl` never ran on GitHub CI.** `TestGroups.yml` builds its matrix from the top-level directories under `test/` and passes the directory name as a `startswith` filter, so a root-level file belongs to no group. Moved to `test/common/exponential.jl`, will tackle this properly in #261 with a matrixfunction test group
- **`test/utilities.jl` was dead** — defined `isapproxone`, never included or called. Removed.


Jutho (approved):

test/README.md
```diff
@@ -0,0 +1,175 @@
+# MatrixAlgebraKit.jl test suite
+
+Tests are driven by [ParallelTestRunner.jl](https://github.com/JuliaTesting/ParallelTestRunner.jl).
+Every `.jl` file under `test/` is auto-discovered and executed in its own worker process, so test
+files share no state and must be self-contained.
+
+The test environment is a [Pkg workspace](https://pkgdocs.julialang.org/dev/workspaces/) member:
```
Jutho, line 7: I forgot to comment that this link doesn't work.

lkdvos: will fix this in #261

PR #266, opened 2026-08-11
Instantiate broadcast for pullbacks

kshyatt:
This prevents an `axes` check from exploding further up the stack when PEPSKit tries to call this

### DIFF
--- a/src/pullbacks/eig.jl
+++ b/src/pullbacks/eig.jl
@@ -30,7 +30,7 @@ function check_and_prepare_eig_cotangents(
     bc = Base.broadcasted(transpose(D), D, VᴴΔV₁) do d₁, d₂, v
         return abs(d₁ - d₂) < degeneracy_atol ? v : zero(v)
     end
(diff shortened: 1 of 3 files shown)


lkdvos (approved):

Jutho:
The `broadcasted` approach was suggested by @lkdvos a while ago as a way to avoid an extra copy/allocation. If we then end up instantiating it again, this seems like a convoluted way to replace the simpler code that was there before Lukas’ suggestion, but I might be misremembering.

kshyatt:
It's again our frenemy the GPU libraries not being able to "detect" that a `Broadcasted` across complicated GPU objects is itself spiritually a GPU object. If we can find what the previous implementation was (`git blame`, I guess), I can return to the drawing board and try to find a nicer way to achieve the same ends.

lkdvos:
I don't think `instantiate` here means the same as `materialize`, or at least that's what I convinced myself of when I was looking at this, and I don't think this is materialized

PR #267, opened 2026-08-12
Extract R before constructing Q in Householder QR/LQ

lkdvos:
`ormqr` (`unmqr!`) requires a workspace of ~2·m·n elements whose size can only be queried through a 32-bit integer, and cuSOLVER exposes no 64-bit variant.
Past m ≈ 35000 the query itself returns `CUSOLVER_STATUS_INVALID_VALUE`, so `qr_compact`/`qr_full` on a large `CuMatrix` cannot construct `Q` at all.

`orgqr` (`ungqr!`) needs no workspace and measured 1.15–1.5× faster at every size tested on my GPU — but it generates `Q` *in the array holding the reflectors*, which the old order (construct `Q`, then copy `R` out of the packed `A`) could not combine with returning `R`.
Note that the old order was there for the blocked version reasons, which is not on the GPU anyways and still exists for that reason.

`qr_householder!`/`lq_householder!` extract `R`/`L` before constructing `Q`.
An inplace `Q` (`Q === A`) therefore now combines with computing `R`/`L` and with `positive = true`; only `blocksize = 1`, a tall (wide) `A`, and `R`/`L` not aliasing `A` remain required. The blocked path still constructs `Q` first, since `R` doubles as workspace for `T` there.

CUSOLVER now always constructs `Q` with `ungqr!`. LAPACK keeps `unmqr!`, where the comparison is mixed (`unmqr!` up to 6× faster for small/skinny, `ungqr!` up to 1.9× faster from ~2000 up). `qr_null` keeps `unmqr!` regardless: `N` holds the trailing columns of `Q`, which `orgqr` cannot generate.

Also fixed: LQ gauge fixing ran even with `positive = false` when `L` was not computed; `Native`/`GLA` silently returned garbage for `Q === A`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -24,11 +24,20 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Changed
 
+- `qr_compact!`, `qr_full!`, `lq_compact!` and `lq_full!` now extract `R` (or `L`) before
+  constructing `Q`, so that an inplace `Q` (supplying `A` itself as output for `Q`) can be combined
+  with computing `R` (or `L`) and with `positive = true`.
+- The CUSOLVER driver constructs `Q` with `ungqr!` instead of `unmqr!`, which is both faster and
+  avoids the large workspace of `ormqr`, whose 32-bit size query fails altogether for large
+  matrices.
+
 ### Deprecated
 
 ### Removed
 
 ### Fixed
+
+- LQ decompositions no longer gauge fix `Q` when `positive = false` and `L` is not computed.
 
 ### Performance
 
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -34,8 +34,6 @@ for f in (:geqrf!, :ungqr!, :unmqr!)
 end
 
 MatrixAlgebraKit.prefers_ungqr(::CUSOLVER) = true
-MatrixAlgebraKit.supports_unmqr(::CUSOLVER, side, trans, A, τ, C) =
-    YACUSOLVER.unmqr_worksize(side, trans, A, τ, C) > 0
 
 MatrixAlgebraKit.supports_svd_full(::CUSOLVER, f::Symbol) = f in (:qr_iteration, :jacobi, :svd_polar)
 
--- a/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
@@ -33,6 +33,10 @@ for f in (:geqrf!, :ungqr!, :unmqr!)
     @eval $f(::CUSOLVER, args...) = YACUSOLVER.$f(args...)
 end
 
+MatrixAlgebraKit.prefers_ungqr(::CUSOLVER) = true
+MatrixAlgebraKit.supports_unmqr(::CUSOLVER, side, trans, A, τ, C) =
+    YACUSOLVER.unmqr_worksize(side, trans, A, τ, C) > 0
+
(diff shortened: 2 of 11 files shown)


docs/src/changelog.md
```diff
@@ -24,12 +24,21 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Changed
 
+- `qr_compact!`, `qr_full!`, `lq_compact!` and `lq_full!` now extract `R` (or `L`) before
+  constructing `Q`, so that an inplace `Q` (supplying `A` itself as output for `Q`) can be combined
+  with computing `R` (or `L`) and with `positive = true`.
+- The CUSOLVER driver constructs `Q` with `ungqr!` instead of `unmqr!`, which is both faster and
```
kshyatt, line 30: Is this a problem ROCSolver also has?

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -33,6 +33,8 @@ for f in (:geqrf!, :ungqr!, :unmqr!)
     @eval $f(::CUSOLVER, args...) = YACUSOLVER.$f(args...)
 end
 
+MatrixAlgebraKit.prefers_ungqr(::CUSOLVER) = true
```
kshyatt, line 36: Again, what about ROCSolver?

src/implementations/lq.jl
```diff
@@ -144,34 +151,39 @@ function lq_householder!(
     computeL = length(L) > 0
     inplaceQ = Q === A
 
-    (inplaceQ && (computeL || positive || blocksize > 1 || n < m)) &&
-        throw(ArgumentError("inplace Q only supported if matrix is wide (`m <= n`), L is not required, and using the unblocked algorithm (`blocksize = 1`) with `positive = false`"))
+    if inplaceQ
+        # unglq! builds Q in the space of A, so L has to be extracted first and cannot alias A
+        (blocksize == 1 && n >= m) ||
+            throw(ArgumentError("inplace Q only supported if matrix is wide (`m <= n`) and using the unblocked algorithm (`blocksize = 1`)"))
```
kshyatt: should we actually print the values of `m`, `n`, and `blocksize`?

src/implementations/lq.jl
```diff
@@ -144,34 +151,39 @@ function lq_householder!(
...
+        (computeL && Base.mightalias(L, A)) &&
+            throw(ArgumentError("inplace Q only supported if L does not share memory with A"))
```
kshyatt: "inplace Q is only supported" sounds more natural to me

src/implementations/lq.jl
```diff
@@ -303,6 +317,9 @@ end
 function lq_via_qr!(
         A::AbstractMatrix, L::AbstractMatrix, Q::AbstractMatrix, qr_alg::AbstractAlgorithm
     )
+    # Q is written before L, so an L that aliases A would corrupt an inplace Q
+    (A === Q && !isempty(L) && Base.mightalias(L, A)) &&
+        throw(ArgumentError("inplace Q only supported if L does not share memory with A"))
```
kshyatt: "is only supported" again (imo)

docs/src/changelog.md
lkdvos: I actually don't know, and since I don't really have a way of trying this out I decided to punt on this. It's only really a problem if you have more than ~2^30 rows, so it is somewhat niche anyways?

kshyatt: It might not be if someone tries a really chonky PEPS on LUMI 😹 . I have access to an AMD card, if you have a MWE I can test if it also applies to ROCSolver?

lkdvos:
```julia
using AMDGPU, MatrixAlgebraKit
A = AMDGPU.rand(2^32, 4)
Q, R = qr_compact!(A)
```
(Needs to be a big GPU to make that fit)

kshyatt: OK my card is not big enough, let's let this lie for now and we can always update later.

kshyatt (approved):

PR #268, opened 2026-08-13
Update pipeline.yml

kshyatt:
Don't even try running Mooncake + ROCm, save several minutes per build


kshyatt:
OK this has the behaviors I wanted, no need to waste more AMDGPU time. I will force-merge.

PR #270, opened 2026-08-17
[WIP] Allow turning off allowscalar copies for yacusolver

kshyatt:
The idea of this PR is to make the checks involving `@allowscalar` optional. These copies are extremely expensive and block the device, so by (optionally) disabling them, we should be able to make better use of the GPU.

So far I've just done the CUSOLVER ones, but I can also do ROCSOLVER if we like the idea. I think it would make sense to have a settable flag that turns the checks on and off so that users can control it, or a `CheckedCUSOLVER` driver. Anyone have thoughts on that angle?

### DIFF
--- a/ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
+++ b/ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
@@ -26,7 +26,8 @@ for (bname, fname, elty, relty) in
                 A::StridedCuMatrix{$elty},
                 S::StridedCuVector{$relty} = similar(A, $relty, min(size(A)...)),
                 U::StridedCuMatrix{$elty} = similar(A, $elty, size(A, 1), min(size(A)...)),
-                Vᴴ::StridedCuMatrix{$elty} = similar(A, $elty, min(size(A)...), size(A, 2))
+                Vᴴ::StridedCuMatrix{$elty} = similar(A, $elty, min(size(A)...), size(A, 2));
+                check::Bool = false,
             )
             chkstride1(A, U, Vᴴ, S)
             m, n = size(A)
@@ -90,8 +91,10 @@ for (bname, fname, elty, relty) in
             end
             CUDA.unsafe_free!(rwork)
 
-            info = @allowscalar dh.info[1]
-            cuSOLVER.chkargsok(BlasInt(info))
+            if check
+                info = @allowscalar dh.info[1]
+                cuSOLVER.chkargsok(BlasInt(info))
+            end
 
             return (S, U, Vᴴ)
         end
@@ -103,7 +106,8 @@ function gesvdp!(
         S::StridedCuVector = similar(A, real(T), min(size(A)...)),
         U::StridedCuMatrix{T} = similar(A, T, size(A, 1), min(size(A)...)),
         Vᴴ::StridedCuMatrix{T} = similar(A, T, min(size(A)...), size(A, 2));
-        tol = norm(A) * eps(real(T))
+        tol = norm(A) * eps(real(T)),
+        check::Bool = false,
     ) where {T <: BlasFloat}
     chkstride1(A, U, S, Vᴴ)
     m, n = size(A)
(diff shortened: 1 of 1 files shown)


lkdvos (commented):
Definitely looks reasonable to me, although it would be nice to get a sense of how bad the performance is to try and decide if we want to turn this on or off by default.

Is this specifically for the case where we apply the factorizations to a bunch of blocks and don't want to sync inbetween? There might also be a case for writing a batched version that tries to only sync at the end or something similar

kshyatt:
> Is this specifically for the case where we apply the factorizations to a bunch of blocks and don't want to sync inbetween?

Indeed it helps with this but not only here, another example is `calc_convergence` in PEPSKit where we're computing `svd_vals` 8 times (4 corners, 4 edges).

lkdvos:
Yeah fair enough, I definitely think the keyword argument is a good solution, just slightly on the fence about disabling checks by default. 🤷 it's a bit annoying to think that PEPSKit would run without these and then if at some point something doesn't converge it just silently goes haywire, and then you have to do a full rerun with checks on to figure out what happened, but I guess that's what it has to be

kshyatt:
I agree, that's why I think in fact this should be settable at `__init__` or with an env var. But what to call it?

Jutho:
It seems like all four of `cuSOLVER.chkargsok`, `cuSOLVER.chklapackerror`, `chkargsok` and `chklapackerror` are used. We should probably replace this to something more consistent. Are `chklapackerror` and `chkargsok` equivalent? Is one the `LinearAlgebra.LAPACK` version and the other the `cuSOLVER` version?

Jutho:
Also, I don't fully understand why this needs to be so expensive. Is there something with `@allowscalar` that would be different then e.g. `only(collect(info))` or `only(copy!(Vector{Int}(undef,1), info))`. It does seem like this is the standard way that `gesvd` is called:
https://github.com/NVIDIA/CUDALibrarySamples/blob/6e7b3fa0debcfc857495b57df196a278d1931e5a/cuSOLVER/gesvd/cusolver_gesvd_example.cu#L110

lkdvos:
The expensive part is less about the cost of the operation, and more about the fact that sending back a scalar requires synchronization, so it breaks the asynchronous programming model, which im assuming is dominating the runtime for the tail of small blocks. Probably this can be improved in a batched implementation, but I do see the value in having a switch for this, although I do want the default to be having the checks on

kshyatt:
Indeed I think `@allowscalar` blocks and synchronizes the entire device, so it's quite painful from an "overlapping work" point of view...

src/MatrixAlgebraKit.jl
```diff
@@ -145,4 +145,19 @@ include("pushforwards/svd.jl")
...
+perform expensive checking operations after library calls. These libraries implement
+the check by doing an `@allowscalar` read of a GPU array, which forces device
+synchronization. Default is `Ref(true)` (checks are performed), but this can be
+changed by setting `MatrixAlgebraKit.CHECK_LIBRARY_CALLS[] = false`.
+
+!!! warning
+  Disabling the checks means you may encounter errors later in program execution,
+  far from their original source, which will make debugging substantially more difficult.
+"""
+const CHECK_LIBRARY_CALLS = Ref(true)
```
lkdvos, line 161: Would it make sense to turn this into a `ScopedValue` instead of a global switch?

lkdvos (approved):
Overall I'm definitely okay with this, although it would really be nice to have other ways of circumventing this. For example, can we do a non-blocking `NaN`-poison of the results? My main reservations are that we do actually have a LAPACK error every now and again, and I have a hard time gauging how much of a pain it is to not have this error and simply produce garbage.

Ultimately, I think the real win is still the batched solution, but since this PR is such a small change and really doesn't hurt I don't see why not :)

src/MatrixAlgebraKit.jl
kshyatt: Will that play nicely with the extension? I confess I didn't want to look it up

lkdvos: I don't think this is a problem from what I remember, it really is just about how you want to be able to control the flag

kshyatt: TBH I'm not sure, fwiw [TensorKit](https://github.com/QuantumKitHub/TensorKit.jl/blob/main/src/TensorKit.jl#L227) also just uses the `Ref` pattern for now.

Jutho (approved):

kshyatt:
OK, let me add the ROCm checks too tomorrow

lkdvos:
Test failures look unrelated, probably ok to force merge

PR #271, opened 2026-08-17
Make svd_trunc pullback GPU friendly

kshyatt:
Otherwise this hits scalar indexing

### DIFF
--- a/src/pullbacks/svd.jl
+++ b/src/pullbacks/svd.jl
@@ -241,8 +241,9 @@ function svd_trunc_pullback!(
         Y₀ᴴ = iszerotangent(ΔV₊ᴴ) ? zero(Vᴴ) : ldiv!(Diagonal(S), ΔV₊ᴴ)
         US = mul!(ΔAV, U, Smat) # recycle ΔAV
         AP = mul!(copy(A), US, Vᴴ, -1, 1)
-        AP ./= S[end]
-        S⁻¹ = S[end] ./ S
+        minS = minimum(S)
+        AP ./= minS
+        S⁻¹ = minS ./ S
         X₁ = rmul!(AP * Y₀ᴴ', Diagonal(S⁻¹))
         X₁ .+= X₀
         Y₁ᴴ = lmul!(Diagonal(S⁻¹), X₀' * AP)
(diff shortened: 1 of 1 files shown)


src/pullbacks/svd.jl
```diff
@@ -241,8 +241,9 @@ function svd_trunc_pullback!(
         Y₀ᴴ = iszerotangent(ΔV₊ᴴ) ? zero(Vᴴ) : ldiv!(Diagonal(S), ΔV₊ᴴ)
         US = mul!(ΔAV, U, Smat) # recycle ΔAV
         AP = mul!(copy(A), US, Vᴴ, -1, 1)
-        AP ./= S[end]
-        S⁻¹ = S[end] ./ S
+        minS = minimum(S)
```
lkdvos: Does `last` also work? Can we avoid blocking/syncing by using a view of the last element?

kshyatt: I don't think so 😬 let me try and report back

kshyatt: I think the problem is any of these require copying the last element back to the CPU, unless we write a custom kernel/map here which I'm kind of loathe to do...

lkdvos: But isn't that also what minimum has to do, on top of traversing the entire array?
I was kind of hoping we could do something like `AP ./= S[end:end]` , (or maybe put that element in a 0-dim array?) and if nothing else just get a gpu friendly last with allowscalar since there's not much better we can do

kshyatt: Yeah I see, I also wanted to do this in a way that didn't require a separate GPU method... I'll try the view again

lkdvos: absolutely fair enough. At some point we might start considering just having GPUArrays as a hard dependency, so we can actually just have `@allowscalar` there...

src/pullbacks/svd.jl
```diff
@@ -241,8 +241,9 @@ function svd_trunc_pullback!(
         Y₀ᴴ = iszerotangent(ΔV₊ᴴ) ? zero(Vᴴ) : ldiv!(Diagonal(S), ΔV₊ᴴ)
         US = mul!(ΔAV, U, Smat) # recycle ΔAV
         AP = mul!(copy(A), US, Vᴴ, -1, 1)
-        AP ./= S[end]
-        S⁻¹ = S[end] ./ S
+        minS = S[end:end]
```
```suggestion
        @view minS = S[end:end]
```

It seems like it worked, I think these GPU failures are unrelated, but now we are still allocating a slice, so maybe even that can be avoided?

lkdvos (approved):

kshyatt:
GPUArrays released a new branch, let me quickly rerun the failing tests and see if they resolve?

lkdvos:
Am I reading this correctly that the failing tests are not related to this PR, and everything works, or should we still wait for a rerun of these tests?

kshyatt:
I think you're reading it correctly indeed, sorry, this fell off my radar

PR #272, opened 2026-08-18
Work around GPUArrays change in lq.jl

kshyatt:
I think we had this change already for qr. The new GPUArrays version broke things on GPU, this should fix.


kshyatt:
Since the AMD queue is super backed up and it was the Mooncake CUDA test failing, I would merge this if the latter passes even if the former hasn't run.

lkdvos (approved):
I do remember being annoying about trying to avoid the copy in the previous QR implementation too, so lets not relitigate this :)

PR #273, opened 2026-08-18
Add GPU-safe index selection for eig/eigh pullbacks

kshyatt:
Let's see if this removes the need to copy back for the PB steps


kshyatt:
Blocked on https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/272 but this seems to work!!!

Jutho:
This looks like a nice solution.

PR #274, opened 2026-08-19
Set minimum CUDA to 6.3

kshyatt:
Should pick up https://github.com/JuliaGPU/CUDA.jl/pull/3240 and fix the ambiguity error


kshyatt:
Unfortunately, this requires GPUCompiler 2.x which Enzyme doesn't support 😱

PR #275, opened 2026-08-21
Batched SVD support for ROCSOLVER and CUSOLVER

kshyatt:
Basically what it says on the tin. For CTMRG and other algorithms, we're getting absolutely slaughtered on GPU performance for TensorMaps with sectors because we have to spin up huge numbers of very small SVDs. I'm wrapping the batched SVDs each library provides to try to address this. Extremely open to comments but I wanted to get this rolling so I can unblock others.

### DIFF
--- a/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
+++ b/ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
@@ -6,7 +6,8 @@ using MatrixAlgebraKit: one!, zero!, uppertriangular!, lowertriangular!
 using MatrixAlgebraKit: diagview, sign_safe
 using MatrixAlgebraKit: ROCSOLVER, LQViaTransposedQR, TruncationStrategy, NoTruncation, TruncationByValue, AbstractAlgorithm
 using MatrixAlgebraKit: default_qr_algorithm, default_lq_algorithm, default_svd_algorithm, default_eigh_algorithm
-import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesvdj!
+import MatrixAlgebraKit: geqrf!, ungqr!, unmqr!, gesvd!, gesdd!, gesvdx!, gesvdj!
+import MatrixAlgebraKit: gesvdj_batched!, gesvdx_batched!, gesdd_batched!, gesvd_batched!
 import MatrixAlgebraKit: heevj!, heevd!, heev!, heevx!
 import MatrixAlgebraKit: _sylvester, svd_rank, svd_pullback!
 using AMDGPU
@@ -15,10 +16,17 @@ using LinearAlgebra: BlasFloat
 
 include("yarocsolver.jl")
 
-MatrixAlgebraKit.default_driver(::Type{TA}) where {TA <: StridedROCVecOrMat{<:BlasFloat}} = ROCSOLVER()
+MatrixAlgebraKit.default_driver(::Type{TA}) where {TA <: StridedROCArray{<:BlasFloat}} = ROCSOLVER()
+MatrixAlgebraKit.default_driver(::Type{TA}) where {TA <: AbstractVector{<:StridedROCMatrix{<:BlasFloat}}} = ROCSOLVER()
 
-function MatrixAlgebraKit.default_svd_algorithm(::Type{T}; kwargs...) where {T <: StridedROCVecOrMat{<:BlasFloat}}
-    return QRIteration(; kwargs...)
+function MatrixAlgebraKit.default_svd_algorithm(::Type{T}; kwargs...) where {T <: StridedROCMatrix{<:BlasFloat}}
+    return DivideAndConquer(; kwargs...)
+end
+function MatrixAlgebraKit.default_svd_algorithm(::Type{T}; kwargs...) where {T <: StridedROCArray{<:BlasFloat, 3}}
+    return DivideAndConquerBatched(; kwargs...)
+end
+function MatrixAlgebraKit.default_svd_algorithm(::Type{T}; kwargs...) where {T <: AbstractVector{<:StridedROCMatrix{<:BlasFloat}}}
+    return DivideAndConquerBatched(; kwargs...)
 end
 function MatrixAlgebraKit.default_eigh_algorithm(::Type{T}; kwargs...) where {T <: StridedROCVecOrMat{<:BlasFloat}}
     return DivideAndConquer(; kwargs...)
@@ -28,7 +36,7 @@ for f in (:geqrf!, :ungqr!, :unmqr!)
     @eval $f(::ROCSOLVER, args...) = YArocSOLVER.$f(args...)
 end
 
-MatrixAlgebraKit.supports_svd_full(::ROCSOLVER, f::Symbol) = f in (:qr_iteration, :jacobi)
+MatrixAlgebraKit.supports_svd_full(::ROCSOLVER, f::Symbol) = f in (:qr_iteration, :jacobi, :divide_and_conquer, :qr_iteration_batched, :jacobi_batched, :divide_and_conquer_batched)
 
 function gesvd!(::ROCSOLVER, A::StridedROCMatrix, S::StridedROCVector, U::StridedROCMatrix, Vᴴ::StridedROCMatrix; kwargs...)
     m, n = size(A)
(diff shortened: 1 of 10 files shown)


kshyatt:
TODOs here:
- Finish wrapping the `BisectionBatched` logic for AMD
- Add the checks for CUSOLVER `gesvdj_batched` (blocks may not be larger than 32 x 32)
- Finish the tests for `svd_trunc!`
- Pullback/pushforward rules

Since I'll be out for 3 weeks everyone should feel free to just push to this.

test/testsuite/decompositions/svd.jl
```diff
@@ -54,6 +74,47 @@ function test_svd_compact(
...
 end
 
+function test_svd_compact_batched(
+        T::Type, sz, batch_size::Int;
+        atol::Real = 0, rtol::Real = precision(eltype(T)),
+        test_vals::Bool = true, kwargs...
+    )
+    summary_str = testargs_summary(T, sz)
+    return @testset "svd_compact! $summary_str batch_size $batch_size" begin
+        As = [instantiate_matrix(T, sz) for bi in 1:batch_size]
```
lkdvos, line 84: should we try and vary the sizes within the batch?

test/testsuite/TestSuite.jl
```diff
@@ -42,6 +42,20 @@ instantiate_matrix(::Type{AT}, size) where {AT <: Diagonal} = Diagonal(randn(rng
...
+function _collect_batch(As::AbstractVector{<:AbstractMatrix})
+    B = similar(first(As), (size(first(As))..., length(As)))
+    for (i, A) in enumerate(As)
+        copyto!(view(B, :, :, i), A)
+    end
+    return B
+end
+device_batch(As::AbstractVector{<:Array}) = _collect_batch(As)
+device_batch(As::AbstractVector{<:CuArray}) = _collect_batch(As)
+device_batch(As::AbstractVector{<:ROCArray}) = _collect_batch(As)
```
lkdvos, line 57: Is this the same as `stack(As; dims = 3)`, and does that work on both devices and host arrays?

src/implementations/svd.jl
```diff
@@ -42,6 +44,80 @@ function check_input(::typeof(svd_vals!), A::AbstractMatrix, S, ::AbstractAlgori
     return nothing
 end
 
+# batched varieties
+function check_input(::typeof(svd_full!), A::AbstractVector{<:AbstractMatrix}, USVᴴ, ::AbstractAlgorithm)
+    @assert all(==(size(first(A))), size.(A))
```
lkdvos, line 51: Why do we require the sizes to be the same? I'm assuming this is some restriction of some of the drivers, but that does kind of seems to contradict the requirements we need in TensorKit

src/implementations/svd.jl
```diff
@@ -120,10 +237,17 @@ end
...
     @eval $f!(driver::Driver, args...) = throw(ArgumentError("$driver does not provide $($(f!))"))
 end
 
+# declare these as dummies so the GPU extensions can import them safely
+function gesvd_batched! end
+function gesdd_batched! end
+function gesvdj_batched! end
+function gesvdx_batched! end
+
+
```
lkdvos: Isn't this already defined on line 240-241?

lkdvos (commented):
Overall looks great, I think the main thing I am wondering about is about the interface decision around wether we implement this as a set of `BatchedAlg` versions, or rather as a set of `batched_f(...)` functions. I definitely like using dispatch for switching between the strided and non-strided inputs, but I am wondering if there might be a benefit to really having a `batched_svd_compact` etc function.

This is also partially since that allows us to have a CPU version for this as well, so we can just offload all of this from TensorKit to here, (and possibly play with multithreading?).

Jutho:
I mostly agree with Lukas here, I think I would prefer
* separate `batched_f` methods instead of separate algorithms (with possible options provided in the driver rather than the algorithm if we need those)
* support for varying sizes, which we will anyway need to implement, so we can do it on the MAK side I would think

kshyatt:
> support for varying sizes, which we will anyway need to implement, so we can do it on the MAK side I would think

This is already implemented in a separate branch I have over at TensorKit, I can move it over here of course!

src/implementations/svd.jl
kshyatt: It's a limitation of the drivers, you can work around it by padding the 3D Array with zeros as needed

test/testsuite/TestSuite.jl
kshyatt: I think the GPU arrays don't support `stack` 😭

kshyatt:
Latest commit gets rid of the `JacobiBatched` and friends algos, and adds instead `batched_svd_compact` etc. Remaining TODOs:

- [x] Add tests for batches with varying sizes of input matrix (done!)
- [x] Add padding support in here, rather than in TensorKit (done!)
- [x] Add support for `svd_via_adjoint!` for batched arrays (done!)

kshyatt:
One remaining thing to do here is restore the specific path that targets the AMDGPU `ROCVector{<:ROCMatrix}` path, otherwise I think things are looking ok. `stack` not only isn't supported on GPU, it doesn't work for the ragged case (unless we pad separately for each input, which I think is inefficient?).

kshyatt:
Moving the bisection (batched and non) logic out of here to make this easier to understand

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -16,18 +17,30 @@ using LinearAlgebra: BlasFloat
...
     return DivideAndConquer(; kwargs...)
 end
 
+function MatrixAlgebraKit.one!(A::StridedCuArray{T, 3}) where {T <: BlasFloat}
+    length(A) > 0 || return A
+    zero!(A)
+    # TODO use mapslices?
+    for a in eachslice(A, dims = 3)
+        diagview(a) .= one(eltype(a))
+    end
```
lkdvos:
```suggestion
    # TODO use mapslices?
    for a in eachslice(A, dims = 3)
        one!(a)
    end
```

Does this also work?

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -15,20 +16,42 @@ using LinearAlgebra: BlasFloat
...
     return DivideAndConquer(; kwargs...)
 end
 
+function MatrixAlgebraKit.one!(A::StridedROCArray{T, 3}) where {T <: BlasFloat}
+    length(A) > 0 || return A
+    zero!(A)
+    # TODO use mapslices?
+    for a in eachslice(A, dims = 3)
+        diagview(a) .= one(eltype(a))
+    end
```
lkdvos:
```suggestion
    foreach(one!, eachslice(A, dims = 3))
```

I would have to check but this might even remove the need for the `length(A) > 0)` check?

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
```diff
@@ -16,18 +17,30 @@ using LinearAlgebra: BlasFloat
...
+    return Jacobi(; kwargs...)
+end
 function MatrixAlgebraKit.default_eig_algorithm(::Type{T}; kwargs...) where {T <: StridedCuVecOrMat{<:BlasFloat}}
     return QRIteration(; kwargs...)
 end
 function MatrixAlgebraKit.default_eigh_algorithm(::Type{T}; kwargs...) where {T <: StridedCuVecOrMat{<:BlasFloat}}
     return DivideAndConquer(; kwargs...)
 end
 
+function MatrixAlgebraKit.one!(A::StridedCuArray{T, 3}) where {T <: BlasFloat}
```
lkdvos: Probably we can directly define this for `AbstractMatrix{T, 3}`?

ext/MatrixAlgebraKitCUDAExt/yacusolver.jl
```diff
@@ -276,6 +276,90 @@ for (bname, fname, elty, relty) in
...
+            cuSOLVER.with_workspace(dh.workspace_gpu, bufferSize) do buffer
+                return cuSOLVER.$fname(
+                    dh, jobz, m, n, A, lda, S, Ũ, ldu, Ṽ, ldv,
+                    buffer, sizeof(buffer) ÷ sizeof($elty), dh.info,
+                    params[], batch_size
+                )
+            end
+
+            info = collect(dh.info)
+            cuSOLVER.chkargsok.(BlasInt.(info))
```
lkdvos:
```suggestion
            foreach(cuSOLVER.chkargsok ∘ BlasInt, info)
```
I think otherwise the broadcast allocates a vector which we probably don't need?

lkdvos (commented):
Am I correct that we are still assuming all arrays to be of the same size for the batched svd? I know this is more in line with what vendors provide, but I don't immediately see how that will be useful for TensorKit, in general we have completely different sizes for the blocks?

kshyatt:
We now do support ragged block sizes but the vendors only accept uniform batches, so the padding support is now present here. For lots of small blocks I think this still gives a performance benefit because the latency of launching a lot of small, individual SVDs is extremely bad.

src/implementations/batched_svd.jl
```diff
@@ -0,0 +1,452 @@
+# Inputs
+# ------
+copy_input(::typeof(batched_svd_full), As::AbstractVector{<:AbstractMatrix}) = map(A -> copy!(similar(A, float(eltype(A))), A), As)
```
lkdvos:
```suggestion
copy_input(::typeof(batched_svd_full), As::AbstractVector{<:AbstractMatrix}) = map(Base.Fix1(copy_input, svd_full), As)
```

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
```diff
@@ -48,6 +71,31 @@ function gesvdx!(::ROCSOLVER, A::StridedROCMatrix, S::StridedROCVector, U::Strid
     return S, U, Vᴴ
 end
 
+# rocSOLVER's batched `gesvd` requires m ≥ n, so wide matrices go through the adjoint
+function gesvd_batched!(::ROCSOLVER, As::AbstractVector{<:StridedROCMatrix}, Ss::StridedROCMatrix, Us::StridedROCArray{T, 3}, Vᴴs::StridedROCArray{T, 3}; kwargs...) where {T <: BlasFloat}
```
Jutho: Do the output arrays always need to be in the form of a rank-3 array, or can they also be lists of pointers?

ext/MatrixAlgebraKitAMDGPUExt/yarocsolver.jl
```diff
@@ -97,6 +97,417 @@ for (fname, elty, relty) in
...
+                        jobvt = rocSOLVER.rocblas_svect_singular
+                    end
+                elseif size(Vᴴ, 1) == n
+                    jobvt = rocSOLVER.rocblas_svect_all
+                else
+                    throw(DimensionMismatch("invalid row size of Vᴴ"))
+                end
+            end
+            length(S) == minmn * length(A) ||
+                throw(DimensionMismatch("length mismatch between A and S"))
```
Jutho: Don't we want `size(A) == (minmn, length(A))`?

Also, should we check `size(U, 3) == size(Vᴴ, 3) == length(A)` for consistency?

ext/MatrixAlgebraKitAMDGPUExt/MatrixAlgebraKitAMDGPUExt.jl
kshyatt: https://rocm.docs.amd.com/projects/rocSOLVER/en/latest/reference/lapack.html#rocsolver-type-gesvdx-batched I think they need to be rank-3

ext/MatrixAlgebraKitCUDAExt/MatrixAlgebraKitCUDAExt.jl
Jutho: I am not very comfortable with having `one!` operator `AbstractMatrix{T,3}` nor any of the `GPUArray{T<3}` subtypes. It really is the in-place counterpart to `Base.one`, which also doesn't work on this. I think the oneliner by Lukas in the comment above, i.e. `foreach(one!, eachslice(A, dims = 3))` is sufficiently short to be used where necessary, and if not, then I would like for this to be a private `_one!` method.

lkdvos: `batched_one!`?

PR #277, opened 2026-08-26
Implement schur_full for Diagonal inputs, deprecate schur_vals

### LINKED ISSUE #276: `schur_full`/`schur_vals` select `DiagonalAlgorithm` for `Diagonal` but it is not implemented
`schur_full`/`schur_vals` select `DiagonalAlgorithm` for `Diagonal` inputs, but no such implementation exists, so both throw a `MethodError`.

## Reproduction

On MatrixAlgebraKit v0.6.9, Julia 1.12.6:

```julia
julia> using LinearAlgebra, MatrixAlgebraKit

julia> D = Diagonal([2.0, 3.0, 5.0]);

julia> MatrixAlgebraKit.select_algorithm(MatrixAlgebraKit.schur_full!, D, nothing)
DiagonalAlgorithm{@NamedTuple{}}(NamedTuple())

julia> schur_full(D)
ERROR: MethodError: no method matching schur_full!(::Diagonal{Float64, Vector{Float64}}, ::Tuple{Diagonal{Float64, Vector{Float64}}, Matrix{Float64}, Vector{ComplexF64}}, ::DiagonalAlgorithm{@NamedTuple{}})

julia> schur_vals(D)
ERROR: MethodError: no method matching schur_vals!(...)

julia> eig_full(D)   # control: works
```

## Cause

Schur reuses eig's algorithm selection, which routes `Diagonal` to `DiagonalAlgorithm`:

- `src/interface/schur.jl:44-48` — `default_algorithm(::typeof(schur_full!), ::Type{A}; kwargs...) = default_eig_algorithm(A; kwargs...)`, for both `schur_full!` and `schur_vals!`
- `src/interface/eig.jl:167-169` — `default_eig_algorithm(::Type{T}) where {T <: Diagonal} = DiagonalAlgorithm(; kwargs...)`

But `src/implementations/schur.jl` only implements `QRIteration` (lines 92 and 97). There is no `DiagonalAlgorithm` method, nor `check_input`/`initialize_output` for it — the existing ones at lines 7-38 are `::AbstractAlgorithm`, so selection succeeds and dispatch then fails.

`eig`, `eigh`, `svd`, `qr` and `lq` all have `DiagonalAlgorithm` implementations; schur appears to be the one that was missed.

## Suggested fix

Add the diagonal case to `src/implementations/schur.jl`. A diagonal matrix is already in Schur form, so `T = A`, `Z = I`, `vals = complex.(diagview(A))` — mirroring `eigh_full!(A::Diagonal, DV, ::DiagonalAlgorithm)` at `src/implementations/eigh.jl:179`. Unlike `eigh`, no sorting is implied by the Schur interface.

Alternatively, give schur its own `default_schur_algorithm` instead of reusing `default_eig_algorithm`, so `Diagonal` no longer selects an unimplemented algorithm. That turns the `MethodError` into a clearer one but still leaves `schur_full(::Diagonal)` unsupported.

Happy to open a PR for whichever direction you prefer.

Note: `src/interface/schur.jl:23` already carries `# TODO: is this useful? Is there any difference with simply eig_vals?` above `schur_vals`. If the answer is to drop `schur_vals`, that covers half of this.

## Downstream impact

This reaches `TensorKit.DiagonalTensorMap`, which forwards to `select_algorithm` like every other factorization:

```julia
julia> using TensorKit

julia> d = DiagonalTensorMap(randn(4), Vect[Z2Irrep](0 => 2, 1 => 2));

julia> schur_full(d)   # MethodError; eig_full(d) and svd_compact(d) are fine
```

Found via `JET.report_package(TensorKit)`.

lkdvos:
Fixes #276: `Diagonal` input selects `DiagonalAlgorithm`, which schur never implemented, so `schur_full` threw a `MethodError`.

A diagonal matrix is already in Schur form, so `T = A`, `Z = I` (kept `Diagonal`, as `qr`/`lq` keep `Q` and `R`) and `vals = diagview(A)`, complexified as everywhere else. One choice worth flagging: **no reordering**, unlike the `Diagonal` path of `eig`, which sorts.

Also **deprecates `schur_vals`/`schur_vals!` in favour of `eig_vals`/`eig_vals!`**, answering the `# TODO: is this useful? Is there any difference with simply eig_vals?` that stood above it. There is a difference, and it is not in schur's favour: LAPACK's `gees` balances with permutation only where `geev` also scales, so `schur_vals` returns exactly `eig_vals(A; scale = false)`. On a matrix graded over 1e12 that is a relative error of 33 against 1e-15 for `eig_vals`, and it is no faster — `gees` computes the Schur form where `geev` asks `hseqr` for eigenvalues only. `schur_full` still returns the `gees` eigenvalues alongside `T` and `Z`. Callers of `schur_vals(A; expert = true)` need to drop the keyword; `eig_vals` has no equivalent.

Enables the `Diagonal` schur tests that were commented out as "not supported yet", CPU and CUDA. Note this alone does not fix `schur_full(::DiagonalTensorMap)`: TensorKit omits schur from its `default_algorithm` forwarding, so selection fails there for every tensor, diagonal or not. That needs a separate PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -37,6 +37,7 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Fixed
 
+- `schur_full` and `schur_vals` now support `Diagonal` inputs through `DiagonalAlgorithm` ([#276](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/276)).
 - LQ decompositions no longer gauge fix `Q` when `positive = false` and `L` is not computed.
 
 ### Performance
--- a/docs/src/changelog.md
+++ b/docs/src/changelog.md
@@ -33,11 +33,16 @@ When releasing a new version, move the "Unreleased" changes to a new version sec
 
 ### Deprecated
 
+- `schur_vals` and `schur_vals!` in favour of `eig_vals` and `eig_vals!`. LAPACK's `gees` balances
+  without scaling, so the eigenvalues it returns are those of `eig_vals(A; scale = false)`, which is
+  considerably less accurate for badly scaled matrices and no faster. Note that `eig_vals` does not
+  accept the `expert` keyword argument.
+
 ### Removed
 
 ### Fixed
 
-- `schur_full` and `schur_vals` now support `Diagonal` inputs through `DiagonalAlgorithm` ([#276](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/276)).
+- `schur_full` now supports `Diagonal` inputs through `DiagonalAlgorithm` ([#276](https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/issues/276)).
 - LQ decompositions no longer gauge fix `Q` when `positive = false` and `L` is not computed.
 
 ### Performance
--- a/docs/src/user_interface/decompositions.md
+++ b/docs/src/user_interface/decompositions.md
@@ -113,11 +113,10 @@ The [Schur decomposition](https://en.wikipedia.org/wiki/Schur_decomposition) tra
 It rewrites an arbitrary complex square matrix as unitarily similar to an upper triangular matrix whose diagonal elements are the eigenvalues of `A`.
 For real matrices, the same decomposition can be achieved in real arithmetic by allowing `T` to be quasi-upper triangular, i.e. triangular with blocks of size `(1, 1)` and `(2, 2)` on the diagonal.
 
-This decomposition is also useful for computing the eigenvalues of a matrix, which is exposed through the [`schur_vals`](@ref) function.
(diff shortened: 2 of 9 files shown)


Jutho:
What is the rational for `schur_vals` of a real diagonal not changing to complex? `eig_vals` does do this, and `schur_vals` on the equivalent full matrix would also return complex results.

lkdvos:
No rationale... I forgot we had this whole discussion with the eigenvalues as well, I'll change it

lkdvos:
After some reconsideration I decided to just deprecate the `schur_vals` altogether, since it is just the same as `eig_vals` anyways, and we already had a `TODO` comment somewhere in the source about this. I quite like deleting more code, so this seems like a good thing?

Jutho (approved):

PR #281, opened 2026-09-17
Support Bisection for ROCSOLVER

kshyatt:
Moving this out of https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/pull/275 since it's not really batching.

Wrapped the `Bisection()` algo for ROCm, and added some helper functions in `src/implementations/svd.jl` so that it can also be used for `svd_full!`. I put the helpers there since NVIDIA might add a similar algorithm someday, or we could also allow LAPACK users to combine the two. Can remove that if we don't want it. Also added tests.


ext/MatrixAlgebraKitAMDGPUExt/yarocsolver.jl
```diff
@@ -175,6 +175,109 @@ for (fname, elty, relty) in
...
+"""
+    _gesvdx_zero_unconverged!(S, nsv)
+
+Zero the entries of `S` that `gesvdx` did not write.
+"""
+function _gesvdx_zero_unconverged!(S::StridedROCVector, nsv::ROCVector{Cint})
+    nv = @allowscalar Int(nsv[1])
+    nv < length(S) && fill!(view(S, (nv + 1):length(S)), zero(eltype(S)))
+    return S
+end
```
lkdvos: Do we want to just inline this function?

ext/MatrixAlgebraKitAMDGPUExt/yarocsolver.jl
```diff
@@ -175,6 +175,109 @@ for (fname, elty, relty) in
...
+        vu = convert(T, get(kwargs, :vu, +Inf))
+        return rocSOLVER.rocblas_srange_value, vl, vu, 0, 0
+    else
+        return rocSOLVER.rocblas_srange_all, zero(T), zero(T), 0, 0
+    end
+end
+
+function _gesvdx_maxnsv(srange, il::Integer, iu::Integer, minmn::Integer)
+    return srange == rocSOLVER.rocblas_srange_index ? iu - il + 1 : minmn
+end
```
lkdvos: I think this is only used once, can we just inline this? I feel like the code is easier to read than the function name :)

ext/MatrixAlgebraKitAMDGPUExt/yarocsolver.jl
```diff
@@ -175,6 +175,109 @@ for (fname, elty, relty) in
...
+                U::StridedROCMatrix{$elty} = similar(A, $elty, size(A, 1), min(size(A)...)),
+                Vᴴ::StridedROCMatrix{$elty} = similar(A, $elty, min(size(A)...), size(A, 2));
+                kwargs...
+            )
+            chkstride1(A, U, Vᴴ, S)
+            m, n = size(A)
+            minmn = min(m, n)
+            srange, vl, vu, il, iu = _gesvdx_range($relty, kwargs)
+            maxnsv = _gesvdx_maxnsv(srange, il, iu, minmn)
+            jobu, jobvt = _gesvdx_jobs(U, Vᴴ, m, n, maxnsv)
```
lkdvos, line 237: Actually it seems like most of the other implementations have these just expanded out, so unless we can reuse this and share the helpers with the YALAPACK implementation, might be better to just keep the same style?

https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/9cd1f86116392c5e795085d6b5628d5b451430ad/src/yalapack.jl#L2216-L2231

src/implementations/svd.jl
```diff
@@ -142,10 +142,16 @@ function svd_via_adjoint!(f!::F, driver::Driver, A, S, U, Vᴴ; kwargs...) where
...
 
 # LAPACK
-for f! in (:gesdd!, :gesvd!, :gesvdx!, :gesdvd!)
+for f! in (:gesdd!, :gesvd!, :gesdvd!)
     @eval $f!(::LAPACK, args...; kwargs...) = YALAPACK.$f!(args...; kwargs...)
 end
 
+function gesvdx!(::LAPACK, A, S, U, Vᴴ; kwargs...)
+    YALAPACK.gesvdx!(A, S, U, Vᴴ; kwargs...)
+    _complete_svd_basis!(U, Vᴴ, length(S))
```
lkdvos: This might have to go inside of the YALAPACK implementation, I think that one checks that the rows/columns of V,U are not larger than `minmn`:

https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/9cd1f86116392c5e795085d6b5628d5b451430ad/src/yalapack.jl#L2251-L2252

src/implementations/svd.jl
```diff
@@ -221,7 +227,24 @@ for (f, f_lapack!, Alg) in (
...
 
 supports_svd_full(::Driver, ::Symbol) = false
-supports_svd_full(::LAPACK, f::Symbol) = f in (:safe_divide_and_conquer, :divide_and_conquer, :qr_iteration)
+supports_svd_full(::LAPACK, f::Symbol) = f in (:safe_divide_and_conquer, :divide_and_conquer, :qr_iteration, :bisection)
+
+# Some methods (e.g. `gesvdx`) only compute the leading `min(m, n)` singular vectors.
+# If `U` or `Vᴴ` is square (`svd_full!`), the remaining columns (row) of
+# `U` (`Vᴴ`) need to be filled with an orthonormal basis for the complement of the
+# computed singular vectors.
+function _complete_svd_basis!(U::AbstractMatrix, Vᴴ::AbstractMatrix, minmn::Int)
```
lkdvos: If we are using this in the extensions, I'd be fine with making this a non-underscore function as well?

src/implementations/svd.jl
```diff
@@ -221,7 +227,24 @@ for (f, f_lapack!, Alg) in (
...
+    if size(U, 2) > minmn
+        N = qr_null!(copy(view(U, :, 1:minmn)))
```
lkdvos:
```suggestion
        N = qr_null(view(U, :, 1:minmn))
```

src/implementations/svd.jl
```diff
@@ -221,7 +227,24 @@ for (f, f_lapack!, Alg) in (
...
+        copyto!(view(U, :, (minmn + 1):size(U, 2)), N)
+    end
+    if size(Vᴴ, 1) > minmn
+        V = view(Vᴴ, 1:minmn, :)
+        N = qr_null!(adjoint!(similar(V, reverse(size(V))), V))
```
lkdvos: Can we write this with `lq_null`?

src/implementations/svd.jl
lkdvos: I'm quite confused actually why the tests are passing, it seems like that shouldn't be happening 😕

src/implementations/svd.jl
kshyatt: Ah yeah likely

ext/MatrixAlgebraKitAMDGPUExt/yarocsolver.jl
kshyatt: I was doing this as a bit of a trial run for others, let me see if I can share the logic between YALAPACK and the extension

kshyatt: (Also we can reuse it for the batched version 😉 )

src/implementations/svd.jl
kshyatt: Wait, I'm confused about what's confusing

src/implementations/svd.jl
```diff
@@ -221,7 +227,23 @@ for (f, f_lapack!, Alg) in (
...
+supports_svd_full(::LAPACK, f::Symbol) = f in (:safe_divide_and_conquer, :divide_and_conquer, :qr_iteration, :bisection)
+
+# Some methods (e.g. `gesvdx`) only compute the leading `min(m, n)` singular vectors.
+# If `U` or `Vᴴ` is square (`svd_full!`), the remaining columns (row) of
+# `U` (`Vᴴ`) need to be filled with an orthonormal basis for the complement of the
+# computed singular vectors.
+function complete_svd_basis!(U::AbstractMatrix, Vᴴ::AbstractMatrix, minmn::Int)
+    if size(U, 2) > minmn
+        N = qr_null(view(U, :, 1:minmn))
+        copyto!(view(U, :, (minmn + 1):size(U, 2)), N)
```
Jutho, line 239: I guess this requires / assumes that `size(U, 2) == m`? 
Is it possible to directly do 
```julia
qr_null!(view(U, :, 1:minmn), view(U, :, (minmn+1):size(U,2))
```

src/implementations/svd.jl
```diff
@@ -221,7 +227,23 @@ for (f, f_lapack!, Alg) in (
...
+    end
+    if size(Vᴴ, 1) > minmn
+        N = lq_null(view(Vᴴ, 1:minmn, :))
+        copy!(view(Vᴴ, (minmn + 1):size(Vᴴ, 1), :), N)
```
Jutho, line 243: Same

src/implementations/svd.jl
kshyatt: ah yeah I assume `size(U, 2) == m` here since it's for the `svd_full` case. I'll try the inplace one.

src/implementations/svd.jl
Jutho: Me too, the checks on the number of column of `U` and the number of rows of `Vᴴ` are inequality checks, so more columns/rows then necessary are allowed:

https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/9cd1f86116392c5e795085d6b5628d5b451430ad/src/yalapack.jl#L2238
https://github.com/QuantumKitHub/MatrixAlgebraKit.jl/blob/9cd1f86116392c5e795085d6b5628d5b451430ad/src/yalapack.jl#L2247

kshyatt: Ah yes this is again for `svd_full` where we want `size(U, 2)` to be `m`, but I should explicitly check it's not `> m` also.

src/implementations/svd.jl
kshyatt: Ah the problem with doing this is `qr_null!` and `lq_null!` overwrite their input, so we'll have to do a copy in any case.

lkdvos: yes, but only one copy (the input), not the output right :)

kshyatt: cool, added this with a check to try to avoid copying the output...

src/implementations/svd.jl
lkdvos: Does this mean we are now allowing cases like `size(A) = (m, n)` with `m < n`, `size(U, 2) = k` with `m <= k <= n` as a valid input?

kshyatt: Oh yea I see, sorry, it should indeed not be `>=`. The `range == 'I'` case throws this stuff off, it's annoying

Jutho (approved):

